PR #1, opened 2019-06-12
add task and tests

vpolets:


src/rentalCarCost.js
```diff
@@ -0,0 +1,26 @@
...
+ * the car for 3 or more days, you get $20 off your total.
+ *
+ * Implement rentalCarCost function which returns the total
+ * amount for different count of days.
+ *
+ * @param {number} days
+ *
+ * @return {number}
+ */
+function rentalCarCost(days) {
```
roman-mate: function names typically start with a verb. calculateRentalCost

src/rentalCarCost.test.js
```diff
@@ -0,0 +1,18 @@
...
+  expect(total(2))
+    .toBe(80);
+});
+
+test('Should add the discount for 3 and more days of rent', () => {
+  expect(total(4))
+    .toBe(140);
+});
+
+test('Should add an additional discount for 7 and more days of rent', () => {
```
roman-mate: When testing the code, we typically split values into equivalence classes and test their boundaries. 

In this case, equivalence classes would be:
[0 - 2], 
[3 - 6], 
[7 - inf)

you typically want to test boundary values: 2, 3, 6, 7 because that's where errors typically happen.

PR #26, opened 2019-08-12
Fs jul19 alex andrutskii solution

aleks-and:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,17 @@
...
+  const price = 40;
+  const discountSevenDays = 50;
+  const discountThreeDays = 20;
+
+  if (days >= 7) {
+    return days * price - discountSevenDays;
+  } else if (days >= 3) {
+    return days * price - discountThreeDays;
+  } else {
+    return days * price;
```
danheim: else тут не нужен, можно сразу return (на новой строке, отделенный пустой строкой)
действие `days * price` повторяется, попробуй исправить ситуацию

danheim (changes requested):

src/calculateRentalCost.js
aleks-and: Roger that!
How about the next one? Now we can add as many discounts as we want (not caring about the order).

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  let discountsList = [];
```
YuriiHoliuk: use let only when necessary, here const will be enough

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+  class Discount {
```
YuriiHoliuk: class is really overkill here

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+    constructor(discountDays, discountValue) {
+      this.days = discountDays;
+      this.value = discountValue;
+    }
+  }
+  function createDiscountList(dList) {
```
YuriiHoliuk: do not declare functions inside function(There are reasons for it but not in this case)

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+    dList.forEach(function(dItem) {
+      discountsList.push(new Discount(dItem[0], dItem[1]));
```
YuriiHoliuk: functions should be pure and do not use variables in closure

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+  const price = 40;
+  let discountsList = [];
+  class Discount {
+    constructor(discountDays, discountValue) {
+      this.days = discountDays;
+      this.value = discountValue;
+    }
+  }
+  function createDiscountList(dList) {
+    dList.forEach(function(dItem) {
```
YuriiHoliuk: use arrow functions for callbacks

src/calculateRentalCost.js
YuriiHoliuk: dList and discountsList, looks over complicated

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+      discountsList.push(new Discount(dItem[0], dItem[1]));
+    });
+    discountsList = sortDiscountsFromMaxToMin(discountsList);
+  }
+  function compareDiscount(a, b) {
```
YuriiHoliuk: There are simpler way to compare 2 numbers

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+  }
+  function sortDiscountsFromMaxToMin(dList) {
+    return dList.sort(compareDiscount);
+  }
+  function rentalCost(rentalDays, priceForDay) {
+    let discountValue = 0;
+    for (let i = 0; i < discountsList.length; i++) {
+      if (rentalDays >= discountsList[i]['days']) {
+        discountValue = discountsList[i]['value'];
+        break;
```
YuriiHoliuk: do not use break, it is hard to maintain and bad practice

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+      return -1;
+    }
+    return 0;
+  }
+  function sortDiscountsFromMaxToMin(dList) {
+    return dList.sort(compareDiscount);
+  }
+  function rentalCost(rentalDays, priceForDay) {
+    let discountValue = 0;
+    for (let i = 0; i < discountsList.length; i++) {
```
YuriiHoliuk: do not use for loop for arrays use array methods instead

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+      if (rentalDays >= discountsList[i]['days']) {
```
YuriiHoliuk: use dynamic property access only when neccessary

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+    }
+    if (a.days > b.days) {
+      return -1;
+    }
+    return 0;
+  }
+  function sortDiscountsFromMaxToMin(dList) {
+    return dList.sort(compareDiscount);
+  }
+  function rentalCost(rentalDays, priceForDay) {
```
YuriiHoliuk: function name should start with a verb

src/calculateRentalCost.js
```diff
@@ -20,7 +20,45 @@
...
+    for (let i = 0; i < discountsList.length; i++) {
+      if (rentalDays >= discountsList[i]['days']) {
+        discountValue = discountsList[i]['value'];
+        break;
+      }
+    }
+    return rentalDays * priceForDay - discountValue;
+  }
+
+  createDiscountList([[3, 20], [7, 50]]);
```
YuriiHoliuk: maybe array of array is not the best data structure for this propose?

YuriiHoliuk (changes requested):
Looks like you try to create reusable, modular solution, that's great, but it is over complicated, try to write as simple code as you can

To0 make solution modular, with some utils in separate functions, these functions should be written separately from main function for you and other developers can reuse them

aleks-and:
> Looks like you try to create reusable, modular solution, that's great, but it is over complicated, try to write as simple code as you can
> 
> To0 make solution modular, with some utils in separate functions, these functions should be written separately from main function for you and other developers can reuse them

Got it. Thanks!
Remaking the current code according to your comments is too difficult for me at the moment, therefore, I just solve the task.

YuriiHoliuk (approved):

PR #42, opened 2019-09-23
RentalCost

Annetastar:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
```
gestap: better to create a variable for car cost.

src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
...
+  }
+  if (days >= 3 && days < 7) {
```
gestap: no need for this check
```suggestion
  if (days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
...
+    return 40 * days - 20;
+  }
+  if (days >= 7) {
```
gestap: also, this check unnecessary
```suggestion
```

src/calculateRentalCost.js
Annetastar: Ok)

ost-ptk (approved):

PR #43, opened 2019-09-23
Solution Transportation

lubremets:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,13 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
```
gestap: It will be cool if you create a variable for a car cost and use it.

ost-ptk (approved):

PR #49, opened 2019-09-23
fs_aug19_sasha_g

Shurkodr:
Aleksandr Glazunov
passed all local tests

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -21,15 +21,12 @@
  */
 function calculateRentalCost(days) {
   // write code here
-  let price;
   if (days >= 7) {
-    price = days * 40 - 50;
-  } else if (days >= 3) {
-    price = days * 40 - 20;
-  } else {
-    price = days * 40;
+    return days * 40 - 50;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -21,6 +21,15 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  let price;
```
gestap: I think you no need this variable

src/calculateRentalCost.js
```diff
@@ -21,6 +21,15 @@
...
+  if (days >= 7) {
+    price = days * 40 - 50;
```
gestap: we can return it at once
```suggestion
    return days * 40 - 50;
```

src/calculateRentalCost.js
```diff
@@ -21,6 +21,15 @@
...
+  } else if (days >= 3) {
+    price = days * 40 - 20;
+  } else {
```
gestap: if we return it at once, you can remove this `else`

ost-ptk (approved):

PR #52, opened 2019-09-23
Solution

jhunt05:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,14 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  if (days >= 7) {
+    return cost * days - 50;
+  } else if (days >= 3) {
+    return cost * days - 20;
+  } else {
```
gestap: no need for this else
```suggestion
  }
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,14 @@
...
+    return cost * days;
```
gestap: you can create variable for this operation `cost * days;`

src/calculateRentalCost.js
```diff
@@ -20,7 +20,14 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40 * days;
+  let discount = 0;
+  if (days >= 7) {
```
```suggestion

  if (days >= 7) {
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,14 @@
...
+    discount = 50;
+  } else if (days >= 3) {
+    discount = 20;
+  }
+  return cost - discount;
```
```suggestion

  return cost - discount;
```

ost-ptk (approved):

PR #54, opened 2019-09-23
Solution

CrazyNoodl:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,11 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 2 && days <= 6) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  } return days * 40;
```
```suggestion
  } 
  return days * 40;
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,11 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 2 && days <= 6) {
```
gestap: you can write more simple

```
if (days < 3) return ...
if (days < 7) return ...
return ...
```

CrazyNoodl: done

src/calculateRentalCost.js
CrazyNoodl: done

ost-ptk (approved):

PR #59, opened 2019-09-23
calculateRentalCost solution

veronika-donets:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let result;
+  if (days >= 7) {
+    result = days * 40 - 50;
```
gestap: you can write like this
```suggestion
    return days * 40 - 50;
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let result;
```
gestap: create variables for car cost and discount

ost-ptk (approved):

PR #60, opened 2019-09-23
Solution

Andrii-Pavlenko:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let amount = days * 40;
+  if (days >= 7) {
+    amount -= 50;
```
gestap: you can write 
```suggestion
   return amount -= 50;
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,16 @@
...
+    return amount;
+  } else if (days >= 3) {
+    amount -= 20;
+    return amount;
+  } else {
```
gestap: no need for this else
```suggestion
  }
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,13 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const amount = days * 40;
+  if (days >= 7) {
```
```suggestion

  if (days >= 7) {
```

src/calculateRentalCost.js
```diff
@@ -20,7 +20,13 @@
...
+    return amount - 50;
+  } else if (days >= 3) {
+    return amount - 20;
+  }
+  return amount;
```
```suggestion

  return amount;
```

ost-ptk (approved):

PR #101, opened 2020-01-20
solution

BDenysovets:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const REGULAR_PRICE = days * 40;
```
danheim: Why uppercase? We use camelcase.
```suggestion
  const regularPrice = days * 40;
```

What is `40`?

src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
...
+
+  if (days >= 7) {
+    return REGULAR_PRICE - 50;
```
danheim: What 50 does mean? Create a var for this.

src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
...
+  } else if (days >= 3) {
+    return REGULAR_PRICE - 20;
```
danheim: What 20 does mean? Create a var for this.

PR #103, opened 2020-01-20
solution

oksanaDaniluk:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,21 @@
...
+    total = 40 * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    total -= 20;
+  }
+
+  if (days >= 7) {
+    total -= 50;
+  }
```
danheim: What do 40 50 20 mean? Create variables for these numbers.

danheim (changes requested):

src/calculateRentalCost.js
```diff
@@ -20,7 +20,24 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40;
+  let totalCost = rentCost;
+  const minDiscount = 20;
+  const maxDiscount = 50;
```
danheim:
```
const ...;
const ...;
const ...;
let ...;
```
`let` after `const`

src/calculateRentalCost.js
```diff
@@ -20,7 +20,24 @@
...
+    totalCost = rentCost * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    totalCost -= minDiscount;
+  }
+
+  if (days >= 7) {
+    totalCost -= maxDiscount;
+  }
```
danheim:
```
if (condition) {
 // code
} else if (condition) {
 // code
}
```

danheim (changes requested):

src/calculateRentalCost.js
```diff
@@ -20,7 +20,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40;
+  const minDiscount = 20;
+  const maxDiscount = 50;
```
danheim: threeDaysDiscount
sevenDaysDiscount

oksanaDaniluk: done

PR #118, opened 2020-01-20
task was solved

LeonidShv:
ссылка на код:
https://github.com/LeonidShv/js_task-transportation-on-vacation/blob/develop/src/calculateRentalCost.js

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -21,14 +21,14 @@
  */
 function calculateRentalCost(days) {
   const costPerDay = 40;
-  const smallDiscount = 20;
-  const bigDiscount = 50;
+  const discountThreeDays = 20;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -20,7 +20,15 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return rentalCost - 20;
+  } else if (days >= 7) {
+    return rentalCost - 50;
+  }
```
danheim: what 50 and 20 mean? create vars

LeonidShv: ok
also I created a variable for price per day (40) 
I hope it'll ok

PR #129, opened 2020-01-20
Valeriy Sorochynskyi solution

Valeriy-Sorochynskyi:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,18 @@
...
+  const dayCost = 40;
+  let discount = 0;
+
+  if (days >= 3) {
+    discount = 20;
+  }
+
+  if (days >= 7) {
+    discount = 50;
+  }
```
danheim: what do 20, 50 mean?

src/calculateRentalCost.js
```diff
@@ -20,7 +20,18 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
```
danheim:
```suggestion
  const rentPrice = 40;
```

src/calculateRentalCost.js
danheim:
```
if (condition) {
  // code
} else if (condition) {
  // code
}
```

8hokage (approved):

PR #130, opened 2020-01-20
solution

morfeii:


src/calculateRentalCost.js
```diff
@@ -20,7 +20,20 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
-}
+  const everyDayRent = 40;
+  const lowCashback = 20;
+  const highCashback = 50;
```
danheim: `const threeDaysMore = ...;`
`const sevenDaysMore = ...;`

danheim (approved):

PR #206, opened 2020-05-21
Solution

Mishtal-Andrii:


src/calculateRentalCost.js
```diff
@@ -1,29 +1,19 @@
...
-  // write code here
+  const dayCost = 40;
+  const daysDiscountSmall = 3;
+  const daysDiscountBig = 7;
+  const discountSmall = 20;
+  const discountBig = 50;
+
+  if (days < daysDiscountSmall) {
+    return days * dayCost;
+  } else if (days >= daysDiscountSmall && days < daysDiscountBig) {
```
nazmeln, line 12: (days >= daysDiscountSmall you don't need this condition in second statement

src/calculateRentalCost.js
```diff
@@ -1,29 +1,19 @@
...
+    return days * dayCost - discountSmall;
+  } else if (days >= daysDiscountBig) {
```
nazmeln, line 14: also hear u can just write else without condition

src/calculateRentalCost.js
Mishtal-Andrii: без того не проходить тест

PR #208, opened 2020-05-21
Solution

campari229:
Hi


coroboX:
Look's like all good.
But think about the reason of changing `rentPrice` value.
It will be better to store result in some variable, and keep the `rentPrice` as const.

PR #237, opened 2020-08-01
solution

oggyrock:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
```
nanccyy, line 34: it makes sense to add 'days * 40' to a variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
+
+  if (days >= 7) {
+    return cost - 50;
+  } else if (days >= 3 && days < 7) {
```
vkryvytskyy, line 35: you already checked that days < 7
```suggestion
  } else if (days >= 3) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
+
+  if (days >= 7) {
```
vkryvytskyy, line 33: consider to use switch..case instead of if..else

PR #245, opened 2020-08-03
solved

nanccyy:


varseniuk (approved):
👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const payment = 40 * days;
+
+  if (days < 3) {
+    return payment;
+  }
+
+  if (days >= 3 && days <= 6) {
```
```suggestion
  if (days <=6) {
```
you already checked, that days not less than 3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const payment = 40 * days;
+
+  if (days < 3) {
```
vkryvytskyy, line 33: consider to use switch case instead of if..else

vkryvytskyi (approved):

PR #246, opened 2020-08-03
solution

furude-rika:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  let result = 0;
+
+  for (let i = 0; i < days; i++) {
+    result += cost;
+  }
```
DmJSCourse, line 36: loop here is redundant, you can simply multiply cost * days or 40 * days

DmJSCourse (commented):
Nice!

src/calculateRentalCost.js
furude-rika: thank you

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  let result = cost * days;
+
+  if (days >= 7) {
+    result -= 50;
```
vsm2k17, line 35:
```suggestion
    return result - 50;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  } else if (days >= 3 && days < 7) {
+    result -= 20;
```
vsm2k17, line 37:
```suggestion
  } else if (days >= 3) {
    return result - 20;
```

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):

PR #253, opened 2020-08-03
solution

Andrii256:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return (days * 40) - 20;
+  } else {
+    return (days * 40) - 50;
+  }
```
vsm2k17, line 37:
```suggestion
  }
  return (days * 40) - 50;
  
```

vsm2k17: consider adding a variable for price

PR #257, opened 2020-08-03
solution

varseniuk:


kostyasabada (approved):
Робота виконана)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  } else if (days < 7) {
+    return 40 * days - 20;
+  } else {
+    return 40 * days - 50;
+  }
```
vsm2k17, line 37: last else not necessary
```suggestion
  }
  return 40 * days - 50;
 
```

vsm2k17: consider adding a variable for the price, you use the same value 3 times

mykola-yarchuk (approved):

PR #258, opened 2020-08-03
Solution

oivannikov:


mpone (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
+
+  if (days >= 3 && days <= 6) {
+    return cost - 20;
+  } else if (days >= 7) {
```
vsm2k17, line 35: you can start with this, that's let you avoid the double condition above

mykola-yarchuk (approved):

PR #260, opened 2020-08-03
Solution

nznmv:


Totowa95 (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  const discountBig = 50;
+  const discountSmall = 20;
+  const pricePerDay = 40;
+
+  if (days >= 3 && days < 7) {
+    return (pricePerDay * days) - discountSmall;
+  }
+
+  if (days >= 7) {
```
vsm2k17, line 39: if swap this on top, and check this first you can avoid using && in upper "if" block

mykola-yarchuk (changes requested):

src/calculateRentalCost.js
nznmv: исправил

mykola-yarchuk (approved):

PR #264, opened 2020-08-03
solution

JuliaNastenko:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const rentPrice = 40;
+
+  if (days <= 0) {
+    return 0;
+  } else if (days < 3) {
```
vsm2k17, line 36: after return function stopped so no need to write "else if"
```suggestion
  }
  if (days < 3) {
```

PR #265, opened 2020-08-03
Solution

uni-8:


karinaKudatska (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const FULLCOST = days * 40;
+
+  if (days < 3) {
+    return FULLCOST;
+  }
+
+  if (days < 7) {
+    return FULLCOST - 20;
+  }
```
vkryvytskyy, line 39: use if..else or switch..case

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const FULLCOST = days * 40;
```
vkryvytskyy, line 31: use camel case for naming

vkryvytskyi (approved):

src/calculateRentalCost.js
uni-8: > use camel case for naming

i read that constants are written in uppercase and variables that are declared with let - in camelcase. is it correct to write in camelcase anyway?

PR #272, opened 2020-08-03
Solution

tykhan:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return days * 40 - 20;
```
vkryvytskyy, line 34:
```suggestion
    return (days * 40) - 20;
```
add parentheses or move `days * 40` to some variable

PR #273, opened 2020-08-03
Solution

mpone:


yuriimovchko (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const fromThreeDaysDiscount = 20;
+  const fromSevenDaysDiscount = 50;
```
vkryvytskyy, line 33:
```suggestion
  const smallDiscount = 20;
  const largeDiscount = 50;
```
Usually 2-3 words is enough to describe variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return days * costPerDay;
+  }
+
+  if (days < 7) {
+    return (days * costPerDay) - fromThreeDaysDiscount;
+  }
```
vkryvytskyy, line 41: use if..else or switch..case

vkryvytskyi (approved):

PR #276, opened 2020-08-03
Solution

KaterinaSavvova:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,11 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const rent = 40;
+
+  return days < 3 ? days * rent
+    : days < 7 ? days * rent - 20
+      : days * rent - 50;
```
vkryvytskyy, line 36: don't use ternary operator for conditions with more than one condition. In this case it is better to use if..else or switch..case

src/calculateRentalCost.js
```diff
@@ -29,6 +29,11 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const rent = 40;
+
+  return days < 3 ? days * rent
+    : days < 7 ? days * rent - 20
```
vkryvytskyy, line 35:
```suggestion
    : days < 7 ? (days * rent) - 20
```
use parentheses or move it to a variable

vkryvytskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -31,9 +31,15 @@ function calculateRentalCost(days) {
...
-  return days < 3 ? days * rent
-    : days < 7 ? days * rent - 20
-      : days * rent - 50;
+  if (days < 3) {
+    return (days * rent);
+  }
+
+  if (days < 7) {
+    return ((days * rent) - 20);
+  }
```
vkryvytskyy, line 40: combine into single if..else statement

vkryvytskyi (approved):

PR #280, opened 2020-08-03
solved

Chelovekozavr:


artsashko (approved):
Good!)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const sum = days * 40;
+
+  if (days < 3) {
+    return sum;
+  } else if (days >= 3 && days < 7) {
+    return sum - 20;
+  } else {
+    return sum - 50;
+  }
```
vkryvytskyy, line 39:
```suggestion
  }
return sum - 50;
```
you could omit last else

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const sum = days * 40;
+
+  if (days < 3) {
+    return sum;
+  } else if (days >= 3 && days < 7) {
```
vkryvytskyy, line 35:
```suggestion
  } else if (days < 7) {
```
you already check that days >= 3 in previous if

vkryvytskyi (approved):

PR #282, opened 2020-08-03
Solution

iamtheru:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return 40 * days - 50;
+  } else if (days >= 3) {
+    return 40 * days - 20;
+  } else {
+    return 40 * days;
+  }
```
```suggestion
  }
return 40 * days;
```
you could omit last else

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return 40 * days - 50;
```
vkryvytskyy, line 32:
```suggestion
    return (40 * days) - 50;
```
use parentheses or move this equation to the variable

PR #326, opened 2020-09-22
Develop

brainwashers:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let money = 0;
+
+  if (days < 3) {
+    money = days * 40;
+  }
+
+  if (days > 2 && days < 7) {
```
roman-mate, line 38:
```suggestion
  if (days >= 3 && days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+    money = days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    money = days * 40 - 50;
```
roman-mate, line 43: 50 and 20 are mgic numbers, it's good to have explicit formulas to understand where these magic numbers are coming from.

roman-mate (approved):

PR #344, opened 2020-09-23
Solution

qamay204:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,12 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days < 3) {
+    return 40 * days;
+  } else if (days >= 3 && days < 7) {
```
vkryvytskyy, line 34: you already cheked that days>=3, if this if is running

src/calculateRentalCost.js
```diff
@@ -29,6 +29,12 @@
...
+    return (40 * days) - 20;
+  } else {
+    return (40 * days) - 50;
+  }
```
vkryvytskyy, line 38: remove this else

vkryvytskyy:
```suggestion
  }
  return (40 * days) - 50;
```

vkryvytskyi (changes requested):

PR #373, opened 2020-11-26
Solution

rost-st:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40 * days;
+
+  if (days >= 3 && days < 7) {
+    return rentalCost - 20;
+  }
+
+  if (days >= 7) {
+    return rentalCost - 50;
```
vkryvytskyy, line 38: move 40, 20 and 50 to variables

vkryvytskyi (approved):

src/calculateRentalCost.js
rostyslav-meleshko: New variables declared.

PR #374, opened 2020-11-30
solution

Vasilkoff:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  } else {
+    return days * 40;
+  }
```
vkryvytskyy, line 37: move all numbers to the variables. So if I wanted to change something, I could change it in one place once

vkryvytskyi (approved):

PR #375, opened 2020-11-30
Solution

serhii-web:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days <= 6) {
+    return days * 40 - 20;
+  }
+
+  return days * 40 - 50;
 }
```
vkryvytskyy, line 38: move all numbers to variables(3, 6, 40, 20, 50). This way, if something changed - you need apply your changes only in one place

vkryvytskyy (approved):

src/calculateRentalCost.js
vpolets: Не пофікшено

vpolets (dismissed):

PR #386, opened 2020-12-07
Solution

ShyianO:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  if (days >= 7) {
+    result = days * 40 - 50;
+
+    return result;
+  } else if (days >= 3) {
+    result = days * 40 - 20;
```
Layrovell, line 38: числа 50 та 20 можна винести в окремі змінні, наприклад discountLarge = 50. також вираз days * 40 винести у  fullPrice

Layrovell (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
varseniuk, line 31: No need for this variable. In simple functions, you may return simple expressions directly
```suggestion

```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+    return result;
+  } else if (days >= 3) {
+    result = fullPrice - discountSmall;
+
+    return result;
+  } else {
+    result = fullPrice;
+
+    return result;
```
varseniuk, line 47:
```suggestion
  if (days >= 7) {
    return (fullPrice - discountLarge);
  } else if (days >= 3) {
    return (fullPrice - discountSmall);
  }
  return fullPrice;
```

wrap expressions in brackets.
As you may see, without unnecessary variable "result" the code becomes more cleaner

Also, you don't need the last "else" because if the first two checks won't pass, the function will return the last result by default

varseniuk (changes requested):

varseniuk (approved):

PR #388, opened 2020-12-07
Solution

tatarastrata:


MarchD (approved):
good :)

varseniuk (approved):
You are using the same number 40 (standard price) three times.
It would be a good practice to make a variable for that purpose.

e.g. 
```
const price = 40
...
return days * price - 20
```
etc...

mykola-yarchuk (approved):

PR #391, opened 2020-12-08
Solution

troospieler:


NazarDovhanchuk (approved):
nice!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40;
```
```suggestion
  }
  return days * 40;
```

there's no need in this last 'else' because if no one from the first two checks won't fit, the function will return the last case by default.

Also, it's a good practice to add a variable (constant) for data that is used several times. Like number 40 (general price) in this case.

For example
```
const price = 40;
....
return days * price;
```

varseniuk (approved):

src/calculateRentalCost.js
troospieler: thanks! will try to keep it in mind next time)

varseniuk (approved):

PR #392, opened 2020-12-08
solution

NazarDovhanchuk:


loralevitska (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicCost = 40;
+  const minSale = 20;
+  const maxSale = 50;
```
varseniuk, line 33: 👍 
Nice to create variables for reusable data.
You could also create another one for (days * basicCost)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return (days * basicCost) - maxSale;
+  } else if (days >= 3 && days < 7) {
+    return (days * basicCost) - minSale;
+  } else {
+    return days * basicCost;
+  }
```
```suggestion
  }
  return days * basicCost;
  
```
There's no need in last else. The function will return the last case by default if previous ones won't pass

varseniuk (approved):
Good 👍

varseniuk (approved):

PR #393, opened 2020-12-08
Solution

vorfolomeevav:


Serhii-Alieksieievych (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
+  const discountFromThreeDays = 20;
+  const discountFromSevenDays = 50;
+  const totalAmount = days * dayCost;
```
varseniuk, line 34: 👍 
Good practice to create variables for main reusable data

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days < 3) {
+    return totalAmount;
+  } else if (days >= 3 && days < 7) {
+    return totalAmount - discountFromThreeDays;
+  } else {
+    return totalAmount - discountFromSevenDays;
+  }
```
```suggestion
  }
    return totalAmount - discountFromSevenDays;
```

There's no need in last 'else', because if the previous two checks will not pass, so the function will return the last case by default

varseniuk (approved):
Great job!

varseniuk (approved):

PR #397, opened 2020-12-08
Vacation transportation - Solution

VinnieJ-2k20:


AlieksieienkoAndriy (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyCost = 40;
+  const disc3 = 20;
+  const disc7 = 50;
```
varseniuk, line 33: It's good practice to make variables for reusable data, but let's give it more meaningful names.
It's not clear what does 'disc3' or 'disc7' means. A name like 'discount3' would be more understandable.

varseniuk (approved):
In the future use more meaningful and understandable names of variables

varseniuk (approved):

PR #400, opened 2020-12-08
solution

dtsehelnyk:


varseniuk (changes requested):
Where is the student's review?

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+
+  if (days >= 7) {
+    return days * dailyPrice - discountMax;
+  } else if (days >= 3) {
+    return days * dailyPrice - discountMin;
+  } else if (days >= 1) {
+    return days * dailyPrice;
+  } else {
+    return 'something is wrong';
+  }
```
kerpichonka, line 44: good solution))

kerpichonka (approved):
good job)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const dailyPrice = 40;
+  const discountMin = 20;
+  const discountMax = 50;
+
+  if (days >= 7) {
+    return days * dailyPrice - discountMax;
```
```suggestion
  if (days >= 7) {
    return days * (dailyPrice - discountMax);
```

vkryvytskyy: use parentheses

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+  } else if (days >= 3) {
+    return days * dailyPrice - discountMin;
```
vkryvytskyy, line 39:
```suggestion
    
    if (days >= 3) {
    return days * dailyPrice - discountMin;
```
else could be removed, since you use return in previous if

src/calculateRentalCost.js
vkryvytskyy, line 44: if previous two ifs didn't work, then it's definitely number, that is less than 7 and 3, so you could just return days * dailyPrice

vkryvytskyy (changes requested):

varseniuk (approved):
FYI:
you can also add variables like 
`const totalCost = (days * dailyPrice)`

PR #402, opened 2020-12-09
Solution

paramonovfrontend:


MaximZhuravlov (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayOfRent = 40;
```
varseniuk, line 31: The name of the variable 'dayOfRent' is not meaningful at all. In the future try to use some simple and meaningful names. In this case, it would be more understandable if you would call it 'general price'.

But it's a good practice declaring variables for reusable data 👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days >= 7) {
+    sum = dayOfRent * days - 50;
+  } else if (days >= 3) {
+    sum = dayOfRent * days - 20;
+  } else {
+    sum = dayOfRent * days;
+  }
+
+  return sum;
```
```suggestion
  } 
  
  return sum = dayOfRent * days;
```

You don't need the last 'else' because if the previous two checks won't pass, the function will return the last case by default

varseniuk (approved):
👍

varseniuk (approved):

PR #403, opened 2020-12-09
Solution

MaximZhuravlov:


SofiiaAsieieva (approved):
Cool solution

src/calculateRentalCost.js
```diff
@@ -22,13 +22,21 @@
...
- * @param {number} days
+ * @param {number} days the length of the vacation.
  *
- * @return {number}
+ * @return {number} the total cost of the vacation.
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = 40 * days;
```
varseniuk, line 31: It's good practice to use variables. 👍 
You could also add other variables for different discounts (just advice for the future)

src/calculateRentalCost.js
```diff
@@ -22,13 +22,21 @@
  *  - calculateRentalCost(3) === 100
  *  - calculateRentalCost(7) === 230
  *
- * @param {number} days
+ * @param {number} days the length of the vacation.
```
varseniuk, line 25: Writing a description of your functions is always a good idea 👍

varseniuk (approved):
Good job

PR #415, opened 2020-12-09
Solution

Layrovell:


xipholena (approved):
Well done!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return price - discontLarge;
+  }
+
+  if (days >= 3) {
+    return price - discontSmall;
+  }
+
+  if (days < 3) {
+    return price;
+  }
```
varseniuk, line 46:
```suggestion
  return price;
```

Last 'if' is not necessary. Because if the first two checks won't pass, so the function will return the last one by default

varseniuk (approved):

PR #421, opened 2020-12-09
solution

Ruslanwp:


varseniuk (changes requested):
Where is the student's review?

Pasha5555 (approved):
Good!

varseniuk (approved):
Good job : 👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const discount = 20;
+  const bigDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return days * paymentPerDay - discount;
+  }
+
+  if (days >= 7) {
+    return days * paymentPerDay - bigDiscount;
+  }
```
varseniuk, line 41: Swap the first and the second checks and remove && < 7.
In that case, the logic is next: 
1. If days >=7 - return sum with big discount
2. if days >=3 - return sum ith regular discount
3. if not >=7 or not >=3 - return just sum

Your solution is correct but always try to simplify code

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  return days * paymentPerDay;
```
```suggestion
  return (days * paymentPerDay);
```
Wrap expressions in brackets

varseniuk (approved):

taniazt (approved):

PR #423, opened 2020-12-09
solution

Pasha5555:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return 40 * days - 50;
```
Ruslanwp, line 32: create some variable and use instead of number '40'

Ruslanwp: else you can delete one of your conditions and leave only `return....`. Nothing change

Ruslanwp (changes requested):

Ruslanwp (approved):
Great!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return rent * days - 50;
+  }
+
+  if (days >= 3) {
+    return rent * days - 20;
+  } else {
+    return rent * days;
+  }
```
vkryvytskyy, line 41:
```suggestion
  }
  
  return rent * days;
```
else is redundant

vkryvytskyi (approved):

PR #424, opened 2020-12-09
Solution

NikitaTudakov:


varseniuk (changes requested):
Where's student's review?

NikitaTudakov:
> Where's student's review?

i forgot for it, sry

vitalya2626 (approved):
Nice!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return (days * 40);
+  } else {
+    if (days < 7) {
```
varseniuk, line 34: There's no need for such construction.
You may write `else if {}` or just another `if{} `after the first check. Because if the first check pass and some result will be returned, the function will break immediately.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+      return (days * 40 - 20);
+    } else {
+      return (days * 40 - 50);
+    }
```
varseniuk, line 38:
```suggestion
    }
    return (days * 40 - 50);
```

No need in this 'else'. This result will be returned by default if both previous checks won't pass

varseniuk (changes requested):
It's good practice to create variables for reusable data and not to use just 'magic numbers'.
For example, you use the number 40 three times in your code. You can just declare 
`const standardPrice = 40` and use it everywhere you need. Or, you may add variable 
`const totalPrice = (days * 40)`
and so on

src/calculateRentalCost.js
NikitaTudakov: all remarks has corrected!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    return totalPrice;
+  }
+
+  if (days < 7) {
+    return (totalPrice - 20);
+  }
+
+  return (days * 40 - 50);
```
```suggestion
  return (totalPrice - 50);
```

You can also add variables for numbers 20 and 50 (e.g. named minDiscount and maxDiscount)

varseniuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalPrice = days * 40;
+  const minDisc = 20;
+  const maxDisc = 50;
```
varseniuk, line 33: 'disc' - it's not a meaningful name. Use 'discount' instead. Somebody else will not understand what means 'disc' in your code.

varseniuk (approved):

PR #428, opened 2020-12-09
solution

5Mountains:


Nvalja (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+  const discount = 20;
+  const maxDiscount = 50;
+
+  if (days < 3) {
+    return (price * days);
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * price - discount);
+  }
```
```suggestion
  if (days < 7) {
    return (days * price - discount);
  }
```
no need to check if days >=3 because it's obvious if the first check (days < 3) didn't pass.

varseniuk (changes requested):
You may also add a variable for this expression (days * price)
e.g.
`const totalCost = (days * price)`
and use it in your code

src/calculateRentalCost.js
5Mountains: done!

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
```
varseniuk, line 31: Next time remove this line (or any other comments)

varseniuk (approved):
👍

PR #430, opened 2020-12-09
Solution

andrew-zaycev:


sashalarin01 (approved):

sashalarin01:
Great job😎

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
```
varseniuk, line 31: It's a good practice to create variables for reusable data, but let's create other variables for discounts (50 and 20) an use them instead of 'magic numbers'

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days < 3) {
+    return cost;
+  } else if (days < 7) {
+    return cost - 20;
```
```suggestion
    return (cost - 20);
```
use brackets with expressions

varseniuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return cost;
+  } else if (days < 7) {
+    return (cost - smallDiscount);
+  } else {
+    return (cost - bigDiscount);
+  }
```
```suggestion
  }
  return (cost - bigDiscount);
```

there's no need in last 'else' because if both previous checks won't pass, the function will return the last one by default

varseniuk (changes requested):

varseniuk (approved):

varseniuk (approved):

PR #431, opened 2020-12-09
solution calculate

Nvalja:


5Mountains (approved):
Nice job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
-  // write code here
+  const rent = 40;
+  const sale1 = 20;
+  const sale2 = 50;
+  const duration1 = 3;
+  const duration2 = 7;
+
+  if (days < duration1) {
+    return days * rent;
+  }
```
```suggestion
  if (days < duration1) {
    return (days * rent);
  }
```
use expressions in brackets

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < duration2) {
+    return days * rent - sale1;
+  } else {
+    return days * rent - sale2;
+  }
```
varseniuk, line 45:
```suggestion
  }
  return (days * rent - sale2);
```

you don't need the last `else`. Because if none from both previous checks won't return anything, this function will return the last case by default

varseniuk (changes requested):
You may also add expression (days * rent) as variable
`const totalCost = (days * rent)`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  const sale1 = 20;
+  const sale2 = 50;
+  const duration1 = 3;
+  const duration2 = 7;
```
varseniuk, line 35: Use more meaningful names of variables in future.
E.g. not 'sale1' and 'sale2' but 'smallDiscount' and 'bigDiscount' and not 'duration1' but 'shortDuration' or 'shortPeriod' etc.

varseniuk (approved):

PR #433, opened 2020-12-09
Solution

taniazt:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+  if (days < 3) {
+    sum = days * 40;
+  };
+
+  if (days > 2 && days < 7) {
+    sum = days * 40 - 20;
+  };
+
+  if (days > 6) {
+    sum = days * 40 - 50;
```
ShyianO, line 43: Все ок, але в ідеалі потрібно винести числа в окремі змінні. Наприклад, 50 - discountLage, fullPrice = days * 40;

ShyianO (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let sum;
```
varseniuk, line 32:
```suggestion
```
No need for this variable. In simple functions, you may return simple expressions directly

Don't leave commented lines in your code before pushing solutiong

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
...
+  const dayilyPay = 40;
+  const discountSmall = 20;
+  const discountBig = 50;
+
+  if (days < 3) {
+    sum = days * dayilyPay;
+  };
```
```suggestion
  if (days < 3) {
    return (days * dayilyPay);
  };
```

  wrap expressions in brackets

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
...
+
+  if (days > 2 && days < 7) {
+    sum = days * dayilyPay - discountSmall;
+  };
```
```suggestion
  if (days < 7) {
   return (days * dayilyPay - discountSmall);
  };
```

you may simplify your function if you will return something directly inplace

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
...
+
+  if (days > 6) {
+    sum = days * dayilyPay - discountBig;
+  }
+
+  return sum;
```
varseniuk, line 49:
```suggestion
    return (days * dayilyPay - discountBig);
```

no need for that last check, because if the first two checks won't pass, so the function will return this result by default

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let sum;
+  const dayilyPay = 40;
+  const discountSmall = 20;
+  const discountBig = 50;
+
```
varseniuk, line 36: You may also add another variable e.g. totalPrice = days * dayilyPay

varseniuk (changes requested):

taniazt:
All suggestions applied.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayilyPay = 40;
+  const discountSmall = 20;
+  const discountBig = 50;
+  const totalPrice = days * dayilyPay;
+
+  if (days > 2 && days < 7) {
```
```suggestion
  if (days >= 3 && days < 7) {
```
It's ok but better to stick with numbers that describe the number of days more precisely.
It's not critical but good practice

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+    return (totalPrice - discountSmall);
+  };
+
+  if (days > 6) {
```
```suggestion
  if (days >= 7) {
```
The same

varseniuk (approved):
Left two little comments. Take it into consideration

PR #436, opened 2020-12-09
Solution

Alexandr-Petrenko:


Droshik (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let payment = days * 40;
+
+  if (days > 6) {
+    payment -= 50;
+  } else if (days > 2) {
```
```suggestion
  if (days >= 7) {
    payment -= 50;
  } else if (days >= 3) {
```
For better understanding, it would be better to use days number with 'more-or-equal' or 'less-or-equal' signs.

Next time use variables instead of 'magic numbers' 50 and 20

varseniuk (approved):

PR #437, opened 2020-12-10
Solution

viktor-morhun:
.

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -6,17 +6,17 @@ function calculateRentalCost(days) {
   const longTermDiscount = 50;
   const shortTermPeriod = 3;
   const shortTermDiscount = 20;
-  let totalCost = dailyPrice * days;
+  const totalCost = dailyPrice * days;
 
-  if (days >= shortTermPeriod) {
-    totalCost -= shortTermDiscount;
-
-    if (days >= longTermPeriod) {
-      return (totalCost - (longTermDiscount - shortTermDiscount));
-    }
+  if (days < shortTermPeriod) {
+    return totalCost;
   }
 
-  return totalCost;
+  if (days < longTermPeriod) {
+    return (totalCost - shortTermDiscount);
+  }
+
+  return totalCost - longTermDiscount;
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -1,34 +1,6 @@
 'use strict';
 
-/**
- * After a hard quarter in the office you decide
- * to get some rest on a vacation.
- * So you will book a flight for you
- * and your family and try to leave all the mess behind you.
- *
- * You will need a rental car in order for you to get around in your vacation.
- * The manager of the car rental makes you some good offers.
- *
- * Every day you rent the car costs $40.
- * If you rent the car for 7 or more days, you get $50 off your total.
- * Alternatively, if you rent the car for 3 or more days,
- * you get $20 off your total.
- *
- * Implement calculateRentalCost function
- * which returns the total amount for different count of days.
- *
- * Examples:
- *  - calculateRentalCost(1) === 40
- *  - calculateRentalCost(3) === 100
- *  - calculateRentalCost(7) === 230
- *
- * @param {number} days
- *
- * @return {number}
- */
-
 function calculateRentalCost(days) {
-  // write code here
   const dailyPrice = 40;
   const longTermPeriod = 7;
   const longTermDiscount = 50;
@@ -40,7 +12,7 @@ function calculateRentalCost(days) {
     totalCost -= shortTermDiscount;
 
     if (days >= longTermPeriod) {
-      totalCost -= (longTermDiscount - shortTermDiscount);
+      return (totalCost - (longTermDiscount - shortTermDiscount));
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+  const longTermDiscount = 50;
+  const shortTermPeriod = 3;
+  const shortTermDiscount = 20;
+  let totalCost = dailyPrice * days;
+
+  if (days >= shortTermPeriod) {
+    totalCost -= shortTermDiscount;
+
+    if (days >= longTermPeriod) {
+      totalCost -= (longTermDiscount - shortTermDiscount);
```
andrew-zaycev, line 43: I think need use construction 'else if' instead of another 'if'

andrew-zaycev (changes requested):
Good job ^_^
But i think you need to be changed something

src/calculateRentalCost.js
viktor-morhun: we need second `if` only if first condition is true, that`s why i didn`t use `else if`

andrew-zaycev: Okay, maybe you are right)

andrew-zaycev (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const dailyPrice = 40;
+  const longTermPeriod = 7;
+  const longTermDiscount = 50;
+  const shortTermPeriod = 3;
+  const shortTermDiscount = 20;
+  let totalCost = dailyPrice * days;
```
varseniuk, line 37:
```suggestion
  const totalCost = dailyPrice * days;
```
it's better to use constant here and in any place you need just `return totalCost` or `(totalCost - someTermDiscount)`. Don't forget brackets around expressions.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
```
varseniuk, line 31:
```suggestion

```
remove all commented lines before sending solution

varseniuk (changes requested):
Your solution is working but let's try to make it more efficient and productive.
Now your code ALWAYS does two checks:
if (days >= shortTermPeriod) and some operations after that
and right after that the next check
if (days >= longTermPeriod) and some operations after that again,
even if first was 'true'. So we ALWAYS have to checks, even if it's not necessary.
Let's cut the number of operations of this program. 

Let's return some expression right after checking the number of days.
Watch [this Misha's video](https://www.youtube.com/watch?v=YaR7e9z69vk&feature=emb_logo) for a better understanding.

viktor-morhun:
> Now your code ALWAYS does two checks:
no, it doesn`t
> if (days >= shortTermPeriod) and some operations after that
> and right after that the next check
> if (days >= longTermPeriod) and some operations after that again,
> even if first was 'true'. So we ALWAYS have to checks, even if it's not necessary.
ONLY if first was 'true', we use second `if` [proof](https://www.loom.com/share/29bce697f8124268a24efe22d5c2f4ff)

> Watch [this Misha's video](https://www.youtube.com/watch?v=YaR7e9z69vk&feature=emb_logo) for a better understanding.
i watched this video and understood, that Misha`s code has the same number of checks (screenshot below)
image

varseniuk:
Of course, there MUST be THE SAME NUMBER OF CHECKS (at least two). I said that there should be LESS operations. Now I'll try to explain, what I meant step by step in your code.
Let's suppose, we have `days = 8`. Then what your code does:
1. First check
```
 if (days >= shortTermPeriod) {
    totalCost -= shortTermDiscount;
```
as 8 > than 3 then
FIRST mathematical operation = totalCost = 40 * 8 - 20 = 300

2. Right after that operation executes, here goes the next check and next mathematical operation (in any case, because function returns nothing and continues to execute):
```
if (days >= longTermPeriod) {
      totalCost -= (longTermDiscount - shortTermDiscount);
    }
```
so now the SECOND mathematical operation is  totalCost = 300 - (50 - 20) = 270
And now function returns number 270.
We had TWO mathematical operations.

Now let's see the code from the video.
1. First check
```
if (days < shortTermPeriod) {
  return basePrice
}
```
because days(8) not less than shortTermPeriod (3) function DOES/RETURN NOTHING and goes to next check
2. Second check
```
if (days < longTermPeriod) {
  return basePrice - shortTimeDiscount;
}
```
again, 8 not < than 7, so AGAIN FUNCTION DOES/RETURN NOTHING and goes to DEFAULT CASE (if none of the checks haven't passed)
3. Function returns (and it's the FIRST mathematical OPERATION)
basePrice - longTermDiscount = 8 * 40 - 50 = 270 and returns that value.

Conclusion: we had TWO CHECKS and ONLY ONE mathematical operation instead of TWO CHECKS and TWO mathematical operations in your code.
And that's what I meant - you don't need to implement less checks, but you need to try to optimize your code to make less operation of different kinds.
Of course, your code is working correctly, this task is small and extra mathematical operation will not ruin your program and will not slow down the PC, but our purpose is to learn to write simple and effective code.

Hope I explained what I mean.

varseniuk (approved):
Good job on mistakes and respect for two solutions 👍

PR #452, opened 2021-01-24
My solution

IhorDrul:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const cost = days * 40;
+
+  if (days < 3) {
+    return cost;
+  } else if (days >= 3 && days < 7) {
+    return cost - 20;
+  } else {
+    return cost - 50;
```
danheim, line 39: create vars for the discounts

danheim (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const cost = days * 40;
+  const afterThreeDays = cost - 20;
+  const afterSevenDays = cost - 50;
```
danheim, line 34: Not that way.
```suggestion
  const threeDaysDiscount = 20;
  const sevenDaysDiscount = 50;
```

danheim (changes requested):

danheim (approved):

PR #499, opened 2021-02-26
Solution

herkoss:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let res = 0;
+
+  for (let index = 0; index < days; index++) {
+    res += 40;
+  }
```
VinnieJ-2k20, line 35: No need to use loops. This is simple arithmetic calculation.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return res;
+  } else if (days >= 3 && days < 7) {
```
VinnieJ-2k20, line 39: If you use `return` inside `if` block, using `else` is useless.
This is the correct structure:
```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```

VinnieJ-2k20 (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const res = days * 40;
+
+  if (days < 3) {
```
aionova, line 33: don't use magic numbers. Put all the numbers in variables.

aionova (approved):
Good job!
Fix magic numbers

PR #500, opened 2021-02-26
Solution

Andriykooo:


VinnieJ-2k20 (approved):
Interesting solution. I'll approve because it works correctly.

Some of advice for future tasks:
- If it is possible to solve task without loops - don't use loops. For massive calculations they slow down the operations.
- Implement variables for reusable data, like discounts, costPerDay etc.

PR #509, opened 2021-02-28
Solution

Julietta-M:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let amount = 0;
+
+  if (days >= 7) {
+    amount = (40 * days) - 50;
```
danheim, line 34: create var for the price

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days >= 3) {
+    amount = (40 * days) - 20;
```
danheim, line 36: create vars for the discounts (50, 20)

danheim (changes requested):

danheim (approved):

PR #513, opened 2021-03-01
Solution

vladdezhniuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let result = 0;
+
+  if (days < 3) {
+    result += days * 40;
+  } else if (days < 7) {
+    result += days * 40 - 20;
+  } else {
+    result += days * 40 - 50;
+  }
```
danheim, line 39: create vars for the price and discounts

danheim (changes requested):

danheim (approved):

PR #515, opened 2021-03-01
Solution

TarasChaban:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 6) {
+    return days * 40 - 50;
+  } else if (days > 2) {
+    return days * 40 - 20;
+  } else {
+    return days * 40;
```
danheim, line 36: create vars for the price and discounts

danheim (changes requested):

danheim (approved):

PR #516, opened 2021-03-02
Solution

bohdan-prokopov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
```
danheim, line 36: create vars for the price and discounts

danheim (changes requested):

danheim (approved):

PR #521, opened 2021-03-02
Solution

ruslanvasylyshyn:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  let money = 0;
```
danheim, line 33:
```suggestion
  let rentalCost = 0;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+
+  if (days < 3) {
+    money = 40 * days;
+  } else if (days > 2 && days <= 6) {
+    money = (40 * days) - 20;
+  } else if (days >= 7) {
+    money = (40 * days) - 50;
```
danheim, line 40: create vars for the price and discounts

danheim (changes requested):

danheim (approved):

PR #523, opened 2021-03-02
Develop

den-bulaev:


VinnieJ-2k20 (approved):
Good job! 👍 

You can also use variables for discounts, as it is reusable data.

PR #526, opened 2021-03-07
Solution

Taras-Konovalov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  }
 }
```
aliceblunt8, line 38: the last else if can be omitted

```suggestion
  if (days < 3) {
    return days * 40;
  } else (days >= 3 && days < 7) {
    return days * 40 - 20;
  } 
    return days * 40 - 50;
  
}
```

aliceblunt8 (approved):

PR #532, opened 2021-03-12
Solution

iryna-vys:


VinnieJ-2k20 (approved):
Some advice for the future:
- If the calculation is repetative (here it is `days * 40`) - move it to variable and write just once.
- Reusable data like daily cost, discounts etc should also be used with variables.

PR #533, opened 2021-03-12
Solution

SerhiiKolesnyk:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let sum = 0;
```
aliceblunt8, line 32: let's calculate the total sum of rent for 40 days once and then just subtract the discount in each case

aliceblunt8 (changes requested):
let's

aliceblunt8 (approved):
Nice, but I think it's better to calculate the total sum of rent for 40 days once and then just subtract the discount in each case

PR #535, opened 2021-03-12
Successful solution

IvanGrekov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
+    return days * 40 - 20;
+  } else {
+    return days * 40;
+  }
```
aionova, line 37:
```suggestion
  }

  return days * 40;
```
 no need to use else after return statement

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
aionova, line 32: let's not use magic numbers. Put all the numbers in variables and then use them

aionova (changes requested):
Great start!

aionova (approved):
Great!

PR #537, opened 2021-03-13
Solution

MykolaButylkov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  if (days === 0) {
+    return result;
+  }
```
aionova, line 35: do you really need this condition?

aionova (changes requested):
Let's not use magic numbers

PR #539, opened 2021-03-14
solution

Andrey203:


aionova (approved):
Good job!
Use variables for numbers

PR #543, opened 2021-03-16
Solution

amitheutos:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,9 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const sum = days * 40;
+
+  return days >= 7 ? sum - 50 : days >= 3 ? sum - 20 : sum;
```
aionova, line 34: Use if else or switch case here. This operation is quite hard to read. We don't write like that on real projects.
Also don't use magic numbers. Put all the numbers in variables

aionova (changes requested):
Great start!

PR #550, opened 2021-03-26
solution

Namart:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days >= threeDays && days < sevenDays) {
+    result = days * rentalPrice - threeDaysDiscount;
+  } else if (days >= sevenDays) {
+    result = days * rentalPrice - sevenDaysDiscount;
+  } else {
+    result = days * rentalPrice;
+  }
+
+  return result;
```
aliceblunt8, line 46: you can return the value right away, no need to create a variable
```suggestion

  if (days >= threeDays && days < sevenDays) {
    return days * rentalPrice - threeDaysDiscount;
  } else if (days >= sevenDays) {
    return days * rentalPrice - sevenDaysDiscount;
  } else {
    return days * rentalPrice;
  }

```

aliceblunt8 (approved):

PR #560, opened 2021-04-19
Rental cost

leonid-vegera:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
+  } else if (days >= 3) {
```
VinnieJ-2k20, line 33: If you write `return` inside `if` block, using `else` after that doesn't make sense.
```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):

For the future:
Use variables for reusable data, like costPerDay, discounts etc.
If you have repetitive calculations (like `days * 40` in this case) - move it also to a variable and calculate just once.

PR #562, opened 2021-04-19
Solution

Nazar-Kudatsky:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
```
aliceblunt8, line 31: let's create variables for discounts and total amount of days

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  let count = 0;
+
+  if (days >= 3 && days <= 6) {
```
aliceblunt8, line 33:
```suggestion
  if (days >= 3 && days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    count += days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    count += days * 40 - 50;
+  }
+
+  if (days < 3) {
+    count += days * 40;
```
aliceblunt8, line 42: let's calculate the total sum of rental only once, not 3 times in each case

aliceblunt8 (changes requested):

src/calculateRentalCost.js
aliceblunt8: what about variables?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
+  const sum = days * 40;
+
+  if (days >= 3 && days < 7) {
+    count += sum - 20;
```
aliceblunt8, line 35: I believe you don't need count variable now, cause you can return the result straight away

```suggestion
    return sum - 20;
```

aliceblunt8 (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
+  const priceday = 40;
+  const mincount = 20;
+  const maxcount = 50;
```
aliceblunt8, line 34: use snake case for variables

```suggestion
  const carRentalPrice = 40;
  const minDiscount = 20;
  const maxDiscount = 50;
```

aliceblunt8 (approved):

PR #563, opened 2021-04-19
solution

RuslanaPodshyvalova:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+    case (days < 3):
+      totalCost = costPerDay * days;
+      break;
+    case (days >= 3 && days < 7):
+      totalCost = costPerDay * days - 20;
+      break;
+    case (days >= 7):
+      totalCost = costPerDay * days - 50;
+      break;
+  }
```
VinnieJ-2k20, line 48: This is not the best way to use `switch/case`
Please, rewrite it using `if/else` structure

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  let totalCost = 0;
+
+  if (days < 0) {
+    return null;
+  }
```
VinnieJ-2k20, line 36: 👍

VinnieJ-2k20 (changes requested):
Almost done. Some changes needed.

VinnieJ-2k20 (approved):
I'll accept because it works correctly.
Some advice for the future tasks:

If there's an option to avoid using `else` - don't use it.
Better structure in this case would be:
```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```

PR #565, opened 2021-04-20
solution

SergeyZub:


VinnieJ-2k20 (approved):

For the future:
Use variables for reusable data, like costPerDay, discounts etc.
If you have repetitive calculations (like `days * 40` in this case) - move it also to a variable and calculate just once.

PR #566, opened 2021-04-20
solution

amitheutos:
magic numbers were fixed

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceOfOneDayRent = 40;
+  const discountForWeek = 50;
+  const discountForThreeDays = 20;
+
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const discountForWeek = 50;
+  const discountForThreeDays = 20;
+
+  const finalAmount = days * priceOfOneDayRent;
+
+  switch (true) {
+    case days >= 7 : return finalAmount - discountForWeek;
+    case days >= 3 : return finalAmount - discountForThreeDays;
+    default : return finalAmount;
+  }
```
aionova, line 41:
```suggestion
  switch (true) {
    case days >= 7:
      return finalAmount - discountForWeek;
    case days >= 3:
      return finalAmount - discountForThreeDays;
    default : return finalAmount;
  }
```

aionova (approved):
Good job!

PR #567, opened 2021-04-20
Solution

igor-stasiv:


SkyPKWolf (approved):
Nice

VinnieJ-2k20 (approved):

For the future:
If you have repetitive calculations (like `days * dayCost` in this case) - move it also to a variable and calculate just once.

PR #568, opened 2021-04-20
Solution

pkuzmenkoo:


igor-stasiv (approved):
Good

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
-}
+  const costRentDay = 40;
+  const discountMin = 20;
+  const discountMax = 50;
+
+  if (days >= 7) {
+    return days * costRentDay - discountMax;
+  } else if (days >= 3) {
+    return days * costRentDay - discountMin;
+  }
```
VinnieJ-2k20, line 39: If you write `return` inside `if` block, using `else` after that doesn't make sense.

```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```

VinnieJ-2k20 (changes requested):
Use of variables - 👍 

The rest needs small edits.

aliceblunt8 (approved):

PR #573, opened 2021-04-21
first solution

scarlettjohanssons:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let totalCost = 0;
+
+  if (days < 3) {
+    totalCost = days * 40;
```
Ihor-Karpyn, line 35: days * 40 - you do this 3 times. You should do this on line 32.

Ihor-Karpyn (changes requested):

aliceblunt8 (approved):

PR #574, opened 2021-04-21
Develop

Ihor-Karpyn:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
```
scarlettjohanssons, line 31: total what?
Maybe rename to something specific?

scarlettjohanssons (changes requested):

src/calculateRentalCost.js
Ihor-Karpyn: The function is called calculateRentalCost, so total is the full rental amount. I don't think I should change the name of a variable

scarlettjohanssons (approved):

aliceblunt8 (approved):

PR #575, opened 2021-04-21
Solution

avgusta92:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  const price = days * 40;
+
+  switch (true) {
+    case days >= 7:
+      return price - 50;
+    case days >= 3:
+      return price - 20;
+    default:
+      return price;
+  }
```
VinnieJ-2k20, line 40: This is not the best way to `use switch/case` structure.

Rewrite it using `if`

VinnieJ-2k20 (changes requested):
Some edits needed.

Also, reusable data like costPerDay, discounts etc should also be moved to variables.

src/calculateRentalCost.js
avgusta92: Ok) I will fix it. But I don't understand why I can't use switch/case for this task?

avgusta92:
> Some edits needed.
> 
> Also, reusable data like costPerDay, discounts etc should also be moved to variables.

Can you clarify? If I use value just one time I should use variables for this value too? Because, at now, I use "costPerDay" and "discounts" values just one time.

src/calculateRentalCost.js
VinnieJ-2k20: Switch with `true` works, but it should not be used this way.
It contradicts the logic of `switch/case` purpose where only one case should be true.

VinnieJ-2k20:
> > Some edits needed.
> > Also, reusable data like costPerDay, discounts etc should also be moved to variables.
> 
> Can you clarify? If I use value just one time I should use variables for this value too? Because, at now, I use "costPerDay" and "discounts" values just one time.
> 
> Or we should use variables for all values? 3 days of rent and 7 days of rent – also should be variables?

Yes, in this particular task discounts are used only once, but if some time later your calculation will become more complicated variables will come in handy.

Also, for someone looking at your code line `return totalPrice - 50` may not be so understandable.
Line `return totalPrice - bigDiscount` shows the business logic of your calculation much more clearly.

VinnieJ-2k20 (approved):
Good job 👍

PR #577, opened 2021-04-21
solution

tanyakovchuk:


olha-tsarova (approved):
Nice!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const littleDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return dayPrice * days;
+  } else if (days >= 3 && days <= 6) {
```
VinnieJ-2k20, line 37: If you write `return` inside `if` block, the next `else` is useless.

```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```

VinnieJ-2k20, line 37: `days >= 3` check is useless here, we already passed it in line 35.

```suggestion
  if (days < 3) {
    return dayPrice * days;
  } else if (days <= 6) {
```

VinnieJ-2k20 (changes requested):
Almost done, some edits needed.

VinnieJ-2k20 (approved):
Good job 👍

PR #578, opened 2021-04-21
Solution

olha-tsarova:


tanyakovchuk (approved):
cool! 👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
+  } else if (days >= 3) {
```
VinnieJ-2k20, line 33: If you write `return` inside `if` block, the next `else` is useless.

```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```

VinnieJ-2k20 (changes requested):
Some edits needed.

Also, use variables for reusable data like daily cost, discounts etc.

VinnieJ-2k20 (approved):
I'll approve cause it works correctly.

For the future - if you have repetitive calculations (in this case - `days * price`) you also should move it to variable and calculate only once.

PR #580, opened 2021-04-21
solution

ched88xtc:


Grygoriy-Shytikov (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 1) {
+    return 'Invalid data';
+  } else if (days < 3) {
+    return currentRentalCost;
+  } else if (days < 7) {
+    return currentRentalCost - 20;
+  } else {
+    return currentRentalCost - 50;
+  }
```
aionova, line 41:
```suggestion
  }
  
  return currentRentalCost - 50;
```

aionova (changes requested):
Let's use variables for all the numbers. Don't use magic numbers

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const currentRentalCost = days * 40;
```
aionova, line 31: you can also add costPerDay variable

aionova (approved):
Great!

PR #581, opened 2021-04-21
Solution

BoikoEva:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let cost = 0;
+
+  if (days >= 3 && days <= 6) {
+    cost = (days * 40) - 20;
+  } else if (days < 3) {
+    cost = days * 40;
+  } else if (days >= 7) {
```
aionova, line 38: Do you need 'else if' here or maybe it can be just 'else'?

aionova (changes requested):
Let's not use magic numbers. Put all the numbers in variables

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let rentCost = 0;
+  const cost = 40;
```
aionova, line 33:
```suggestion
  const costPerDay = 40;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+  const off20$ = 20;
```
aionova, line 34:
```suggestion
  const discountSmall = 20;
```

aionova (approved):
Good job!

PR #582, opened 2021-04-21
Solution

MinchukO:


aliceblunt8 (changes requested):
nice start ;)

- let's create variables for different discounts
- also, you've calculate (40 * days) 3 times inside the function, let's put this to a varible at the beginning
- no need for the last if, it can be default case

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const cost = 40 * days;
+
+  if (days < 3) {
+    return cost;
+  }
+
+  if (days >= 3 && days < 7) {
+    const discountMore3Days = 20;
```
aliceblunt8, line 39: move this variables to the top of the file (right after cost)

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #584, opened 2021-04-21
Solution

bohdanklius:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const totalPrice = days * 40;
+
+  if (days < 3) {
+    return totalPrice;
+  }
+
+  if (days >= 7) {
+    return totalPrice - 50;
+  } else {
+    return totalPrice - 20;
```
chumasov, line 40: It would be easier to reuse code if You will move all numbers to variables.

chumasov (approved):

src/calculateRentalCost.js
aliceblunt8: create variables for discounts

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #587, opened 2021-04-21
done

cheytac200:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days >= 3 && days >= 7) {
+    return (days * 40) - 50;
+  } else {
+    return (days * 40) - 20;
+  }
```
aionova, line 39:
```suggestion
  }
    
  return (days * 40) - 20;
```

aionova (changes requested):
Let's use variables for all the numbers

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days >= 3 && days >= 7) {
+    rent = (days * 40) - 50;
+
+    return rent;
+  } else {
+    rent = (days * 40) - 20;
+
+    return rent;
+  }
```
aionova, line 47:
```suggestion
  }
    rent = (days * 40) - 20;

    return rent;
```

aionova (approved):
What I meant is to use variables for all the numbers in this task (3, 7, 20, 40, 50).
You can return rent only once in the end or return it inside each if block

PR #590, opened 2021-04-21
Solution

progmari:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
illisogori, line 32: 40 можно вынести в переменную.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+  } else if (days < 7) {
+    return (days * 40) - 20;
```
illisogori, line 34: 20 можно вынести в переменную.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+  } else {
+    return (days * 40) - 50;
```
illisogori, line 36: 50 можно вынести в переменную.

ghost (approved):
Good job!

VinnieJ-2k20 (changes requested):
If you write `return` inside `if` block, the next `else` is useless.

```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```
Also, Igor is right - it's better to use variables for reusable data.
And if you make some calculation multiple times, like `days * 40` here, it should be used as a variable and used just once.

VinnieJ-2k20 (approved):
It works correctly, so I approve.

But for the future - it's better to use variables for reusable data like discounts.

PR #592, opened 2021-04-21
Solution

GoldenSpade:
[Done](https://github.com/BilliZane/js_task-transportation-on-vacation/blob/develop/src/calculateRentalCost.js)

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  let discount = 0;
+
+  if (days >= 7) {
+    discount = 50;
+  } else if (days >= 3) {
+    discount = 20;
+  }
+
+  return (days * pricePerDay) - discount;
 }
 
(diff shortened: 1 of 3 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  let discount = 0;
+
+  if (days >= 7) {
+    discount = 50;
```
illisogori, line 35: можно вынести в переменную.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  } else if (days >= 3) {
+    discount = 20;
```
illisogori, line 37: можно вынести в переменную.

illisogori (approved):
Отлично!

VinnieJ-2k20 (approved):
I approve, because it works correctly.

But for the future, Igor is right. 
Discounts are better used in variables here.
That shows the logic of your calculations much more clearly.

PR #593, opened 2021-04-21
solution

ivan-shpynda:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days * 40 >= 280) {
+    return (days * 40) - 50;
+  } else if (days * 40 >= 120) {
```
VinnieJ-2k20, line 33: If you write `return` inside `if` block, using `else` after that doesn't make sense.

```
if (condition1) {
  return x;
}

if (condition2) {
  return y
}

return z;
```

VinnieJ-2k20 (changes requested):
Some edits needed.

If you have calculation that is used multiple times (like `days * 40` here), move it to variable and execute only once
Reusable data, like daily cost or discounts should also be used with variables.

src/calculateRentalCost.js
ivan-shpynda: thank you for your remarks. I've fixed the code

VinnieJ-2k20 (approved):
`days * price` is calculated three times. It also should be moved to a variable.
For now, i'll approve.
Remember about it for future tasks.

PR #594, opened 2021-04-21
Solution

TimurHrynenko:


aionova (changes requested):
3, 7, 20 and 50 are magic numbers. It's better to put them in variables

aionova (approved):
LGTM!

PR #597, opened 2021-04-22
solution

C-y-y-y:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,26 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  let count = 0;
+  const price = 40;
+  const maxOff = 50;
+  const minOff = 20;
```
aliceblunt8, line 36:
```suggestion
  const maxDiscount = 50;
  const minDiscount = 20;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,26 @@
...
+
+  if (days >= 7) {
+    count += sum - maxOff;
+  }
+
+  if (days < 3) {
+    count += sum;
+  }
+
+  return count;
```
aliceblunt8, line 51: no need to create count variable, just return the result straight away

aliceblunt8: also you don't need 3 if, try to use only 2 if and then return the default value

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #599, opened 2021-04-23
Solution

MykolaButylkov:


aionova (approved):
You need to put all the numbers in variables.

PR #600, opened 2021-04-25
Solution

stetsyshyn:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
+    return days * 40 - 20;
+  }
+
+  return days * 40;
```
vkryvytskyy, line 39: all is good, just make variables for all numeric values

vkryvytskyi (changes requested):

src/calculateRentalCost.js
stetsyshyn: Thanks. 
Done.

stetsyshyn:
Hi, 
Do I have to do anything else to approve the commit?
Thanks,

PR #672, opened 2021-06-22
Solution

Kandio16:


Liuda-Yevchuck (approved):
Everything's okay, as for me) I like that u used variables for all values

Kandio16:
Maybe use a new variable)

solaryasha (approved):

PR #735, opened 2021-08-01
Solution

ivanchenko-kirill:


Andrii-Pinchuk (approved):
Good work! 👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const bigDiscount = 50;
+  let total = days * perDayCost;
+
+  if (days >= 3 && days < 7) {
+    total -= smallDiscount;
+  }
+
+  if (days >= 7) {
+    total -= bigDiscount;
+  }
```
```suggestion
  if (days >= 7) {
    total -= bigDiscount;
  } else if (days >= 3) {
    total -= smallDiscount;
  }
```

You can switch these statements by place and get rid of one redundant check.

varseniuk (changes requested):
Good job. Need one small fix

varseniuk (approved):

PR #739, opened 2021-08-02
Task transportation on vacation

UlaDutchak:


alex-glotov67 (approved):
Looks good!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+  const payPerDay = 40;
+  const discountAfterThreeDays = 20;
+  const discountAfterSevenDays = 50;
+
+  if (days < 3) {
+    result = days * payPerDay;
+  } else if (days >= 3 && days < 7) {
```
varseniuk, line 38:
```suggestion
  } else if (days < 7) {
```
The first check is not necessary. If first case `(days < 3)` won't pass it will mean that days >=3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+    result = (days * payPerDay) - discountAfterThreeDays;
+  } else {
+    result = (days * payPerDay) - discountAfterSevenDays;
```
```suggestion
  }
    result = (days * payPerDay) - discountAfterSevenDays;
```

You can get rid of this `else` here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  }
+
+  return result;
```
varseniuk, line 44: You can return result right inside your `if` statements

varseniuk (approved):
Good. Left some comments

PR #744, opened 2021-08-02
Solution

maximbalyk:


Anna-Tkachenko (approved):
👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
+  } else if (days >= 3) {
+    return (days * 40) - 20;
+  } else {
+    return days * 40;
```
```suggestion
  }
    return days * 40;
```
No need in this else. This case will be returned automatically if both previous ones will fail

varseniuk (approved):
Good.
You could create some variable for `(days * 40)` not to repeat it three times

src/calculateRentalCost.js
maximbalyk: fixed it !)

PR #745, opened 2021-08-02
Solution

denyschemeriakov:


maximbalyk (approved):
good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,9 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = days * 40;
```
varseniuk, line 31:
```suggestion
  const totalCost = days * 40;
```
What is 40? Create variable for this

src/calculateRentalCost.js
```diff
@@ -28,7 +28,9 @@
...
+
+  return days >= 7 ? totalCost - 50 : days >= 3 ? totalCost - 20 : totalCost;
```
varseniuk, line 33:
```suggestion
  return days >= 7 ? totalCost - 50 : days >= 3 ? totalCost - 20 : totalCost;
```
Don't use ternary operators here. Use if-else if

Also, create variables for 50 and 20. It's not clear what are these numbers

varseniuk (changes requested):

varseniuk (approved):
Left one comment

PR #749, opened 2021-08-02
tasks was done

Crystallord:


maximbalyk (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let cost = days * 40;
+
+  if (days >= 3 && days < 7) {
+    cost -= 20;
+  }
+
+  if (days >= 7) {
+    cost -= 50;
+  }
```
```suggestion
  if (days >= 7) {
    cost -= 50;
  } else if (days >= 3) {
    cost -= 20;
  }
```
Change these `if`s by places and you will get rid of one redundant check

varseniuk (changes requested):
Let's create variables for numbers 50 and 20. It's not clear what they should mean.

varseniuk (approved):
👍

PR #751, opened 2021-08-02
Solution

alex-glotov67:


UlaDutchak (approved):
👍

varseniuk (changes requested):
Use `if-else if` construction. Also, create variables for numbers 20 and 50. It's not clear now what they should mean

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const rentCost = 40;
+  const discountFromOneToSevenDay = 20;
+  const discountFromSevenDays = 50;
+
+  if (days < 3) {
+    return days * rentCost;
+  } else if (days < 7) {
+    return (days * rentCost) - discountFromOneToSevenDay;
+  } else if (days >= 7) {
+    return (days * rentCost) - discountFromSevenDays;
```
```suggestion
  }
    return (days * rentCost) - discountFromSevenDays;
```
The last `else` is not necessary here

varseniuk (approved):
Good. Left one comment

PR #757, opened 2021-08-02
Develop

ihor-jpeg:


DmytroMykoliv (approved):

Ihor-Karpyn (changes requested):
A smart decision 🥸 but not very intuitive 🙄, though. This code will be difficult to maintain in the future. Don't make other developers think a lot. Simplify it.

Ihor-Karpyn (approved):
Great 👌

PR #761, opened 2021-08-02
solution

00-22-11:


RomanFedusevych (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days > 6) {
```
varseniuk, line 33:
```suggestion
  } else if (days >= 7) {
```
It's more meaningful

varseniuk (changes requested):
Create variables for numbers 40, 50, 20. Now it's not clear what they should mean

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const rentalCarCost = 40;
+  const discountMoreThen7days = 50;
+  const discountMoreThen3days = 20;
+
+  if (days <= 2) {
+    return days * rentalCarCost;
+  } else if (days >= 7) {
+    return days * rentalCarCost - discountMoreThen7days;
+  } else {
+    return days * rentalCarCost - discountMoreThen3days;
```
```suggestion
  }
    return days * rentalCarCost - discountMoreThen3days;
```
This `else` is not necessary here

varseniuk (approved):
Left one comment

PR #768, opened 2021-08-03
checking the task by a mentor

Andrii-Kalaida:


Serhii-Usenko (commented):
good work)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const smallCashBack = 20;
+  const largeCashBack = 50;
+  const priceDayRentCar = 40;
+  const wholeAmount = days * priceDayRentCar;
+  let priceRentCar = 0;
```
varseniuk, line 35:
```suggestion
  let priceRentCar;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  if (days >= 3) {
+    priceRentCar = wholeAmount - smallCashBack;
+  }
+
+  if (days < 3) {
+    priceRentCar = wholeAmount;
+  }
+
+  if (days >= 7) {
+    priceRentCar = wholeAmount - largeCashBack;
```
varseniuk, line 46: Let's reduce the amount of checks. Use only two checks:
if (days < 3) and if (days <=7) or if (days >=7) and if (days >=3)
(in that order)

Also, use `if - else if` construction instead of just many `if`s

varseniuk (changes requested):
All is good, but need small fixes

src/calculateRentalCost.js
```diff
@@ -28,7 +29,18 @@
...
+  const priceDayRentCar = 40;
+  const wholeAmount = days * priceDayRentCar;
+
+  if (days >= 7) {
+    return wholeAmount - largeCashBack;
+  } else if (days >= 3) {
+    return wholeAmount - smallCashBack;
+  }
+
+  return days * priceDayRentCar;
```
```suggestion
  return wholeAmount;
```
You have a variable for this

varseniuk (approved):
Left one comment

PR #771, opened 2021-08-03
first commit

VVHimiak:


RomanFedusevych (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const manyDaysOff = 50;
+  const fewDaysOff = 20;
+  const manyDaysCount = 7;
+  const fewDaysCount = 3;
+
+  if (days >= manyDaysCount) {
+    return (days * perDayCost - manyDaysOff);
+  }
+
+  if (days >= fewDaysCount && days < manyDaysCount) {
```
```suggestion
  else if (days >= fewDaysCount) {
```
This check is redundant.
Also, use `else if` construction

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return (days * perDayCost - fewDaysOff);
+  } else {
+    return (days * perDayCost);
+  }
```
varseniuk, line 45:
```suggestion
  }
    return (days * perDayCost);
```
This `else` is not necessary here

varseniuk (changes requested):
Some small fixes left

varseniuk (approved):

PR #773, opened 2021-08-03
Solution

MaksymBeliever:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const carRentalCost = 40;
+  const minimumDiscount = 20;
+  const maximumDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return days * carRentalCost - minimumDiscount;
+  }
```
TarasHoliuk, line 37: a mentor recommended use `else if`

TarasHoliuk (approved):
good job!

src/calculateRentalCost.js
MaksymBeliever: Ok, I will use it in the next tasks

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 7) {
+    return days * carRentalCost - maximumDiscount;
+  }
```
```suggestion
  if (days >= 7) {
    return days * carRentalCost - maximumDiscount;
  } else if (days >= 3) {
    return days * carRentalCost - minimumDiscount;
  }
```
If you swap these statements vice versa you can get rid of one redundant check.

varseniuk (changes requested):
Some small changes left

varseniuk (approved):

PR #827, opened 2021-09-13
Solution

caelum29:


Ihor-Karpyn (changes requested):
St review required  ⚠️

just-a-vlad (approved):
Хорошая работа)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+
+  if (days < 3) {
+    return price * days;
```
Anna-Tkachenko, line 34: Don't repeat yourself (days * price), create a separate variable for this

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days < 7) {
+    return days * price - 20;
+  } else {
+    return days * price - 50;
+  }
```
Anna-Tkachenko, line 39: Last else is redundant here
```suggestion
  return days * price - 50;
```

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  const price = 40;
+  const totalPrice = days * price;
+
+  if (days < 3) {
+    return totalPrice;
+  } else if (days < 7) {
+    return totalPrice - 20;
+  } else {
+    return totalPrice - 50;
+  }
```
Anna-Tkachenko, line 40: Last else is redundant here
```suggestion
  return totalPrice - 50;
```

Anna-Tkachenko (approved):
Good job👍

PR #839, opened 2021-09-14
Solution

Yashkindzhus:


Bogdansgit:
Nice job!

Bogdansgit (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,13 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
```
Anna-Tkachenko, line 33: Don't repeat yourself. Create a separate variable for days * 40

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
   // write code here
+  const rental = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return rental - 20;
+  } else if (days >= 7) {
+    return rental - 50;
+  } else {
+    return rental;
+  }
```
Anna-Tkachenko, line 40: You can remove last else:
```suggestion
  return rental;
```

Anna-Tkachenko (approved):

PR #840, opened 2021-09-14
solution

AVronskyi:


sTorba24 (approved):
Great!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let off = 0;
```
Anna-Tkachenko, line 31: Let's create more meaningful name (offCost or discount)
```suggestion
  let offCost = 0;
```

Anna-Tkachenko (approved):

PR #841, opened 2021-09-14
solution

vlad-shulzhenko:


MarusaLoza (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let amount = 0;
+
+  if (days <= 2) {
+    amount = 40 * days;
```
Anna-Tkachenko, line 35: Don't repeat yourself (40 * days), create a separate variable for this

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #842, opened 2021-09-14
solution

sTorba24:


AVronskyi (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * 40;
```
Anna-Tkachenko, line 31: Let's create more meaningful name for this variable:
```suggestion
  const totalCost = days * 40;
```

Anna-Tkachenko (approved):

PR #850, opened 2021-09-15
Solution

DY-JS:


KatyKuzmenko (approved):
Good job!)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const price = 40;
+
+  if (days >= 7) {
+    return price * days - 50;
```
Anna-Tkachenko, line 34: Don't repeat yourself, let's create a separate variable for price * days

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  }
 
+  if (days >= 3) {
+    return price * days - 20;
+  } else {
+    return price * days;
+  }
```
Anna-Tkachenko, line 41: Last else is redundant here:
```suggestion
  return price * days;
```

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const price = 40;
+
```
aliceblunt8, line 32:
```suggestion
const cost = price * days;
```

src/calculateRentalCost.js
aliceblunt8:
```suggestion
    return cost - 50;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  if (days >= 7) {
+    return price * days - 50;
+  }
 
+  if (days >= 3) {
+    return price * days - 20;
```
aliceblunt8, line 38:
```suggestion
    return cost - 20;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  }
+
+  return price * days;
```
aliceblunt8, line 41:
```suggestion
  return cost;
```

aliceblunt8 (approved):

PR #868, opened 2021-10-04
solution

HaqpaHR:


Igor-MateAcademy (approved):

solaryasha (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+  const maxDiscount = 50;
+  const minDiscount = 20;
+
+  if (days >= maxDaysDiscount) {
+    return (days * rentPrice) - maxDiscount;
+  } else if (days >= minDaysDiscount) {
+    return (days * rentPrice) - minDiscount;
+  } else if (days >= withoutDiscount) {
+    return days * rentPrice;
+  }
```
VinnieJ-2k20, line 45: `else` is useless with `return` inside `if` block.
Also, you calculate `days * rentPrice` three times. Use variable for it.
```suggestion
  if (days >= maxDaysDiscount) {
    return (days * rentPrice) - maxDiscount;
  } 
  
  if (days >= minDaysDiscount) {
    return (days * rentPrice) - minDiscount;
  }
  
  return days * rentPrice;
```

VinnieJ-2k20 (changes requested):
Almost done.
Fix if structure and some variables.

VinnieJ-2k20 (approved):
Good job 👍

PR #879, opened 2021-10-04
Solution

Linetstg:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
```
Olexandr6, line 31: You should create variables for:

1. cost renting per day(40$)
2. discounts
3. days for both discounts(3, 7)

Now it simple program, but in the future if it will be more complicated and you just want to change for example: days for discounts, you have to change it all places where you used it, but with variable you just can change it in one place, where you initialize. So, you should use variables.

G-YaSha (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return days * costPerDay;
+  }
+
+  if (days < discountDayBig) {
+    return days * costPerDay - discountLittle;
+  }
+
+  if (days >= discountDayBig) {
+    return days * costPerDay - discountBig;
+  }
```
VinnieJ-2k20, line 47: You calculate `days * costPerDay` three times. Do it once and use variable.

VinnieJ-2k20 (changes requested):
Almost done.

VinnieJ-2k20 (approved):
Good job 👍

PR #880, opened 2021-10-04
solution

Aleksey060790:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const daysForMinDiscount = 3;
+  const daysForMaxDiscount = 7;
+  const discountPerThreeDays = 20;
```
Olexandr6, line 34: I would rename it
```suggestion
  const minDiscount = 20;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const discountPerWeek = 50;
```
Olexandr6, line 35:
```suggestion
  const maxDiscount = 50;
```

src/calculateRentalCost.js
Olexandr6: 'Cause you already have var, that discribes "how many days for min/max discount".

G-YaSha (approved):
Good job👍 But check comments👇

src/calculateRentalCost.js
VinnieJ-2k20: Agree. You don't want to rename your variables if daysForMinDiscount changes to 4.

src/calculateRentalCost.js
VinnieJ-2k20: Same here.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const rentalPriceWithoutDiscount = days * costPerDay;
+
+  if (days >= daysForMaxDiscount) {
+    return rentalPriceWithoutDiscount - discountPerWeek;
+  } else if (days >= daysForMinDiscount) {
+    return rentalPriceWithoutDiscount - discountPerThreeDays;
+  } else {
+    return rentalPriceWithoutDiscount;
+  }
```
VinnieJ-2k20, line 44: `else` with `return` inside `if` block is useless.

```suggestion
  if (days >= daysForMaxDiscount) {
    return rentalPriceWithoutDiscount - discountPerWeek;
  } 
  
  if (days >= daysForMinDiscount) {
    return rentalPriceWithoutDiscount - discountPerThreeDays;
  } 
  
  return rentalPriceWithoutDiscount;
```

VinnieJ-2k20 (changes requested):
Almost done. Fix some varaibles names and `if` structure.

VinnieJ-2k20 (approved):
Good job 👍

PR #883, opened 2021-10-05
Solution

Anton-Karpena:


DionStrzh (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return rentCost * days;
+  }
+
+  if (days >= daysCountMinDiscount && days < daysCountMaxDiscount) {
+    return rentCost * days - minDiscount;
+  }
+
+  if (days >= daysCountMaxDiscount) {
+    return rentCost * days - maxDiscount;
+  }
```
VinnieJ-2k20, line 47: Don't duplicate conditions. If previous condition is `false`, the opposite is always `true`.
And last condition is useless in this case.

```suggestion
  if (days < daysCountMaxDiscount) {
    return rentCost * days - minDiscount;
  }
    
  return rentCost * days - maxDiscount;
```

VinnieJ-2k20 (changes requested):
Almost done, fix conditions.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const daysCountMinDiscount = 3;
+  const daysCountMaxDiscount = 7;
+
+  if (days < daysCountMinDiscount) {
+    return rentCost * days;
+  }
+
+  if (days >= daysCountMinDiscount && days < daysCountMaxDiscount) {
+    return rentCost * days - minDiscount;
+  }
```
VinnieJ-2k20, line 43: Same. duplicated conditions.
```suggestion
  if (days < daysCountMaxDiscount) {
    return rentCost * days - minDiscount;
  }
```

VinnieJ-2k20 (approved):
Good job 👍

PR #938, opened 2021-11-08
Car rent solution

Dima-Ziakhor:


lera-permiakova (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
+
+  for (let i = 0; i < days; i++) {
+    sum += 40;
+  }
```
AlinaMatuschak, line 35: Don't add unnecessary steps. You can multiply days and costs )

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 7) {
+    return sum - 50;
```
AlinaMatuschak, line 38: Move all the magic numbers 40, 50, 20 to constants

AlinaMatuschak (changes requested):
Nice work 😉. But fix small issues:

AlinaMatuschak (approved):
🚀

PR #940, opened 2021-11-08
Calculate my vacation rent for a car

dmtrhrytsak:


ImIkator (approved):
Nice implemented solution !

src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPrice = 40;
+  const threeDaysDiscount = 20;
+  const weekDiscount = 50;
+
+  const getDiscount = () =>
+    days >= 7 ? weekDiscount : days >= 3 ? threeDaysDiscount : 0;
```
illiarerun, line 36: Nested ternaries is a bad practice, if you have more than 1 conditions, it'd be better to use `if`'s. Also even if you use ternary, you should properly format it

illiarerun (changes requested):
Great start 🔥

illiarerun (approved):
LGTM 🔥

PR #943, opened 2021-11-08
Added solution

Vitalii-Lytvyn01:


alex-kostomarov (approved):
Looks good!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
illiarerun, line 32: use variables for the main values (40, 20, 50) so that you can reuse them and give them descriptive names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  }
+
+  if (days >= 3) {
+    return days * 40 - 20;
+  }
+
+  return days * 40;
```
illiarerun, line 39: Follow DRY, move that calculation to a variable

illiarerun (changes requested):
Good start 🔥

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+    return result;
+  }
+
+  if (days >= 3) {
+    result = days * price - smallDiscount;
+
+    return result;
+  }
+  result = days * price;
```
illiarerun, line 47: I've meant that you can move this calculation to a const before all the checks, and use it in return statements

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
```
illiarerun, line 31: Also, be more specific with variable names
```suggestion
  const dailyPrice = 40;
```

illiarerun (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyPrice = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  let result = dailyPrice * days;
+
+  if (days >= 7) {
+    result = days * dailyPrice - bigDiscount;
```
illiarerun, line 37: and return value here, without reassigning result value

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyPrice = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  let result = dailyPrice * days;
```
illiarerun, line 34: Again, I've meant that you can store this as const

illiarerun (changes requested):

src/calculateRentalCost.js
Vitalii-Lytvyn01: Oh, like that?
```suggestion
   return result = days * dailyPrice - bigDiscount;
```

Vitalii-Lytvyn01: I think i understood wrong, i'll try another thing.

Vitalii-Lytvyn01:
```suggestion
    return days * dailyPrice - smallDiscount;
```
or something like this

Vitalii-Lytvyn01: like if we have a constant we cant change it, but we have to in order to make correct result.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyPrice = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const result = dailyPrice * days;
+
+  if (days >= 7) {
+    return days * dailyPrice - bigDiscount;
```
illiarerun, line 37: Look, you could've used your `result` variable here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * dailyPrice - smallDiscount;
```
illiarerun, line 41: And here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyPrice = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const result = dailyPrice * days;
```
illiarerun, line 34: But also, result is a bit confusing name of variable, proper variant would be something like `totalPrice`, as I commented above

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 7) {
+    return days * dailyPrice - bigDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
```
illiarerun, line 40: And you don't need this additional check, cause you check the opposite of this condition in previous `if` clause
```suggestion
  if (days >= 3) {
```

illiarerun (changes requested):

illiarerun (approved):
LGTM 🔥

PR #944, opened 2021-11-08
Solution

alex-kostomarov:


Nadiyahr (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
illiarerun, line 32: use variables for the main values (20, 40, 50) so that you can reuse them and give them descriptive names.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+  } else if (days >= 3) {
+    return days * 40 - 20;
+  } else {
+    return days * 40;
```
illiarerun, line 36: Move this calculation to variable, don't repeat yourself

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
```
illiarerun, line 33: don't use else after return statement. If return is performed - code after it won't be executed anyway.

illiarerun (changes requested):
Good start 🔥

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const paymentPerDay = 40;
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
+
+  let totalPrice = paymentPerDay * days;
```
illiarerun, line 35: You don't really need to change the value of this calculation, add a return statements like `return totalPrice - discount'
```suggestion
  const totalPrice = paymentPerDay * days;
```

illiarerun (changes requested):

illiarerun (changes requested):
.

illiarerun (approved):
Great job 🔥

PR #945, opened 2021-11-08
solution

Nadiyahr:


dariia-romanova (approved):

illiarerun (changes requested):
Good start 🔥

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const half = 20;
+  const week = 50;
+
+  if (days >= 7) {
+    return cost - week;
+  }
+
+  if (days >= 3) {
+    return cost - half;
+  } else {
```
illiarerun, line 41: Again, you don't need to use `else` after return statement. If return is performed - code after it won't be executed anyway.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
+  const half = 20;
+  const week = 50;
```
illiarerun, line 33: What are `cost`, `half` and `week`, be more specific with your variable names
```suggestion
  const totalDailyCost = days * 40;
  const smallDiscount = 20;
  const bigDiscount  = 50;
```

illiarerun (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+    return totalDailyCost - smallDiscount;
+  }
+
+  return totalDailyCost;
+
+  // return days >= 7
+  //   ? totalDailyCost - bigDiscount
+  //   : days >= 3
+  //     ? totalDailyCost - smallDiscount
+  //     : totalDailyCost;
```
illiarerun, line 49: Don't leave commented code like that
```suggestion
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalDailyCost = days * 40;
```
illiarerun, line 31: Again, add a variable for this `40` value

illiarerun (changes requested):

illiarerun (approved):
LGTM 🔥

PR #946, opened 2021-11-08
First solution

WorkAccCom:


lanebx (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
```
AlinaMatuschak, line 32: Move all the magic numbers 40, 50, 20 to constants

AlinaMatuschak (changes requested):
Good start 👌

src/calculateRentalCost.js
WorkAccCom: magic numbers 😄

AlinaMatuschak (approved):

PR #950, opened 2021-11-08
Solution

Val-space:


SerhiiKirik (approved):
Well done 🔥

dmtrhrytsak (approved):
Works right, but you could get rid of all magic variables, hence making your code more clear and maintainable 

image

I mean just assigning all numbers to some named variables)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
illiarerun, line 32: use variables for the main values (20, 40, 50) so that you can reuse them and give them descriptive names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+  } else if (days >= 3) {
+    return days * 40 - 20;
+  } else {
+    return days * 40;
```
illiarerun, line 36: Follow DRY, move that calculation to a variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
```
illiarerun, line 33: don't use else after return statement. If return is performed - code after it won't be executed anyway.

illiarerun (changes requested):
Good job 🔥

illiarerun (approved):
LGTM 🔥

PR #951, opened 2021-11-08
Solution

SerhiiKirik:


junglq:
very good

dmtrhrytsak:
image

Your code is concise, but there are some advantages to putting some values in variables:) Think about that

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPrice = 40;
+  const threeDaysDiscount = 20;
+  const weekDiscount = 50;
+  const allDays = days * rentPrice;
+
+  return days >= 7 ? allDays - weekDiscount : days >= 3
+    ? allDays - threeDaysDiscount : allDays;
```
illiarerun, line 37: Nested ternaries is a really bad practice. It'd be better to use simple `if`'s if you have multiple conditions

illiarerun (changes requested):
Great start 🔥

illiarerun (approved):
LGTM 🔥

PR #957, opened 2021-11-08
Solution

dlugash:


Val-space (approved):
Good! You could assign numbers to variables for better readability

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
+    return days * 40 - 20;
+  } else {
```
illiarerun, line 35: There's no need in `else` if you return a value in `if` block

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
illiarerun, line 32: What are 40 and 50 and 20? You should create constants for such values

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+  } else if (days >= 3) {
+    return days * 40 - 20;
+  } else {
+    return days * 40;
```
illiarerun, line 36: Also you repeat yourself a lot with a `day * 40` calculation, too, make a variable for that

illiarerun (changes requested):
Good start 🔥

illiarerun (approved):
LGTM 🔥

PR #959, opened 2021-11-08
Solution

alinaShchavinska:


IrynaHerasymchuk (approved):
You should create a variable for total so your program will calculate days * 40 only once

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return days * carPrice;
+  }
+
+  if (days > 6) {
+    return days * carPrice - bigDiscount;
+  }
+
+  if (days > 2) {
+    return days * carPrice - smallDiscount;
+  }
```
AlinaMatuschak, line 45: you can delete the last condition. It will default cancellation: if the number is bigger than 2 (first condition false) and smaller than 7 (second condition false))

```suggestion
  if (days < 3) {
    return days * carPrice;
  }

  if (days > 6) {
    return days * carPrice - bigDiscount;
  }

  return days * carPrice - smallDiscount;
```

AlinaMatuschak (approved):
Good work 👌

PR #960, opened 2021-11-08
solution

oleksiipochkaiev:


IrynaHerasymchuk (approved):
You should create a variable for total so your program would count days * 40 only once

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
AlinaMatuschak, line 32: Move all the magic numbers 40, 50, 20 to constants

AlinaMatuschak (changes requested):
Almost perfectly 👌

AlinaMatuschak (approved):

PR #961, opened 2021-11-08
Solution

IrynaHerasymchuk:


alinaShchavinska (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+
+  const total = price * days;
+
+  if (days >= 7) {
+    return total - bigDiscount;
+  } else if (days >= 3) {
+    return total - smallDiscount;
+  }
```
illiarerun, line 41: There is no need to use `else` if you return a value from previous `if` clause

illiarerun (changes requested):
Great start 🔥

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
```
illiarerun, line 31: Be more specific with variable names
```suggestion
  const dailyPrice = 40;
```

illiarerun (approved):
LGTM 🔥

PR #967, opened 2021-11-08
Solution

andriibulbuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
+    return days * 40 - 20;
+  }
+
+  return days * 40;
```
Veronika-Korona, line 39: Good job! Now just add variables for all the magic numbers (40, 50, 20).

Veronika-Korona (changes requested):
Good job!

src/calculateRentalCost.js
illiarerun: Agree

illiarerun: Also you can move do this calculation before all the checks to not repeat yourself in every return statement

illiarerun (changes requested):
Grat start 🔥

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discountMoreThreeDays = 20;
+  const discountMoreSevenDays = 50;
+  const rentalCost = 40;
+
+  let price = days * rentalCost;
```
illiarerun, line 35: There is no need in `let` variable here, use `const` instead you can just put return statements in your if blocks and use this variable right in return

illiarerun (changes requested):

illiarerun (approved):
LGTM 🔥

PR #986, opened 2021-11-29
Solution

pavlo-vovk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  if (days < 3) {
```
tnboyko, line 31: it is better to put all the numbers into variables

tnboyko (approved):

PR #996, opened 2021-12-10
Solution

yulyapushyk:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const rentalCost = days * 40;
+
+  if (days >= 1 && days <= 2) {
```
aliceblunt8, line 34:
```suggestion
  if (days < 3) {
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+    return rentalCost;
+  }
+
+  if (days >= 3 && days <= 6) {
```
aliceblunt8, line 38:
```suggestion
  if (days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+    return rentalCost - 20;
+  }
+
+  if (days >= 7) {
+    return rentalCost - 50;
+  }
+
+  return 0;
```
aliceblunt8, line 46:
```suggestion
    return rentalCost - 50;
```

aliceblunt8 (approved):
good, but you could write less logic for achieveing the same result

PR #1007, opened 2021-12-13
Solution

Sesile26:


mkyloa (approved):

G-YaSha (changes requested):
Create separate vars for all numbers in your solution, don't harcode

YevheniiKorzun (approved):
Well done!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const amountPerDay = 40;
+  const discount3Days = 20;
+  const discount7Days = 50;
```
Olexandr6, line 33: Don't use numbers in var's naming, cause if you change the value in the future, you have to change var's name, as well

G-YaSha (approved):
Good job 👍

PR #1008, opened 2021-12-13
solution

luckdenny:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return 0;
+  }
+
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days < 7 && days >= 3) {
+    return days * 40 - 20;
+  } else {
```
MaksimusCRS, line 41: 3. [CODE STYLE] - You should return the value immediately in `if` statements. And don't use `else` after return statement. If `return` is performed - code after it won't be executed anyway.

MaksimusCRS (changes requested):

MaksimusCRS (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 1) {
+    return 0;
+  }
+
+  if (days < 3) {
+    return days * 40;
```
Olexandr6, line 36: You use the expression: `(days * 40`) in other places, so let's create a separate variable for this.

G-YaSha (changes requested):
Create separate vars for all numbers in your solution, don't hardcode

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const amount = days * 40;
```
```suggestion
  const totalRentPrice = days * 40;
```

G-YaSha (changes requested):
***Create separate vars for all numbers in your solution, don't hardcode
Check 5 item in*** [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

G-YaSha (changes requested):
You should ***Create separate vars for all numbers in your solution, don't hardcode them
Check 5 item in*** [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

***It means you should create separate variables for numbers `3, 7, 20, 50`***

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+  if (days < 1) {
+    return 0;
+  }
+
+  if (days < lowDiscontDuration) {
+    return totalRentPrice;
+  }
+
+  if (days < maxDiscontDuration && days >= lowDiscontDuration) {
```
Olexandr6, line 45:
```suggestion
  if (days < maxDiscontDuration) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalRentPrice = days * 40;
```
```suggestion
  const carRentCost = 40;
  const totalRentPrice = days * carRentCost;
```

src/calculateRentalCost.js
Olexandr6: This condition is unnecessary 👆

G-YaSha (approved):

PR #1009, opened 2021-12-13
My solution

SavelovSerhii:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
+
+  for (let i = 0; i < days; i++) {
+    count += 40;
+  }
```
Sesile26, line 35:
```suggestion
  let count = days * 40;
```
Для чого цикл тут використовувати

Sesile26 (approved):

src/calculateRentalCost.js
Olexandr6: Agree, you don't need cycle here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
```
Olexandr6, line 31: Rename the variables

G-YaSha (changes requested):
For all numbers in your solution, you should create separate values.
5 item in [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentPrice = 40 * days;
```
```suggestion
	const carRentCost = 40;
  let totalRentPrice = carRentCost * days;
```

G-YaSha (approved):
Good job 👍

PR #1010, opened 2021-12-13
rental cost solution

levix2l:
one line solution

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,7 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days >= 7 ? days * 40 - 50 : days >= 3 ? days * 40 - 20 : days * 40;
 }
 
 module.exports = calculateRentalCost;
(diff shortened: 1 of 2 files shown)


mkyloa (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,7 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days >= 7 ? days * 40 - 50 : days >= 3 ? days * 40 - 20 : days * 40;
```
Denchepornyuk, line 31: ternary operator is not the best idea for this
Also one ternary inside another ternary makes code unreadable

It is better to use if conditions here

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
+    return days * 40 - 20;
+  }
+
+  return days * 40;
```
Denchepornyuk, line 39: would be great to create variables for numbers

denys-cheporniuk (approved):

PR #1016, opened 2021-12-13
solution

taras-zherebetskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRISE = 40;
+  const SMALLSALE = 20;
+  const BIGSALE = 50;
+  let total = 0;
```
marina-tilniak, line 34: тут краще зразу написати вираз, щоб лишній раз не переприсвоювати зміннну

marina-tilniak (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRISE = 40;
+  const SMALLSALE = 20;
+  const BIGSALE = 50;
+  const TOTAL = days * PRISE;
+
+  if (days >= 7) {
```
Olexandr6, line 36: For numbers: `3` and `7`, you should create separate vars, as well

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRISE = 40;
+  const SMALLSALE = 20;
+  const BIGSALE = 50;
```
Olexandr6, line 33: In JS we don't use uppercase for all constants. We use uppercase for some specific constants, for example:
`const PI = 3.14...`
`const BASE_URL = https://github.com/`
etc.

G-YaSha (changes requested):
Almost done 👍

G-YaSha (approved):
You've done it 🔥

PR #1025, opened 2021-12-13
solution

Valeria-Valuyskaya:


pasha-melnyk (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
```
Olexandr6, line 32: You use the expression: `(days * 40`) in other places, so let's create a separate variable for this.

G-YaSha (changes requested):
Create separate vars for all numbers in your solution, don't hardcode

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40 * days;
```
```suggestion
  const carRentCost = 40;
  const rentCost = carRentCost * days;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const shortPeriod = 3;
+  const longPeriod = 7;
```
Olexandr6, line 33: Better to show in what units you measure the period 👇
```suggestion
  const shortPeriodDays = 3;
  const longPeriodDays = 7;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const discountSmall = 20;
+  const discountBig = 50;
+
+  if (days < shortPeriod) {
+    return rentCost;
+  }
+
+  if (days >= shortPeriod && days < longPeriod) {
```
Olexandr6, line 41: This condition is too complex, you can simplify it 👇
```suggestion
  if (days < longPeriod) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return rentCost - discountSmall;
+  }
+
+  if (days >= longPeriod) {
+    return rentCost - discountBig;
+  }
```
Olexandr6, line 47: Unnecessary `if` statement
```suggestion

  return rentCost - discountBig;
```

G-YaSha (approved):
Good solution 👍  But your code needs some optimization 👇

PR #1030, opened 2021-12-13
Transportation-task

MVprogramm:


AlexAzalor (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDiscount = 20;
+  const preferDiscount = 50;
+  const dayPrice = 40;
+
+  if (days < 3) {
+    return days * dayPrice;
```
Olexandr6, line 36: You use the expression: `(days * 40`) in other places, so let's create a separate variable for this.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDiscount = 20;
+  const preferDiscount = 50;
+  const dayPrice = 40;
+
+  if (days < 3) {
```
Olexandr6, line 35: For numbers: '3' and '7', you should create separate vars, as well

G-YaSha (changes requested):
Almost done

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days < basicDiscountDay) {
+    return days * dayPrice;
+  };
+
+  if (days < preferDiscountDay) {
+    return days * dayPrice - basicDiscount;
+  };
+
+  return days * dayPrice - preferDiscount;
 }
```
Olexandr6, line 46:
```suggestion
  const totalRentPrice = days * dayPrice;

  if (days < basicDiscountDay) {
    return totalRentPrice;
  };

  if (days < preferDiscountDay) {
    return totalRentPrice - basicDiscount;
  };

  return totalRentPrice - preferDiscount;
}
```

G-YaSha (changes requested):
Almost done 👇

G-YaSha (approved):

PR #1031, opened 2021-12-13
solution

Viktor9r:


orest-mykhailyshyn (approved):
Checklist p.1 use variables for the main values so that you can reuse them and give them descriptive names. 
I think 40 20 50 should be `const`

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
 }
```
Olexandr6, line 43:
```suggestion
function calculateRentalCost(days) {
  if (days >= 3 && days < 7) {
    return days * 40 - 20;
  }

  if (days >= 7) {
    return days * 40 - 50;
  }
  
  return days * 40;
}
```

G-YaSha (changes requested):
For all numbers, create separate vars, don't hardcode
5 item in [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const perDay = 40;
+  const firstSale = 20;
+  const secondSale = 50;
+
+  if (days < 3) {
```
Olexandr6, line 36: For days: `3` and `7`, you should create separate values, as well

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+    return days * perDay;
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * perDay - firstSale;
+  }
+
+  if (days >= 7) {
+    return days * perDay - secondSale;
+  }
```
Olexandr6, line 46: This `if` statement is unnecessary here
```suggestion
  return days * perDay - secondSale;
```

G-YaSha (approved):
Good job 👍

PR #1034, opened 2021-12-13
Solution

gnosis-frog:


orest-mykhailyshyn (approved):
what about const for 20 and 50

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const lowDiscount = 20;
+  const highDiscount = 50;
+  const totalCost = costPerDay * days;
+
+  if (days < 3) {
```
Olexandr6, line 36: For numbers: `3 `and `7` create separate vars

Olexandr6 (approved):

PR #1040, opened 2021-12-13
solution_v1.0

Rozdorozhnii:


bcmp4p1 (approved):

OleksandrOstrovskyi (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
```
Olexandr6, line 32: You use the expression: `(days * 40)` in other places, so let's create a separate variable for this.

G-YaSha (changes requested):
Create separate vars for all numbers in your solution

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const paymentPerDay = 40;
+  const threeDaysDiscount = 20;
+  const weekDiscount = 50;
+  const calcPayment = days * paymentPerDay;
```
Olexandr6, line 34:
```suggestion
  const totalRentPrice = days * paymentPerDay;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 7) {
```
Olexandr6, line 36: For days: `7` and `3` create separate values, as well. 
For example: 
```
const shortPeriodDays = 3
const longPeriodDays = 7
```

Better to show in what units you measure the period. 👆

G-YaSha (changes requested):
Almost done 👍

G-YaSha (approved):

src/calculateRentalCost.js
Rozdorozhnii: маю на увазі розрахункове значення оплати

PR #1041, opened 2021-12-13
Solution

ErnestoFernandezUA:


aleksandra-mileant (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,25 @@
...
+      return total - weekDiscount;
+
+    case days > 2 :
+      return total - threeDaysDiscount;
+
+    case days > 0 :
+      return total;
+
+    case true :
+      return 0;
```
Denchepornyuk, line 49: add `default` block to switch instead of `case true`
the `default` block  should always be in switch

src/calculateRentalCost.js
```diff
@@ -29,6 +29,25 @@
...
+  switch (true){
+
+    case days > 6 :
+      return total - weekDiscount;
+
+    case days > 2 :
+      return total - threeDaysDiscount;
+
+    case days > 0 :
+      return total;
```
Denchepornyuk, line 46: I believe you can return `total` in default block, `return 0;` is redundant here

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1042, opened 2021-12-13
Solution

AntonFof:


AntonFof:
Solution of the task "Transportation on vacation"

PR #1048, opened 2021-12-14
save

bcmp4p1:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+
+  if (days < 3) {
+    total = days * 40;
```
Rozdorozhnii, line 34: You use the expression: (days * 40) in other places, so let's create a separate variable for this.

Rozdorozhnii (changes requested):

Rozdorozhnii (approved):

denys-cheporniuk (approved):

PR #1049, opened 2021-12-14
on-vacation

IvanShulhan:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
```
aliceblunt8, line 31: let's add variables for `days * 40` and for different discounts

aliceblunt8 (changes requested):

danylo-onishchuk (approved):

PR #1052, opened 2021-12-23
Done

Alina-Yermolenko:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,16 @@
...
+
 function calculateRentalCost(days) {
-  // write code here
+  if (days <= 2) {
+    return everyDayCost * days;
+  } else if (days >= 3 && days <= 6) {
+    return everyDayCost * days - 20;
+  } else if (days >= 7) {
+    return everyDayCost * days - 50;
+  }
```
vkryvytskyy, line 39: Move all digits to variables

src/calculateRentalCost.js
```diff
@@ -27,8 +27,16 @@
...
 
+const everyDayCost = 40;
+
 function calculateRentalCost(days) {
-  // write code here
+  if (days <= 2) {
+    return everyDayCost * days;
+  } else if (days >= 3 && days <= 6) {
+    return everyDayCost * days - 20;
+  } else if (days >= 7) {
```
```suggestion
  if (days <= 2) {
    return everyDayCost * days;
  }

  if (days <= 6) {
    return everyDayCost * days - 20;
  }
```
You can omit else, since each if has return statement inside

vkryvytskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -27,8 +27,23 @@
...
+    return everyDayCost * days;
+  }
+
+  if (days <= midTerm) {
+    return everyDayCost * days - 20;
+  }
+
+  if (days >= longTerm) {
+    return everyDayCost * days - 50;
+  }
```
vkryvytskyy, line 46: 20 & 50 also better to move to the variables

vkryvytskyi (approved):

PR #1093, opened 2022-01-07
Solution

McBorov:


McBorov:
In the solution, the function was copied and pasted because I was re-creating the repository

danylo-onishchuk (approved):

PR #1106, opened 2022-01-18
Solution

Oleksii-Pinchuk:


dmytroosypiuk (commented):
Ніби все класно.

dmytroosypiuk (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
+    return fixedRentPrice;
+  }
+
+  if (days >= 3 && days < 7) {
+    return fixedRentPrice - shortTermDiscount;
+  }
+
+  if (days >= 7) {
+    return fixedRentPrice - longTermDiscount;
+  }
```
Anna-Tkachenko, line 48: One of this if statement is redundant here, you can return one of values by default

Anna-Tkachenko (approved):

PR #1111, opened 2022-01-19
rent

pogrebnyakp:


dmytroosypiuk (changes requested):
Все ок, але в чеклісті сказано:
[CODE STYLE] - You should return the value immediately in if statements.
[VARIABLES] - use variables for the main values so that you can reuse them and give them descriptive names.
Попробуй використати метод зі switch.

dmytroosypiuk (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return daysPrice;
+  }
+
+  if (days > 2 && days < 7) {
+    return daysPrice - smallDiscount;
+  }
+
+  if (days > 6) {
+    return daysPrice - bigDiscount;
+  }
```
Anna-Tkachenko, line 45: One of this if statements is redundant here, you can return one of values by default

Anna-Tkachenko (approved):

PR #1120, opened 2022-01-19
done!

krasnobokydima:


oleksandr-kozhushko:
Super Solution!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,12 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDiscount = 20;
+  const superDiscount = 50;
+  const price = 40;
+
+  return days < 3 ? days * price : days > 6
+    ? days * price - superDiscount : days * price - basicDiscount;
```
Anna-Tkachenko, line 36: Don't repeat yourself: create variable for days * price value
It's better to use if else statement instead on nesting with ternary operators

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1131, opened 2022-01-20
Solution

ArtemPaskall:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const discount = 20;
+  const discountExtra = 50;
+
+  if (days < 3) {
+    totalAmount = price * days;
+  } else if (days < 7) {
+    totalAmount = (price * days) - discount;
+  } else if (days >= 7) {
+    totalAmount = (price * days) - discountExtra;
+  }
```
Anna-Tkachenko, line 42: Check point 3 in checklist

Anna-Tkachenko: Don't repeat yourself: create variable for price * days value

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const price = 40 * days;
+  const discount = 20;
+  const discountExtra = 50;
+
+  if (days < 3) {
+    return price;
+  } else if (days < 7) {
+    return price - discount;
+  }
```
Anna-Tkachenko, line 39: If you use return in each statement, else is redundant here
```suggestion
  if (days < 3) {
    return price;
  }
  
  if (days < 7) {
    return price - discount;
  }
```

Anna-Tkachenko (approved):

PR #1138, opened 2022-01-25
Solution

German-Mariia:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 1 && days < 3) {
```
aliceblunt8, line 31: let's introduce more variables 
create a variable for days * 40 (you use it 3 times)
and variables for 20 and 50

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #1148, opened 2022-02-05
completed

mishamolnar:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,9 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const result = days < 3 ? days * 40 : days * 40 - 20;
+
+  return result > 220 ? result - 30 : result;
```
aliceblunt8, line 34: let's create variables in this task with total amount and 2 different types of discount
and I think it's better to use if blocks here instead of ternary operator (more readable)

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #1152, opened 2022-02-07
solution

SanyaBratashchuk:


krolich500grn (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 7) {
+    return days * price - discountFor7Days;
+  }
+
+  if (days >= 3) {
+    return days * price - discountFor3Days;
+  }
+
+  return days * price;
```
jstmpelowycz, line 43: I'd recommend moving `days * price` to a var so that not to repeat yourself 3 times

jstmpelowycz (changes requested):
that's ok, needs a single change

GoDfreE25 (approved):

Ihor-Karpyn (approved):

PR #1157, opened 2022-02-08
Solution

Andr1yK:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rent = days * 40;
+
+  if (days >= 7) {
+    return rent - 50;
+  }
+
+  if (days >= 3) {
+    return rent - 20;
```
jstmpelowycz: I would recommend moving discount values 20/50 to vars

jstmpelowycz (changes requested):
Everything is fine, only two changes to apply

jstmpelowycz (approved):
interesting solution!

Ihor-Karpyn (approved):

PR #1166, opened 2022-02-08
Transportation-on-vacation HW

Ostafii4uk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days <= 2) {
+    return days * 40;
+  }
+
+  if (days > 2 && days < 7) {
+    return (days * 40) - 20;
```
Marta1993-droid, line 36: use DRY approach. to avoid repeating code create variable x = days*40 and use it further.

Marta1993-droid (changes requested):

Marta1993-droid (approved):
Nive! You can also put discount amount in variable.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rentalDaysPrice = days * 40;
+
+  if (days <= 2) {
+    return rentalDaysPrice;
+  }
+
+  if (days > 2 && days < 7) {
+    return rentalDaysPrice - 20;
```
Ihor-Karpyn, line 38: What mean 20? Look like magic number. Create var

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  }
+
+  if (days >= 7) {
+    return rentalDaysPrice - 50;
```
Ihor-Karpyn, line 42: same

Ihor-Karpyn (changes requested):

src/calculateRentalCost.js
Ostafii4uk: Ok, got it

src/calculateRentalCost.js
Ostafii4uk: ok

Ihor-Karpyn (approved):

PR #1174, opened 2022-02-09
solution

RomanKushyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const sum = days * 40;
+
+  if (days >= 7) {
+    return sum - 50;
+  }
+
+  if (days >= 3) {
+    return sum - 20;
```
jstmpelowycz, line 38: I'd recommend moving 20/50 discount values to vars

jstmpelowycz (changes requested):
everything is fine, just a single thing to change

Ihor-Karpyn (approved):

PR #1175, opened 2022-02-10
Solution

katiailchenko:


package-lock.json
```diff
@@ -1,8 +1,7255 @@
 {
   "name": "js-task-boilerplate",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
ivadimko: not important, but nice to know:

here you changed lockFile version to 2 by simply running `npm install` because you use node16 and npm8

to avoid such an issue you have to use older version of nodejs & npm. Basically v14 for nodejs and v6 or v7 for npm. You can use `nvm` to manage node versions on the fly

https://github.com/nvm-sh/nvm

PR #1182, opened 2022-02-15
Solution

bohdan-daniliuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
+    return days * 40 - 20;
```
Ihor-Karpyn, line 34: `days * 40` don't repeat yourself

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+  } else {
+    return days * 40;
```
Ihor-Karpyn, line 36: `days * 40` don't repeat yourself

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
```
Ihor-Karpyn, line 33: `else if` redundant when in prev block `return`
```suggestion
  }
  
  if (days >= 3) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+    return days * 40 - 20;
+  } else {
+    return days * 40;
+  }
```
Ihor-Karpyn, line 37:
```suggestion
  }
  
  return days * 40;
  
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 }
```
Ihor-Karpyn, line 38: What is 40, 50, 20, magic number, create variable

Ihor-Karpyn (changes requested):

PR #1202, opened 2022-03-19
Solution

EnjSadman:


ZUBOV-ILLIA (changes requested):
Good job!
Some notice:
Create some variables, because we not undarstand what is 40, 50, 20 ...

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+  const discountThree = 20;
+
+  if ((days === '') || (days === 0)) {
+    return 0;
+  }
+
+  if (days >= 7) {
+    return (days * discountBasic - discountWeek);
+  } else if (days >= 3) {
+    return (days * discountBasic - discountThree);
```
Denchepornyuk, line 43: create a variable for `days * discountBasic` because you use it three times

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1209, opened 2022-03-28
Solution

yaremenko-maksym:


Vakolyyk (approved):
Good job

Anna-Tkachenko (changes requested):
Create variables for 20, 40 and 50 values

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const sevenDayDiscount = 50;
+  const threeDayDiscount = 20;
```
Anna-Tkachenko, line 33: Check point 5 in checklist

Anna-Tkachenko (approved):

PR #1210, opened 2022-03-28
done

Nikitiliok:


SeKaM0 (approved):
Looks good. But you must replace some magic numbers like 20 or 50. Will better if you create const for this value

Anna-Tkachenko (changes requested):
Fix @SeKaM0 comment

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const clearPrice = days * 40;
+  const firstDiscount = 20;
+  const secondDiscount = 50;
+
+  if (days < 3) {
+    return clearPrice;
+  } else if (days < 7) {
+    return clearPrice - firstDiscount;
+  }
```
Anna-Tkachenko, line 39: else is redundant here
```suggestion
  }
  
  if (days < 7) {
    return clearPrice - firstDiscount;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const clearPrice = days * 40;
```
Anna-Tkachenko, line 31: You should also create variable for 40 value

Anna-Tkachenko (approved):

PR #1215, opened 2022-03-28
solution

kkiirill:


SeKaM0 (changes requested):
Looks good. But you must replace some magic numbers like 20 or 50. Will better if you create const for this value

Anna-Tkachenko (approved):
Good job 👍

PR #1216, opened 2022-03-29
solution

Jenia80:


YegorKochetkov (changes requested):
See [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

1.[VARIABLES] - use variables for the main values so that you can reuse them and give them descriptive names.
5.[VARIABLES] - Don't hardcode some values that may change in the future in variable names (numbers etc.)
BAD EXAMPLE:

GOOD EXAMPLE:

const basicDiscount = 50;

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+  const firstDiscountDays = 3;
+  const maxDiscountDays = 7;
+  const price = 40;
+  const firstDiscount = 20;
+  const maxDiscount = 50;
+
+  let cost = 0;
+
+  if (days < firstDiscountDays) {
+    cost = price * days;
```
YegorKochetkov, line 40: don't repeat code, move it to const = price * days

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+  }
+
+  if (days >= firstDiscountDays && days < maxDiscountDays) {
+    cost = price * days - firstDiscount;
+  }
+
+  if (days >= maxDiscountDays) {
```
YegorKochetkov, line 47: && days < maxDiscountDays) - you can remove this. it's redundant

YegorKochetkov (changes requested):

Anna-Tkachenko (approved):
Good job 👍

PR #1220, opened 2022-03-30
done

Yurii837:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let res = days * 40;
+
+  if (days > 2 && days < 7) {
+    res = res - 20;
+  } else if (days > 6) {
+    res = res - 50;
+  }
```
YegorKochetkov, line 37: Good, but I'd not hardcode cost and discounts and move it to variables.

YegorKochetkov (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let res = days * 40;
```
Anna-Tkachenko, line 31: Use full naming for variables

src/calculateRentalCost.js
Anna-Tkachenko, line 37: Check all points from checklist

Anna-Tkachenko (changes requested):

Yurii837:
I understood folowing style-guides. Seems all is ok)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
+  if (days > 2 && days < 7) {
+    result = result - 20;
+  } else if (days > 6) {
+    result = result - 50;
+  }
```
Anna-Tkachenko, line 37: Check checklist)
20, 40, 50 - magic numbers
You should create variables for this values

Anna-Tkachenko (changes requested):

Yurii837:
Fixed!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const price = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  let result = days * price;
+
+  if (days > 2 && days < 7) {
+    result = result - smallDiscount;
+  } else if (days > 6) {
+    result = result - bigDiscount;
+  }
```
Anna-Tkachenko, line 40: You can return needed values immediately in if statements

Anna-Tkachenko (approved):
Good job

PR #1221, opened 2022-03-30
solution

KostiaO:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
```
VETALEK, line 31:
```suggestion
```

VETALEK: just an unnecessary comment

VETALEK (approved):
All good

Anna-Tkachenko (changes requested):
20, 50, 40 - magic numbers, you should create variables for all main values

KostiaO:
added main variables

Anna-Tkachenko (approved):
Good job 👍

PR #1222, opened 2022-03-30
solution

OleksandrNi:


KostiaO:
Хорошее решение, но можно было упростить...

Anna-Tkachenko (changes requested):
Check all points from checklist

OleksandrNi:
add variables

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
   // write code here
+  let sum = days * 40;
+
+  if (days > 6) {
+    sum = sum - 50;
+  } else {
+    if (days > 2) {
+      sum = sum - 20;
+    }
+  }
```
Anna-Tkachenko, line 40: You should create variables for 20 and 50 numbers, now I can't understand their meaning
And check point 3 in checklist (good example of returning value from if statement)

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days > 6) {
+    return sum - bigDiscount;
+  } else {
+    if (days > 2) {
+      return sum - smallDiscount;
+    }
+  }
```
Anna-Tkachenko, line 42: else is redundant here (point 3 in checklist)
```suggestion
  if (days > 6) {
    return sum - bigDiscount;
  }
  
  if (days > 2) {
    return sum - smallDiscount;
  }
  
```

Anna-Tkachenko (approved):

PR #1226, opened 2022-03-30
Solution

strukk311:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days < 3) {
+    return 40 * days;
```
ivan-kadykalo, line 33: Don't repeat your calc `40*days` in every `if`. DRY

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  }
+
+  if (days >= 7) {
+    return 40 * days - 50;
```
ivan-kadykalo, line 37: Don't use magic numbers such a `40` and `50`

ivan-kadykalo (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+    return result;
+  }
+
+  if (days >= 7) {
+    return result - biggerDiscount;
+  }
+
+  if (days >= 3 && days <= 6) {
+    return result - smallerDiscount;
+  }
```
Anna-Tkachenko, line 47: Last if is redundant, you can return one value by default from your function

Anna-Tkachenko (approved):

PR #1227, opened 2022-03-30
solution

ShevchyshynRoman:


andriyshvets:
Good job

andriyshvets (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const moreThreeDaysDiscount = 20;
+  const moreSevenDaysDiscount = 50;
```
Anna-Tkachenko, line 33: Check point 5 in checklist

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return sum;
+  }
+
+  if (days >= 3 && days < 7) {
+    return sum - moreThreeDaysDiscount;
+  }
+
+  if (days >= 7) {
+    return sum - moreSevenDaysDiscount;
+  }
```
Anna-Tkachenko, line 46: Last if is redundant here, you can return one of values by default from function

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const firstDiscount = 20;
+  const secondDiscount = 50;
```
Anna-Tkachenko, line 33: Think about more meaningful variable names, what is first discount, second discount?

Anna-Tkachenko (approved):

PR #1231, opened 2022-03-30
solution

EvgenBabin1:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
VETALEK, line 31: better use more descripful name like `cost`

VETALEK (approved):

Anna-Tkachenko (approved):

PR #1233, opened 2022-03-30
solution commit

IlliaBezditnyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+  const rentPrice = 40;
+  const basicDiscount = 20;
+  const extendedDiscount = 50;
+
+  for (let day = 0; day < days; day++) {
+    total += rentPrice;
+  }
```
YegorKochetkov, line 38: Why iterate over when simply multiplying is enough?
```suggestion
  total = rentPrice * days;
```

YegorKochetkov (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  total = rentPrice * days;
+
+  if (days >= 3 && days < 7) {
+    total -= basicDiscount;
+  }
+
+  if (days >= 7) {
+    total -= extendedDiscount;
+  }
```
Anna-Tkachenko, line 44: Will be better to return some value immediately in if statement

Anna-Tkachenko (approved):

PR #1234, opened 2022-03-30

RuslanaTomnyuk:


viacheslav-petrovskyi (changes requested):
you need to add variables for discount and price

viacheslav-petrovskyi (approved):
i think great)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days < 3) {
+    return days * price;
```
Anna-Tkachenko, line 36: Don't repeat yourself: create separate variable for days * price value

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job 👍

PR #1236, opened 2022-03-30
done

kurochka-vladyslav:


VETALEK (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const sum = 40 * days;
+
+  if (days >= 7) {
+    return sum - 50;
+  } else if (days >= 3) {
+    return sum - 20;
+  }
```
Anna-Tkachenko, line 37: Check point 3 in checklist

Anna-Tkachenko (changes requested):
20, 50 - magic numbers
Create variables for this values

Anna-Tkachenko (approved):
Good job 👍

PR #1237, opened 2022-03-30
Solution

lem47:


YegorKochetkov (changes requested):
See [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

1.[VARIABLES] - use variables for the main values so that you can reuse them and give them descriptive names.
5.[VARIABLES] - Don't hardcode some values that may change in the future in variable names (numbers etc.)
BAD EXAMPLE:

GOOD EXAMPLE:

const basicDiscount = 50;

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicPrice = 40;
+  const lowDiscount = 20;
+  const highDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return days * basicPrice - lowDiscount;
```
Anna-Tkachenko, line 36: Don't repeat yourself: create separate variable for days * basicPrice value

Anna-Tkachenko (approved):

PR #1239, opened 2022-03-30
Finish the task - Transportation on vacation

khozhainov-aleksandr:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalDays = days * 40;
```
VETALEK, line 31:
```suggestion
  const totalCost = days * 40;
```
this looks like you are counting days, not a cost

VETALEK (approved):

src/calculateRentalCost.js
Anna-Tkachenko: agree, change variable name

Anna-Tkachenko (approved):

PR #1240, opened 2022-03-30
Solution

Anastasiia-Sharkova:


plankton9991 (changes requested):

plankton9991 (approved):

Anna-Tkachenko (approved):
40 - magic number, create variable for this value

PR #1243, opened 2022-03-30
solution

Kateryna03:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costs = 40;
+  let total = 0;
+
+  if (days < 3) {
+    total = days * costs;
```
Anna-Tkachenko, line 35: Don't repeat yourself: create separate variable for days * cost value

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  } else if (days > 2 && days < 7) {
+    total = days * costs - 20;
+  } else if (days > 6) {
+    total = days * costs - 50;
```
Anna-Tkachenko, line 39: 20, 50 - magic numbers
Create variables for them
Also, check point 3 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 👍

PR #1250, opened 2022-04-04
Solution

AndriiRut:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,12 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
```
Bohdan-Maksymiuk, line 32: `40 * days` transfer to separate variable. No need to calculate every time.

Bohdan-Maksymiuk (changes requested):
It could be only two if's - `>= 3` and `>=7`

src/calculateRentalCost.js
AndriiRut: ok! thx for advice!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
...
 function calculateRentalCost(days) {
-  // write code here
-}
+  const total = 40 * days;
+
+  if (days >= 7) {
+    return total - 50;
+  } else if (days >= 3 && days < 7) {
+    return total - 20;
+  }
```
Denchepornyuk, line 37: create variables for all numbers

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1253, opened 2022-04-06
solution

koshachina:
calculate Rental Cost task

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  switch (true) {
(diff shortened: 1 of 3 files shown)


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  switch (true) {
```
ArtemTopolskyi, line 33: maybe it's better to use 3 'if' constructions instead of switch(true)

ArtemTopolskyi (changes requested):

ArtemTopolskyi (approved):

PR #1257, opened 2022-04-07
Solution QA feb 22 Roman Stas

LastTen:


src/calculateRentalCost.js
```diff
@@ -1,34 +1,43 @@
...
+ *  - calculateRentalCost(3) === 100
+ *  - calculateRentalCost(7) === 230
+ *
+ * @param {number} days
+ *
+ * @return {number}
+ */
+
+function calculateRentalCost(days) {
+  // write code here
```
Oleksandra-Pasenchenko, line 31: remove this line

src/calculateRentalCost.js
```diff
@@ -1,34 +1,43 @@
...
+  const rentDay = 40;
```
Oleksandra-Pasenchenko, line 32: it's not a day, it's a price, so choose a more appropriate name for this variable

Oleksandra-Pasenchenko (changes requested):
It seems there are extra files in your commit, you need to change only the file with a solution and, if needed, package.json and package-lock.json

Oleksandra-Pasenchenko (approved):

PR #1260, opened 2022-04-07
Alina Tkachuk Solution

alinatkachukkkk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rental = 40;
+  let total = days * rental;
+
+  if (days <= 2) {
+    total = (days * rental);
```
Oleksandra-Pasenchenko, line 35: you can also return the final cost in the if-statement, there are no further calculations in this function

Oleksandra-Pasenchenko (approved):

PR #1261, opened 2022-04-07
Solution

xxxAvadDonxxx:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
+
+  if (days < 3) {
+    sum = days * 40;
```
Oleksandra-Pasenchenko, line 34: you can also return the final cost in the if-statement, there are no further calculations in this function

Oleksandra-Pasenchenko (approved):

PR #1264, opened 2022-04-08
Task

Danylo-Kotliarenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const price = 40;
+  const rent = days * price;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const threeDays = 3;
```
Oleksandra-Pasenchenko, line 35: if you have variables for terms, better change their names to something like `shortTerm` and `longTerm` because it is changed to 4 days but the variable name is `threeDays`, it will be confusing

Oleksandra-Pasenchenko (approved):

PR #1267, opened 2022-04-08
Solution

khrystynaTserkovna:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = 0;
+
+  switch (true) {
```
tnboyko, line 33: use `if` instead of `switch (true)`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    case (days >= 3 && days < 7):
```
tnboyko, line 34: numbers(3, 7, 20, 40, 50) are better set in variables

tnboyko (approved):

PR #1268, opened 2022-04-08
Solution

DaniilDavydov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = 40 * days;
```
tnboyko, line 31: numbers(3, 7, 20, 40, 50) are better set as variables

tnboyko (approved):

PR #1269, opened 2022-04-09
Solution

AndriiHutnyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
```
marina-tilniak, line 31: create variables for all this numbers

marina-tilniak: it's best practice in general

marina-tilniak (changes requested):

marina-tilniak (changes requested):
please also create variables for days and discounts

ArturErmolenko (approved):
Well done

PR #1271, opened 2022-04-10
Develop

YanaOleksiienko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = days * 40;
+
+  if (days >= 3 && days <= 6) {
+    totalAmount = totalAmount - 20;
```
tnboyko, line 34:
```suggestion
    totalAmount -= 20;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days > 6) {
```
tnboyko, line 35: if there is `else if`, then there should be `else` too. or you can just write several `if`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+    totalAmount = totalAmount - 50;
```
tnboyko, line 36:
```suggestion
    totalAmount -= 50;
```

tnboyko (approved):

PR #1280, opened 2022-04-16
Solution

vitalii-heiko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    total = days * 40 - 50;
+  } else if (days >= 3) {
+    total = days * 40 - 20;
+  } else {
+    total = days * 40;
+  }
+
+  return total;
```
aliceblunt8, line 41: use total instead of days * 40
also you can return the final result instead of rewriting it inside each if

aliceblunt8 (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
+
+  if (days >= 7) {
+    total = total - 50;
+  } else if (days >= 3) {
+    total = total - 20;
+  }
```
ArtemTopolskyi, line 37: use variables instead of all magic numbers

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
+
+  if (days >= 7) {
+    total = total - 50;
```
ArtemTopolskyi, line 34: you can return result in 'if' so you don't need to use 'else' conditions

ArtemTopolskyi (changes requested):

aliceblunt8 (approved):

PR #1281, opened 2022-04-19
Solution

59MrRobot:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
   // write code here
+  let total;
+
+  if (days < 3) {
+    total = 40 * days;
+  } else if (days >= 3 && days < 7) {
+    total = (40 * days) - 20;
+  } else {
+    total = (40 * days) - 50;
+  }
```
andreisucman, line 40: You multiply days with the fee 3 times on each check. Your code will require less processing if you multiply days with the fee outside just once outside the conditions, and then deduct the discount inside the conditions.

ansu16281 (approved):
Good job Swazi. I have a suggestion on how to simply it if you want.

src/calculateRentalCost.js
59MrRobot: Thank you, I will implement this

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let total = 40 * days;
+
+  if (days >= 3 && days < 7) {
+    total -= 20;
+  } else if (days >= 7) {
+    total -= 50;
+  }
```
ihor-jpeg, line 38: in this case its okay, but multiple `else if`s are definitely not easy to read. 
what is 40? what is 20? what is 50? imagine somebody asks you to refactor the code. you don't have any idea what these numbers mean.

readability and understandable code are of the most importance.
then you can think about how you can shorten it, but only if it does not hinder the above

ihor-jpeg (changes requested):

ihor-jpeg (approved):
Well done!

PR #1282, opened 2022-04-19
Solution

Bohdan-Dubravin:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  }
+
+  if (days < 7) {
+    return 40 * days - 20;
+  }
+
+  return 40 * days - 50;
```
ArtemTopolskyi, line 39: create variables for all magic numbers (3, 7, 40, 50, 20)

ArtemTopolskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const standartCost = 40;
+  const rentDays3 = 3;
```
ArtemTopolskyi, line 32:
```suggestion
  const smallRentTime= 3;
```
or something like that
if this number will change in future - you should change both variable name in all places and value of this variable

ArtemTopolskyi (changes requested):

marina-tilniak (approved):

PR #1283, opened 2022-04-19
Task completed

ansu16281:


59MrRobot (changes requested):
Good job, just one suggestion.

59MrRobot (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = days * 40;
+
+  if (days >= 3 && days < 7) {
+    cost -= 20;
+  } else if (days >= 7) {
+    cost -= 50;
+  }
```
ihor-jpeg, line 37: i'd suggest using if statements if possible, without else if becaue it adds comlexity. in this case it can be done with just `if`

ihor-jpeg (approved):
nice solution!

src/calculateRentalCost.js
ihor-jpeg: also what are these magic numbers like 20 and 50. at the moment i have no idea what they mean

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #1284, opened 2022-04-19
solution

efdalyalcin:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  let total = 0;
```
amitpatiljc, line 32: I think less variables is good thing in terms of space complexity, may be we can avoid variable total by directly return statements like
return days * costPerDay;

amitpatiljc (changes requested):
Good Job ! but i think we can optimise the solution.

src/calculateRentalCost.js
efdalyalcin: done, good suggestion..

amitpatiljc (approved):
Good Job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const costPerDay = 40;
+
+  if (days < 3) {
+    return days * costPerDay;
+  } else if (days >= 3 && days < 7) {
+    return days * costPerDay - 20;
+  } else {
+    return days * costPerDay - 50;
+  }
```
ihor-jpeg, line 39: if you immediately return do not add unnecessary `else if`. just two `if`s would be cleaner

ihor-jpeg (approved):

src/calculateRentalCost.js
ihor-jpeg: also, when looking at this code i have no idea what 20 and 50 are. it would be good to create variables to signify their meaning

ihor-jpeg (changes requested):
.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return days * costPerDay;
+  }
+
+  if (days < 7) {
+    return days * costPerDay - threeDayDiscount;
+  } else {
+    return days * costPerDay - sevenDayDiscount;
+  }
```
VinnieJ-2k20, line 43: `return` ends function execution, so else here is redundant.
```suggestion
  if (days < 7) {
    return days * costPerDay - threeDayDiscount;
  } 
  
  return days * costPerDay - sevenDayDiscount;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
+
+  if (days < 3) {
+    return days * costPerDay;
```
VinnieJ-2k20, line 36: You calculate `days * costPerDay` three times.
DRY. Move it to variable and do just once.

VinnieJ-2k20 (changes requested):
Couple of improvements left.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return cost;
+  }
+
+  if (days < 7) {
+    return cost - threeDayDiscount;
+  } else {
+    return cost - sevenDayDiscount;
+  }
```
meforoma, line 43: redundant else, you can do the final return without it
```suggestion
  }

  return cost - sevenDayDiscount;
```

meforoma (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #1285, opened 2022-04-20
solution

LukaszGryziewicz:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
```
59MrRobot, line 31: Because you're using this value only, I do not think it is necessary to put in its own variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const totalCost = days * costPerDay;
+
+  if (days >= 7) {
+    return totalCost - 50;
+  }
+
+  if (days >= 3) {
+    return totalCost - 20;
+  }
```
59MrRobot, line 40: I think it'd be cleaner if you put this in an `if` and `else if` statements

59MrRobot (changes requested):
Good job Lukasz, just a couple suggestions.

src/calculateRentalCost.js
LukaszGryziewicz: Hey Swazi, I don't necessarily agree that else if statements === clean code. I was taught before to avoid them if it is possible. They for sure make the code more compact, but at the same time not very readable, especially if there are multiple.

59MrRobot (approved):
Good job!

src/calculateRentalCost.js
ihor-jpeg: agree with Lukasz

ihor-jpeg (approved):
Well done! But at the moment I have no idea what 20 and 50 are. They seem like random numbers and if a new programmer comes, they will have no idea what they mean. I suggest putting them in variables

src/calculateRentalCost.js
ihor-jpeg: but let's give 20 and 50 some context. i have no idea what they are when i look at this code

ihor-jpeg (changes requested):

ihor-jpeg (approved):
Well done!

PR #1286, opened 2022-04-20
Solution

amitpatiljc:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days > 6) {
+    return (days * 40) - 50;
+  }
+
+  if (days > 2) {
+    return (days * 40) - 20;
+  }
+
+  return days * 40;
```
59MrRobot, line 39: This works perfectly but I think `if` and `else if` statements would make the code look more cleaner.

59MrRobot (changes requested):
Good job, I've just got one suggestion

59MrRobot (approved):
Great work Amit!

src/calculateRentalCost.js
ihor-jpeg: i think else if adds complexity as you need a separate varibale. where in this way you can simply return the result

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 6) {
+    return (days * 40) - 50;
+  } else if (days > 2) {
+    return (days * 40) - 20;
+  } else {
+    return days * 40;
+  }
```
ihor-jpeg, line 37: this does not make much sense, if you return the value, then you do not really need `else if` or `else`

ihor-jpeg (approved):

src/calculateRentalCost.js
ihor-jpeg: also let's not have magic numbers in the code. at the moment i dont understand what 20, 40 and 50 are

ihor-jpeg (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days > 6) {
+    return (days * costPerDay) - disountForSevenDays;
+  }
+
+  if (days > 2) {
+    return (days * costPerDay) - disountForThreeDays;
+  }
+
+  return days * costPerDay;
```
ihor-jpeg, line 43: You have repeated `days * costPerDay` a few times. maybe put it in a variable?

ihor-jpeg (approved):
Well done!

PR #1287, opened 2022-04-20
Task complete

MabuzaM:
Task complete

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = 0;
+
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days < 7) {
+    cost = (days * 40) - 20;
+
+    return cost;
+  }
+
+  if (days >= 7) {
+    cost = (days * 40) - 50;
+
+    return cost;
+  }
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,22 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  let cost = 0;
-
   if (days < 3) {
     return days * 40;
   }
 
   if (days < 7) {
-    cost = (days * 40) - 20;
-
-    return cost;
+    return (days * 40) - 20;
   }
 
   if (days >= 7) {
-    cost = (days * 40) - 50;
-
-    return cost;
+    return (days * 40) - 50;
   }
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return days * 40;
+  }
+
+  if (days < 7) {
+    cost = (days * 40) - 20;
+
+    return cost;
+  }
+
+  if (days >= 7) {
```
efdalyalcin, line 43: Just a suggestion: I think you can use `else if` instead of putting spaces between if statements. It is both less lines and easier to read in my opinion.

efdalyalcin (approved):
It works perfectly,  good job!! Just a small suggestion in the comments.

src/calculateRentalCost.js
ihor-jpeg: multiple else if are definitely not easy to read

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days < 7) {
+    cost = (days * 40) - 20;
+
+    return cost;
+  }
```
ihor-jpeg, line 41: what is 40? what is 20? what is 50? i have no idea
also do not create a variable just to then return it

ihor-jpeg (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
VinnieJ-2k20, line 32: You calculate `days * 40` three times. Do it once and use a varaiable.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  if (days < 7) {
+    return (days * 40) - 20;
```
VinnieJ-2k20, line 36: numbers 40, 20 and 50 are not clear.
Create variables with definitive names.

VinnieJ-2k20 (changes requested):
Couple of final edits left.

src/calculateRentalCost.js
MabuzaM: Noted. Have changed it kindly check.

src/calculateRentalCost.js
MabuzaM: Thanks have made changes

meforoma (approved):

PR #1288, opened 2022-04-20
Solution

Uktam19980416:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
 
 function calculateRentalCost(days) {
   // write code here
+  // let sum = 0;
```
ihor-jpeg, line 32: let's remove commented code

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+  const rentalCost = 40;
+  const moreThanThreeDays = 20;
+  const moreThanSevenDays = 50;
```
ihor-jpeg, line 35: maybe
```suggestion
  const smallDiscount = 20;
  const bigDiscount = 50;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+
+  if (days < 3) {
+    return days * rentalCost;
+  } else if (days >= 3 && days < 7) {
+    return days * rentalCost - moreThanThreeDays;
+  } else if (days > 6) {
+    return days * rentalCost - moreThanSevenDays;
+  }
```
ihor-jpeg, line 43: if you return inside of the if, there is no need for such complex structure

ihor-jpeg (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    sum = days * rentalCost;
+  } else if (days >= 3 && days < 7) {
+    sum = days * rentalCost - smallDiscount;
+  } else if (days > 6) {
+    sum = days * rentalCost - bigDiscount;
+  }
```
ihor-jpeg, line 43: why not return immediately?

ihor-jpeg (approved):

PR #1289, opened 2022-04-20
Implemented solution

iLebo-stack:


ansu16281 (changes requested):
Hi Lebogang. Your code is working. But based on Ihor's review of my work I recommend you use variables instead of numbers. Because when you get back to your code after some time it will be hard to remember what those numbers represent.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
```
andreisucman, line 32: Also, you are doing the same calculation in every if block `(days * 40) - 50;` This spends processing power and makes code slower. Consider only changing the variable inside the if block, and doing the calculation outside the if block in the end of the code.

ansu16281 (approved):

ihor-jpeg (approved):

PR #1290, opened 2022-04-21
solution

huberciak:


ansu16281 (approved):
Nice work. But based on Ihor's review of my work he may not like numbers 20 and 50. So I recommend adding a discount variable and changing it inside the if blocks. And then calculating the whole expression after it `const total = days * dayCost - discount;`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  const dayCost = 40;
+  const total = days * dayCost;
+
+  if (days < 3) {
+    return total;
+  } else if (days < 7) {
+    return total - 20;
+  } else {
+    return total - 50;
+  }
```
ihor-jpeg, line 40: let's create variables so it is clear what 20 and 50 mean

ihor-jpeg (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  let discount;
+
+  if (days < 3) {
+    return total;
+  } else if (days < 7) {
+    discount = 20;
+
+    return total - discount;
+  } else {
+    discount = 50;
```
meforoma, line 42: basically declaring with `let` and then assign a value is fine, yet in this case you just have two options,
so better declare them right away and use inside return statements - saves few lines of code
and when you have `if () { return` return terminates function execution, so no actual need to nest `else if / else`:
```suggestion
  }

  if (days < 7) {
    return total - discountTwenty;
  }

  return total - discountFifty
```

meforoma (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
+  const total = days * dayCost;
+  const discountTwenty = 20;
+  const discountFifty = 50;
+
+  if (days < 3) {
+    return total;
+  } else if (days < 7) {
```
VinnieJ-2k20, line 38: `return` statement finishes fucntion execution, so else here is redundant.

VinnieJ-2k20 (changes requested):
One final edit left.

meforoma (approved):

PR #1291, opened 2022-04-21
Task complete

Rashaad-Ebrahim:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const rentalPrice = 40;
+  let total = 0;
```
iLebo-stack, line 33: This variable is not necessary, you can return your results from inside the if statement instead.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+
+  if (days < 3) {
+    total = days * rentalPrice;
```
iLebo-stack, line 36: Instead of doing calculation in multiple lines you can assign a variable to `days * rentalPrice`.

iLebo-stack (changes requested):
Good job, see comments below

src/calculateRentalCost.js
```diff
@@ -30,18 +30,21 @@
...
   // write code here
   const rentalPrice = 40;
+  const rentalAmount = days * rentalPrice;
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
   let total = 0;
 
   if (days < 3) {
-    total = days * rentalPrice;
+    total = rentalAmount;
```
iLebo-stack, line 39: Your code works fine but you could still improve it by returning `rentalAmount` here and the same thing in the following if statements. However you will need to mind the sequence of your if statements.

src/calculateRentalCost.js
```diff
@@ -30,18 +30,21 @@
...
-    total = (days * rentalPrice) - 20;
+    total = rentalAmount - threeDayDiscount;
   }
 
   if (days >= 7) {
-    total = days * rentalPrice - 50;
+    total = rentalAmount - sevenDayDiscount;
   }
 
   return total;
```
iLebo-stack, line 50: When you return inside each if statement you won't need this line of code.

iLebo-stack (changes requested):

iLebo-stack (approved):
Good Job

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+    return rentalAmount;
+  }
+
+  if (days < 7) {
+    return rentalAmount - threeDayDiscount;
+  }
+
+  if (days >= 7) {
+    return rentalAmount - sevenDayDiscount;
+  }
```
meforoma, line 47: this if is implicit - if other two fail, this will apply, so no need for an explicit `if`, just `return`:
```suggestion
    return rentalAmount - sevenDayDiscount;
```

meforoma (changes requested):

src/calculateRentalCost.js
rHashTag: Done!

VinnieJ-2k20 (approved):
Good job 👍

meforoma (approved):

PR #1293, opened 2022-04-22
Develop

Sardorbek1122:


VinnieJ-2k20 (approved):
Good job.
Still, code could be simplified to avoid using `else if`

PR #1295, opened 2022-04-24
Solution

PRO-GRAM-MER:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let cost = 0;
+
+  for (let i = 0; i <= days; i++) {
+    if (days < 3) {
+      cost = 40 * days;
```
VinnieJ-2k20, line 36: 1. You calculate `40 * days` three times. Do it once and use variable.
2. You change one variable and return it right away. Get rid of it and put `return` statement right inside `if` structure. That way you can also avoid using `else`.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+    } else if (days >= 3 && days < 7) {
+      cost = 40 * days - 20;
+    } else {
+      cost = 40 * days - 50;
```
VinnieJ-2k20, line 40: 20 and 50 look like 'magic' numbers. Move them to variables and give them proper names.

VinnieJ-2k20 (changes requested):
Your code works, but it can be improved.

VinnieJ-2k20 (approved):

PR #1296, opened 2022-04-24
task done

frontend112:


src/calculateRentalCost.js
```diff
@@ -28,6 +28,18 @@
...
 function calculateRentalCost(days) {
+  let totalCost = days * 40;
+
+  if (days > 6) {
+    totalCost -= 50;
+  }
+
+  if (days > 2 && days < 7) {
+    totalCost -= 20;
+  }
```
meforoma, line 39: math works, but 20 and 50 are currently what's called `magic numbers`
better be explained = introduced as variables with meaningful names.

meforoma (changes requested):

meforoma (approved):

PR #1297, opened 2022-04-24
solution

WojciechCzarnowski:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    return count - 20;
+  }
+
+  if (days >= 7) {
+    return count - 50;
+  } else {
+    return count;
+  }
```
meforoma, line 42: if with return are conclusive, so you can do final return without else:
```suggestion
  }

  return count;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const count = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return count - 20;
```
meforoma, line 35: usually called `magic numbers` - 20 and 50 are kinda - what are those?
generally same with 40 
declare variables with explanatory names and use them in code so that code is a readme of its own

meforoma (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #1301, opened 2022-04-25
Transportation on vacation task

proznerka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
+
+  if (days >= 7) {
+    total -= 50;
+  } else if (days >= 3) {
+    total -= 20;
```
rHashTag, line 36: For better readability, perhaps use another if statement instead of the else if.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
+
+  if (days >= 7) {
+    total -= 50;
```
rHashTag, line 34: Could I suggest making the discounts into a variable, like `const soManyDaysDiscount = 100` for example. That way, if the discount amount changes in the future, it can easily be changed.

Rashaad-Ebrahim (changes requested):
Good job on the code! Very concise can compact, I like that style.

I've given you some suggestions.

Rashaad-Ebrahim (approved):
LGTM!

VinnieJ-2k20 (approved):

PR #1307, opened 2022-04-26
Solution

Danylo-Dziuban:


vlad-demchuk (approved):
good job!
you can make checks in reverse sequence to avoid several checks. just start from the biggest value and move to the smallest)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    return (days * 40) - 50;
+  }
```
Denchepornyuk, line 41: create variables for all numbers
create a variable for days * 40, you repeat this three times

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1309, opened 2022-04-26
Solution

SavMary:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
iwanttosmile4u, line 32: You can add this part of code - days * 40 - into variable. But I think it's not a mistake, just code style)

iwanttosmile4u (commented):
Good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
Denchepornyuk, line 39: create variables for all numbers
create a variable for `days * 40`, you repeat this three times

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1311, opened 2022-04-26
make function

uliton:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
+  const threeDaysOff = 20;
+  const sevenDaysOff = 50;
+
+  if (days < 3) {
+    return days * dayCost;
```
iwanttosmile4u, line 36: You can also write this code - days * dayCost - into variable. It will be better to read. It just good code style, i think)

iwanttosmile4u (commented):
Good job!

src/calculateRentalCost.js
Denchepornyuk: yep, create a variable for `days * dayCost`, you repeat this three times

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1313, opened 2022-04-26
Solution - ready to check

iwanttosmile4u:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    return totalAmount;
+  }
+
+  if (days < 7) {
+    return totalAmount - 20;
+  } else {
+    return totalAmount - 50;
+  }
```
ZUBOV-ILLIA, line 41: Good job!

ZUBOV-ILLIA (changes requested):

denys-cheporniuk (approved):

PR #1316, opened 2022-04-26
solution

ISRLv2021:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+
+  if (days >= 7) {
+    total = (days * 40) - 50;
+  } else if (days >= 3 && days < 7) {
+    total = (days * 40) - 20;
+  } else {
+    total = days * 40;
```
ZUBOV-ILLIA, line 38: Good job!

ZUBOV-ILLIA (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const longPerid = 50;
+  const shortPeriod = 20;
+  const everyDayPay = 40;
+
+  if (days >= 7) {
+    total = (days * everyDayPay) - longPerid;
+  } else if (days >= 3 && days < 7) {
+    total = (days * everyDayPay) - shortPeriod;
+  } else {
+    total = days * everyDayPay;
```
Denchepornyuk, line 41: create a variable for `days * everyDayPay`, you repeat this three times

Denchepornyuk (changes requested):

PR #1319, opened 2022-04-26
First_commit_Calculate_rental_car

igoree88:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  switch (true) {
+    case days < 3 :
+      return days * 40;
+    case days >= 3 && days < 7 :
+      return days * 40 - 20;
+    case days >= 7 :
+      return days * 40 - 50;
+  }
```
ZUBOV-ILLIA, line 38: Good job!

ZUBOV-ILLIA (changes requested):

src/calculateRentalCost.js
igoree88: Thanks for your comment, I agree with it, so I added some variables.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const discountSeventhDay = 50;
+
+  switch (true) {
+    case days < 3 :
+      return days * rateOfRent;
+    case days >= 3 && days < 7 :
+      return days * rateOfRent - discountFourthDay;
+    case days >= 7 :
+      return days * rateOfRent - discountSeventhDay;
+  }
```
Denchepornyuk, line 42: `switch (true)` isn't a good code, use `if` here, you need only two `ifs` for `days >= 7` and `days >= 3`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const discountFourthDay = 20;
+  const discountSeventhDay = 50;
+
+  switch (true) {
+    case days < 3 :
+      return days * rateOfRent;
+    case days >= 3 && days < 7 :
+      return days * rateOfRent - discountFourthDay;
+    case days >= 7 :
+      return days * rateOfRent - discountSeventhDay;
```
Denchepornyuk, line 41: create a variable for `days * rateOfRent`, you repeat this three times

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1322, opened 2022-04-26
firstJS

Oleksandr-Ivanchenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,12 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
iwanttosmile4u, line 32: Write discount numbers (20, 50), total cost per day (40) into variables, because it isn't understandable what the numbers are. Also you can write such part of code as days * 40 into variable, because you use this part few times.

denys-cheporniuk (approved):

PR #1325, opened 2022-04-26
Solution

vitaliy-kovbasinskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
+
+  switch (true) {
+    case days < 3:
+      return cost;
+    case days < 7:
+      return cost - 20;
```
JustDan17, line 37: Move discounts into variables. Change in all occurrences.

JustDan17 (changes requested):

Denchepornyuk (approved):

PR #1326, opened 2022-04-26
Solution ver 1.0

volodya-leskiv:


enrikadej (approved):
nice short code! 👍

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+
+  if (days >= 7) {
+    return total - 50;
+  }
+
+  if (days >= 3) {
+    return total - 20;
+  }
+
+  return total;
```
ISRLv2021, line 42: Good job!
Fix just this one. Make some variables for values, because for now we don't understand what it is - 40, 50, 20

ISRLv2021 (changes requested):

src/calculateRentalCost.js
volodya-leskiv: Thanks for review, Server!

ISRLv2021 (approved):
Keep it up!

denys-cheporniuk (approved):

PR #1329, opened 2022-04-26
task is solved

enrikadej:


volodya-leskiv (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const smallOff = 20;
+  const bigOff = 50;
+
+  if (days < 3) {
+    return price * days;
```
ISRLv2021, line 36: Create a separate variable for "price * days", you use it three times

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  }
+
+  if (days < 7) {
+    return (price * days) - smallOff;
```
ISRLv2021, line 40: Instead of a return statement, use an initial variable, assigning a new value to it.

ISRLv2021 (changes requested):

ISRLv2021 (approved):
Keep it up!

denys-cheporniuk (approved):

PR #1343, opened 2022-05-05
Solution

vsupruniuk:


mgrinko (changes requested):

mgrinko (changes requested):
`day * 40` must always be the same but is calculated 3 times

Olexandr6 (approved):

PR #1344, opened 2022-05-05
train example

NikitaKarbazin:
Hi, i know about ternary operator, thats bad practice to put in `return` statement. But it really short code.

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,8 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days >= 7 ? ((40 * days) - 50) : days >= 3
+    ? ((40 * days) - 20) : (40 * days);
 }
 
 module.exports = calculateRentalCost;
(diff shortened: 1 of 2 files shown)


mgrinko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (40 * days) - 50;
+  } else if (days >= 3) {
+    return (40 * days) - 20;
+  }
```
illiarerun, line 35: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 1st item. You should create variables for main values - 20, 50, 40 they shouldn't be just magic numbers

illiarerun: Also, checklist 3rd item, why do you have else if after a return statement? Code won't execute past this, use plain if construction

illiarerun (changes requested):

illiarerun (approved):
Gj, but I wouldn't use switch statement here

PR #1347, opened 2022-05-09
Develop

vinnibea:


mgrinko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+  const pricePerDay = 40;
+  const threeDayDiscount = 20;
+  const sevenDaysDiscount = 50;
+
+  if (days < 3) {
+    total = pricePerDay * days;
```
jstmpelowycz, line 37: as you can see, `total` always has the initial value of `pricePerDay * days`, so you can put it there right away

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  } else if (days >= 3 && days < 7) {
+    total = pricePerDay * days - threeDayDiscount;
+  } else {
+    total = pricePerDay * days - sevenDaysDiscount;
+  }
```
jstmpelowycz, line 42: you can simplify it
```suggestion
  if (days >= 7) {
    return total - sevenDaysDiscount; // assume total's initial value is (pricePerDay * days)
  }
  
  if (days >= 3) {
    return total - threeDayDiscount;
  } 
  
  ...
```

jstmpelowycz (changes requested):
good start! let's fix the issues

src/calculateRentalCost.js
vinnibea: > you can simplify it

thanks for hints) Now it looks better

illiarerun (approved):

PR #1354, opened 2022-05-16
Solution

Ksenia-Didukh:


aliceblunt8 (changes requested):
put `days * 40` into a variable, you use it 3 times

aliceblunt8 (approved):

PR #1357, opened 2022-05-18
solution

FANTAZER-NURE:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = 0;
+
+  for (let i = 0; i < days; i++) {
+    cost += 40;
+  }
```
Bohdan-Maksymiuk, line 35: use multiply )

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 7) {
+    cost -= 50;
+  }
+
+  if (days >= 3 && days < 7) {
```
Bohdan-Maksymiuk, line 41:
```suggestion
  if (days >= 3) {
```
no need

Bohdan-Maksymiuk (changes requested):
add variables for all numbers (discounts and calculated default cost)

aliceblunt8 (approved):

PR #1360, opened 2022-05-18
Solution

OleksandraKryvosheina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (40*days)-50;
```
djtyrf312, line 32: > return (40*days)-50;
better to make spaces between variables like
return (40 * days) - 50;

djtyrf312 (approved):

PR #1362, opened 2022-05-19
Solution

marina-bidenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = 40 * days;
+
+  if (days >= 3) {
+    totalAmount -= 20;
+  }
+
+  if (days >= 7) {
+    totalAmount -= 30;
```
djtyrf312, line 38:
```suggestion
    totalAmount -= 50;
```

djtyrf312: did you run tests?

djtyrf312 (changes requested):
did you run tests?

src/calculateRentalCost.js
marina-bidenko: I suppose, exactly -30 is correct, I've run the test and everything was passed. -30 is correct because after the first condition we reduce the total amount by 20 (if more than 3 days), and then if it is even more than 7 days, we reduce it by another 30, so in the end, we have -50 for 7 or more days.

djtyrf312 (approved):

PR #1375, opened 2022-05-26
Solution

Maksym48:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const priceRent = days * 40;
+
+  if (days < 3) {
+    return priceRent;
+  } else if (days >= 3 && days < 7) {
+    return priceRent - 20;
+  } else if (days >= 7) {
+    return priceRent - 50;
+  }
```
ArturErmolenko, line 39:
```suggestion
  if (days < 3) {
    return priceRent;
  } 
   if (days < 7 ) {
    return priceRent - 20;
  } 
    return priceRent - 50;
  }
```
Best practice ---> вынеси дискаунт в переменные

ArturErmolenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const shotrDiscount = 20;
+  const longDiscount = 50;
+
+  if (days < 3) {
+    return priceRent;
+  } else if (days >= 3 && days < 7) {
+    return priceRent - shotrDiscount;
+  } else if (days >= 7) {
+    return priceRent - longDiscount;
+  }
```
vkryvytskyy, line 41: 3 and 7 also should be stored in variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceRent = days * 40;
+  const shotrDiscount = 20;
+  const longDiscount = 50;
+
+  if (days < 3) {
+    return priceRent;
+  } else if (days >= 3 && days < 7) {
```
```suggestion
  }
  
  if (days >= 3 && days < 7) {
```
no need in else if you have return inside previous `if`

vkryvytskyi (changes requested):

aliceblunt8 (approved):

PR #1381, opened 2022-05-30
Solution

Cherednychenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const payment = 40 * days;
+  const stock20 = payment - 20;
+  const stock50 = payment - 50;
```
illiarerun, line 33: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 1st item. You should create variables for main values - 40, 20, 50, they shouldn't be just magic numbers out of nowhere

illiarerun (changes requested):
Nice start, now let's apply few fixes:

jstmpelowycz (approved):
LGTM!

PR #1382, opened 2022-05-30
task is done

annaRav:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const rentalCost = days * 40;
+
+  if (days >= 7) {
+    return rentalCost - 50;
+  }
+
+  if (days >= 3) {
+    return rentalCost - 20;
+  }
```
vkryvytskyy, line 39: Move all numeric values to variables

vkryvytskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const rentalMiddleCost = rentalCost - 20;
+  const rentalLongCost = rentalCost - 50;
+
+  if (days >= 7) {
+    return rentalLongCost;
+  }
+
+  if (days >= 3) {
+    return rentalMiddleCost;
+  }
```
vkryvytskyy, line 41: 7 & 3 also should be moved to variables

vkryvytskyi (approved):

PR #1384, opened 2022-05-31
solution

TanyaHryn:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
+
+  if (days < 3) {
+    count = days * 40 - 0;
+  }
```
aliceblunt8, line 35: no need for this if with 0, it would be default case
create total = days * 40 and change it in the next 2 if block, then return total

aliceblunt8 (changes requested):
put `days * 40` into a variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  if (days < 3) {
+    return total;
+  }
+
+  if (days < 7 & days > 2) {
+    return total - 20;
+  }
+
+  if (days >= 7) {
+    return total - 50;
```
ArtemTopolskyi, line 42: what is 3, 7, 2, 50...
create variables for all magic numbers

ArtemTopolskyi (changes requested):

aliceblunt8 (approved):

PR #1386, opened 2022-06-01
Solution

ik-web:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  const fullPrise = days * pricePerDay;
```
illiarerun, line 32:
```suggestion
  const fullPrice = days * pricePerDay;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const cashbackForTreeDays = 20;
+  const cashbackForSevenDays = 50;
```
illiarerun, line 34: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 5th
item, don't hardcode any values in variable names, value might change, use smth like `basicDiscount`, `advancedDiscount`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  let total = 0;
+
+  if (days >= 7) {
+    total = fullPrise - cashbackForSevenDays;
+  } else if (days >= 3) {
+    total = fullPrise - cashbackForTreeDays;
+  } else {
+    total = fullPrise;
+  }
```
illiarerun, line 43: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 2nd and 3rd items, do you really need this `let` variable? You can return value instantly form `if` block

illiarerun (changes requested):

src/calculateRentalCost.js
ik-web: Fixed

src/calculateRentalCost.js
ik-web: Fixed

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const pricePerDay = 40;
+  const fullPrice = days * pricePerDay;
+  const basicDiscount = 20;
+  const advancedDiscount = 50;
+
+  if (days >= 7) {
+    return fullPrice - advancedDiscount;
+  } else if (days >= 3) {
+    return fullPrice - basicDiscount;
+  }
```
jstmpelowycz, line 40:
```suggestion
  if (days >= 7) {
    return fullPrice - advancedDiscount;
  }
  
  if (days >= 3) {
    return fullPrice - basicDiscount;
  }
```

jstmpelowycz: as far as `return` exits the program right away, you could simplify it

jstmpelowycz (changes requested):
almost done, one thing left to fix

src/calculateRentalCost.js
ik-web: Fixed. Thank you for the hint!

PR #1387, opened 2022-06-01
Solution

liliia-livandovskaya:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  }
```
illiarerun, line 37: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md?rgh-link-date=2022-06-06T07%3A07%3A40Z) 1st item. You should create variables for main values - 20, 50, 40 they shouldn't be just magic numbers

illiarerun: Also, checklist 3rd item, you can return value instantly in if statements, and don't use `else if` statements here, they are redundant

illiarerun (changes requested):

jstmpelowycz (approved):
LGTM 🔥

PR #1391, opened 2022-06-02
solved

mflkvsk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const amount = 40 * days;
+
+  if (days >= 7) {
+    return amount - 50;
+  }
+
+  if (days >= 3) {
+    return amount - 20;
+  }
```
illiarerun, line 39: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 1st item. You should create variables for main values - 20, 40, 50, they shouldn't be just magic numbers

illiarerun (changes requested):
GJ, now let's apply 1 fix:

illiarerun (approved):
GJ 🔥

PR #1392, opened 2022-06-02
Solution

Zykhor-Andrii:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
```
Bohdan-Maksymiuk, line 31:
```suggestion
  const defaultCost = days * 40;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return result - 50;
+  }
+
+  if (days >= 3) {
+    result -= 20;
```
Bohdan-Maksymiuk, line 38:
```suggestion
    return defaultCost - 20;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
+  if (days >= 7) {
+    return result - 50;
```
Bohdan-Maksymiuk, line 34:
```suggestion
    return defaultCost - 50;
```

Bohdan-Maksymiuk (changes requested):
add variables for discount ammounts

ArturErmolenko (approved):

PR #1393, opened 2022-06-02
my solution

Pavlo000:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let price = days * 40;
+
+  if (days >= 3 && days < 7) {
+    price -= 20;
+  } else if (days >= 7) {
+    price -= 50;
+  }
```
illiarerun, line 37: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md?rgh-link-date=2022-06-06T07%3A07%3A40Z) 1st item. You should create variables for main values - 20, 50, 40 they shouldn't be just magic numbers

illiarerun: Also, checklist 3rd item, you can just return values from your if statements instantly to avoid executing redundant code. And you don't really need else statements here in this way

illiarerun (changes requested):
Nice start, now let's try to fix it:

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const bigDiscount = 50;
+  const price = days * defaultPrice;
+
+  if (days >= 3 && days < 7) {
+    return price - smallDiscount;
+  } else if (days >= 7) {
+    return price - bigDiscount;
+  } else {
+    return price;
+  }
```
jstmpelowycz, line 42: too complicated, let's simplify it
```suggestion
  if (days >= 7) {
    return price - bigDiscount;
  } 
  
 if (days >= 3) {
    return price - smallDiscount;
  }
  
  ...
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):
LGTM

PR #1398, opened 2022-06-06
Solution

GavrilyukYura:


Samaev (changes requested):
good job
few changes are needed, look my comment please

Samaev (approved):
good job

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return price;
+  }
+
+  if (days >= 3 && days < 7) {
+    return price - minDiscount;
+  }
+
+  if (days >= 7) {
+    return price - maxDiscount;
+  }
```
illiarerun, line 46: functions should have at least 1 consistent return, you also checked all the conditions before, so this if is redundant
```suggestion
  return price - maxDiscount;
```

illiarerun (approved):
GJ, left one comment

PR #1408, opened 2022-06-06
Solusion

DmytroSukhaniuk:
my solusion

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,14 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  const total = days * 40;
+  const dayRate = 40;
+  const firstSaleDays = 3;
+  const secondSaleDays = 7;
+  const firstSale = 20;
+  const secondSale = 50;
+  const total = days * dayRate;
 
-  if (days >= 3 && days < 7) {
-    return total - 20;
+  if (days >= firstSaleDays && days < secondSaleDays) {
+    return total - firstSale;
   }
 
-  if (days >= 7) {
-    return total - 50;
+  if (days >= secondSaleDays) {
+    return total - secondSale;
   }
 
   return total;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentDay = 40;
+  let total = days * rentDay;
+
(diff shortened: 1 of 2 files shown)


k0nstrukt-or (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
-  // write code here
+  const rentDay = 40;
+  let total = days * rentDay;
+
+  if (days >= 3) {
+    total = total - 20;
+  }
+
+  if (days >= 7) {
+    total = total - 30;
```
jstmpelowycz, line 39: you don't have any value of 30, implement it in a different way

jstmpelowycz: the checklist, p.1

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentDay = 40;
+  let total = days * rentDay;
+
+  if (days >= 3) {
+    total = total - 20;
+  }
```
jstmpelowycz, line 36:
```suggestion
  if (days >= ...) {
    return total - ...;
  }
```

jstmpelowycz (changes requested):
good start, now let's fix the issues

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const total = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return total - 20;
+  }
+
+  if (days >= 7) {
+    return total - 50;
+  }
```
illiarerun, line 39: PAy attention to comments, [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md?rgh-link-date=2022-06-06T18%3A34%3A59Z) 1st item. You should create variables for main values - 20, 50, 40, 3, 7 they shouldn't be just magic numbers

illiarerun, line 39: you can swap these conditions, so you wouldn't have to worry about number of days being higher that 3 and lower than 7

illiarerun (changes requested):

src/calculateRentalCost.js
DmytroSukhaniuk: ок

illiarerun (approved):

PR #1411, opened 2022-06-06
Solution

averstiuk:


stanislav-topikha (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    return fullPrice;
+  } else if (days <= 6) {
+    fullPrice = dayPrice * days - minDiscount;
+
+    return fullPrice;
+  } else {
+    fullPrice = dayPrice * days - maxDiscount;
+
+    return fullPrice;
+  }
```
jstmpelowycz, line 48: the checklist, p.3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  let fullPrice = 0;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days <= 2) {
+    fullPrice = dayPrice * days;
```
jstmpelowycz, line 37: as you can see over your code, `total` always has its default value of `dayPrice * days`, so you put it there right away

jstmpelowycz (changes requested):
good start! let's fix the issues

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+    fullPrice = totalPrice;
+
+    return fullPrice;
+  }
+
+  if (days < 7) {
+    fullPrice = totalPrice - minDiscount;
+
+    return fullPrice;
+  }
```
jstmpelowycz, line 47:
```suggestion

  if (days < 3) {
    return totalPrice;
  }

  if (days < 7) {
    return totalPrice - minDiscount;
  }
```

jstmpelowycz: so, you could get it simpler

jstmpelowycz (approved):

PR #1414, opened 2022-06-06
Solution

illavodonis:


k0nstrukt-or (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
+  }
+
+  if (days >= 5) {
+    return (days * 40) - 50;
+  }
+
+  return days * 40;
```
jstmpelowycz, line 39: the logic is complicated, there is a simper way to implement it

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
```
jstmpelowycz, line 32: `days * 40` looks like a new var, as well as 20, consider applying to the rest of your value (the checklist, p.1)

jstmpelowycz (changes requested):
let's fix the issues

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceRent = 40 * days;
+  const discount3Day = 20;
+  const discount7Day = 50;
```
illiarerun, line 33: Checkslist 5th item, use smth like `basicDiscount` and `vipDiscount`

illiarerun (changes requested):
GJ, left one commen, fix it and we're done:

illiarerun (approved):

PR #1417, opened 2022-06-06
Completed

ant2nta:


dimabelotskyi (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
+
+  for (let q = 1; q <= days; q++) {
```
jstmpelowycz, line 33: suggestion: you don't actually need this for-loop

jstmpelowycz (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const smallDiscount = 20;
+  const bigDiscount = 30;
+  const payPerDays = 40;
+  let sum = payPerDays * days;
+
+  if (days >= 3) {
+    sum -= smallDiscount;
```
jstmpelowycz, line 37: if you hit the spot, you can return a result right here
```suggestion
    return sum - smallDiscount; // like this
```

jstmpelowycz (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const payment = price * days;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days > 6) {
+    return payment - maxDiscount;
+  }
```
jstmpelowycz, line 38:
```suggestion
  if (days >= 7) { // it will be more obvious
    return payment - maxDiscount;
  }
```

jstmpelowycz (changes requested):

illiarerun (approved):
GJ 🔥

PR #1427, opened 2022-06-06
Transportation on vacation task

dima-pyvovarchuk:


vsupruniuk (changes requested):

vsupruniuk (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    return basicPrice;
+  }
+
+  if (days < 7) {
+    return basicPrice - 20;
+  }
+
+  return basicPrice - 50;
```
jstmpelowycz, line 41: let's look at these lines
don't you think you could move those values, 20 and 50, to vars?
by doing so, we simplify the way reviewers understand what's written

jstmpelowycz (changes requested):

jstmpelowycz (approved):
LGTM

PR #1447, opened 2022-06-06
Solution

serhii-vorona:


dima-pyvovarchuk (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return totalPrice;
+  }
+
+  if (days >= 3 && days < 7) {
+    return totalPrice - littleSale;
+  }
+
+  if (days >= 7) {
+    return totalPrice - bigSale;
+  }
```
illiarerun, line 46: You should have at least one consistent return statement (outside of if) and you've already covered all the cases, so this check is redundant
```suggestion
    return totalPrice - bigSale;
```

illiarerun (changes requested):
GJ, but need to fix one thing:

illiarerun (approved):

PR #1449, opened 2022-06-07
Solution

DimaK123:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3 & days < 7) {
+    answer -= 20;
+  }
+
+  if (days >= 7) {
+    answer -= 50;
+  }
+
+  return answer;
```
marina-tilniak, line 41: add variables for every number

marina-tilniak (changes requested):

src/calculateRentalCost.js
DimaK123: +

ArturErmolenko (approved):

PR #1451, opened 2022-06-12
Initial Solution

olekskov:


valerii-derkach (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const total = days * dayPrice;
+  const firstDiscountPeriod = 3;
+  const firstDiscountAmount = 20;
+  const secondDiscountPeriod = 7;
+  const secondDiscountAmount = 50;
+
+  return days < firstDiscountPeriod ? total
```
alexandra-protyanova, line 38: it isn't good practice to use nested ternary

alexandra-protyanova (changes requested):
Well done, but rewrite without nested ternary

PR #1459, opened 2022-06-18
Solution: js_task-transportation-on-vacation

vyurkov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const dailyRent = 40;
+  const totalPrice = days * dailyRent;
+
+  if (days >= 7) {
+    return totalPrice - 50;
+  }
+
+  if (days >= 3) {
+    return totalPrice - 20;
+  }
```
vkryvytskyy, line 40: move all numeric values to variables

vkryvytskyi (changes requested):

vkryvytskyi (approved):

PR #1466, opened 2022-06-27
Solution

DarkMistyRoom:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
+
+  if (days > 6) {
+    total -= 50;
+  } else if (days > 2) {
+    total -= 20;
+  }
```
Bohdan-Maksymiuk, line 37: make variables for all these numbers

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const bigDiscountDays = 7;
+  const bigDiscountSum = 50;
+
+  let total = days * price;
+
+  if (days >= bigDiscountDays) {
+    total -= bigDiscountSum;
+  } else if (days >= smallDiscountDays) {
+    total -= smallDiscountSum;
+  }
```
Bohdan-Maksymiuk, line 43: better use 2 if's instead of `else if` here

Bohdan-Maksymiuk (approved):
good job 👍

PR #1469, opened 2022-06-28
solution

hserpukhovitina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentalCost = 0;
+
+  if (days < 3) {
+    rentalCost = days * 40;
+  } else {
+    rentalCost = days * 40 - 20;
+  }
```
Bohdan-Maksymiuk, line 37: try to avoid this `if else`
you could calculate base cost and then minus discount (you will need 2 if's)

Bohdan-Maksymiuk (changes requested):
pass all numbers to variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rental = 0;
+  const d = days;
+
+  rental = d * 40;
```
Bohdan-Maksymiuk, line 34:
```suggestion
  let rental = days * 40;
```
no need in `d` variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (d >= 3 && d < 7) {
+    rental -= 20;
+  }
+
+  if (d >= 7) {
+    rental -= 50;
+  }
```
Bohdan-Maksymiuk, line 42: you should transfer all numbers to variables
for ex:
```
const  daysForBigDiscount = 7;
```
same for 3, 20 and 50
```suggestion
  if (days >= 7) {
    rental -= 50;
  }
  
  if (days >= 3) {
    rental -= 20;
  }
```

Bohdan-Maksymiuk (changes requested):

ArturErmolenko (approved):

PR #1472, opened 2022-06-29
Solution

JOWISSA:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  }
```
Bohdan-Maksymiuk, line 41: you could calculate base cost and then minus discount (you will need to if's)

Bohdan-Maksymiuk (changes requested):
and move all numbers to constants

ArtemTopolskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dailyValue = 40;
+  const weekSale = 50;
+  const threeDaysSale = 20;
+  const week = 7;
+  const weekend = 3;
+
+  const weekendRent = days * dailyValue - threeDaysSale;
+  const weekRent = days * dailyValue - weekSale;
```
Bohdan-Maksymiuk, line 38:
```suggestion
```

Bohdan-Maksymiuk: should you calculate it every time?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  const dayRent = days * dailyValue;
+
+  if (days >= week) {
+    return weekRent;
```
Bohdan-Maksymiuk, line 42:
```suggestion
    return dayRent - weekSale;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  }
+
+  if (days >= weekend && days < week) {
+    return weekendRent;
+  }
```
Bohdan-Maksymiuk, line 47:
```suggestion
  if (days >= weekend) {
    return dayRent - threeDaysSale;
  }
```
You have already checked it

Bohdan-Maksymiuk (changes requested):

ArturErmolenko (approved):

PR #1473, opened 2022-06-29
Solution

OksanaShvets1:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
Bohdan-Maksymiuk, line 32: you could make it as a default cost `days * 40` and then minus the discount

Bohdan-Maksymiuk (changes requested):
pass all numbers into variables

Bohdan-Maksymiuk (approved):

PR #1492, opened 2022-07-02
solution

AleksandrYuschenko:


LordAsmodey:
Good solution!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+    amount = days * rent;
+
+    return amount;
+  }
+
+  if (days > 2 && days < 7) {
+    amount += days * rent - discountForWeek;
+
+    return amount;
+  }
```
VETALEK, line 46:
```suggestion
  if (days < 7) {
    amount += days * rent - discountForWeek;

    return amount;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+
+  if (days >= 7) {
+    amount += days * rent - discountMoreWeek;
+
+    return amount;
+  }
```
VETALEK, line 52:
```suggestion
    amount += days * rent - discountMoreWeek;

    return amount;
```

VETALEK (changes requested):
GJ 👍 
but you duplicate some `if` statements (e.g. you don't need `a > 2` after `if` with `a < 3`)

VETALEK (changes requested):

VETALEK (approved):
🔥

PR #1500, opened 2022-07-03
solution

blcts:


DashaPoliazz (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return sum;
+  }
+
+  if (days < 7) {
+    return sum - 20;
+  }
+
+  if (days >= 7) {
+    return sum - 50;
+  }
```
VETALEK, line 43:
```suggestion
  return sum - 50;
```

this `if` is always `true`, otherwise it is not even checked because of previous statement

VETALEK (changes requested):
almost 💪

PR #1502, opened 2022-07-04
task is done

yuriiLevchenko2022:


h4p4l4 (approved):
Good job

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  const rentCost = days * price;
+
+  if (days < 3) {
+    return rentCost;
+  }
+
+  if (days >= 3 && days < 7) {
+    return rentCost - 20;
+  }
```
VETALEK, line 41:
```suggestion
  if (days < 7) {
    return rentCost - 20;
  }
```

`days >= 3` is always `true`, cause of previous statement

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 7) {
+    return rentCost - 50;
+  }
```
VETALEK, line 45:
```suggestion
  return rentCost - 50;
```

`days >= 7` is always `true`, cause of previous statement

VETALEK (changes requested):
Good job 👏 

but you have a bit overcomplicated  `if` statements

src/calculateRentalCost.js
yuriiLevchenko2022: ok

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  const discountAfterThreeDays = 20;
+
+  const discountAfterSevenDays = 50;
+
+  if (days < 3) {
+    return rentCost;
+  }
+
+  if (days < 7) {
```
solaryasha, line 43: You still have "magic" numbers, move them to specific variables

solaryasha (changes requested):

solaryasha (approved):

PR #1504, opened 2022-07-04
Solution

Lentyaeev:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return days * 40; 
+  }
+
+  if(days >= 3 && days < 7){
+    return (days * 40) - 20;
+  }
+
+  if(days >= 7){
+    return (days * 40) - 50;
+  }
```
VETALEK, line 41:
```suggestion
    return (days * 40) - 50;
```

unnecessary `if` statement

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  
 }
+console.log(calculateRentalCost(7));
```
VETALEK, line 44: image

VETALEK (changes requested):
Good start 👍

VETALEK (approved):
🚀

PR #1506, opened 2022-07-04
solution

vitmixG:


yuliia-dronova (approved):
Great! 
Good point in adding vars for period and discount separately but with similar name.

VETALEK (approved):
GJ 🚀

PR #1531, opened 2022-07-05
added solution

AlexanderKolomiiets:


dellachaise:
You could ask one of students which already did this task for students review. And after fixes if needed set it for mentor review.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let total = 0;
+
+  if (days < 3) {
+    total = 40 * days;
+  } else if (days >= 3 && days < 7) {
+    total = (40 * days) - 20;
+  } else if (days <= 7) {
+    total = (40 * days) - 50;
+  }
```
sergii-nosachenko, line 39: 🚨 Your condition is wrong - `days <= 7` is need to be `days >= 7`.
Also this part can be simplified to only 2 options: first check if `days >= 7`, after it if `days >= 3`. Remember, you can immediately return result from inside `if (days >= 7) {return}` and stop the execution.

sergii-nosachenko (changes requested):
Hi, Sasha.
I've checked your code and found some issues. Needs to be fixed.

sergii-nosachenko (approved):
Perfect 🔥

dellachaise (changes requested):
Great job 👍 

But better create meaningful variables for 40, 50, 20 and use it.

dellachaise (approved):
Well done 👍

PR #1543, opened 2022-07-08
Solution

ostUp:


Jamal-Amoian (approved):
Well done!

MaksimusCRS (approved):
You can use just simple if block, without if else

PR #1544, opened 2022-07-09
Task-transportation-on-vacation

MarkShapka:
- Implemented calculateRentalCost function

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,16 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  const COST_DAY = 40;
-  const DISCOUNT_OVER_3_DAYS = 20;
-  const DISCOUNT_OVER_7_DAYS = 50;
+  const priceOneDay = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
 
   if (days < 3) {
-    return days * COST_DAY;
-  } else if (days >= 3 && days < 7) {
-    return (days * COST_DAY) - DISCOUNT_OVER_3_DAYS;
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const COST_DAY = 40;
+  const DISCOUNT_OVER_3_DAYS = 20;
+  const DISCOUNT_OVER_7_DAYS = 50;
```
ArtemTopolskyi, line 33: don't use upper case

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const COST_DAY = 40;
+  const DISCOUNT_OVER_3_DAYS = 20;
```
ArtemTopolskyi, line 32: don't use magic number inside variable name, what it will be 4 days in future
```suggestion
  const smallDiscount = 20;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const DISCOUNT_OVER_7_DAYS = 50;
+
+  if (days < 3) {
+    return days * COST_DAY;
+  } else if (days >= 3 && days < 7) {
```
ArtemTopolskyi, line 37: you don't need 'else' structure, cause you return value in if

ArtemTopolskyi (changes requested):

valerii-derkach (approved):

PR #1556, opened 2022-07-12
solution ver1.0

Oleksii-Korovnichenko:
Hello, take a look pls on my solution and give me feedback. 
It's nothing was said about right input data. If we have negative numbers we should 
make test data (days) and return 0 i think so. What is your opinion?
Thank you for your time).

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = days * 40;
+  let baseDiscount = 0;
+
+  if (days <= 0) {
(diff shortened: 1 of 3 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = days * 40;
+  let baseDiscount = 0;
+
+  if (days <= 0) {
```
alexandra-protyanova, line 34: I guess it's a good option, in this task you will not get invalid data, but I would throw error in that case

alexandra-protyanova (approved):

PR #1563, opened 2022-07-14
Solution

Khabailiuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = 0;
+
+  if (days < 7) {
+    totalAmount = days * 40;
```
alexandra-protyanova, line 34: move main data  to variables

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
Khabailiuk: sorry I do not understand

Khabailiuk: not bacause of English just dont understand what to move where

PR #1564, opened 2022-07-14
Develop

smiyka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
Bohdan-Maksymiuk, line 37: what a magic numbers 🪄 
lets put them into variables and then use
also try to avoid nested `else if` conditions

Bohdan-Maksymiuk (changes requested):

valerii-derkach (approved):

PR #1567, opened 2022-07-15
solution

pelahiia:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const daysSecond = 3;
+
+  if (days >= daysFirst) {
+    return sum - discountFirst;
+  } else if (days >= daysSecond) {
+    return sum - discountSecond;
+  }
+
+  return sum;
 }
```
valerii-derkach, line 45:
```suggestion
  const dayCost = 40;
  const sum = dayCost * days;
  const bigDiscount = 50;
  const basicDiscount = 20;
  const daysFirst = 7;
  const daysSecond = 3;

  if (days >= daysFirst) {
    return sum - discountFirst;
  } 
   if (days >= daysSecond) {
    return sum - discountSecond;
  }

  return sum;
}

```

use more semantic names of variables and pass the [`checklist`](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1572, opened 2022-07-20
Task done

sergey-vorobei:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const discounMin = 20;
+  const discounMax = 50;
+
+  if (days <= 2) {
+    return days * costDay;
+  } else if (days <= 6) {
+    return days * costDay - discounMin;
+  } else if (days >= 7) {
+    return days * costDay - discounMax;
+  }
```
ArturErmolenko, line 41: avoid using if else ,you can  return `days * costDay `by default and use only 2 if

ArturErmolenko (changes requested):

sergey-vorobei:
I understand right?

valerii-derkach (approved):

PR #1575, opened 2022-07-22
calcRental

taras-nakonechnyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let finalPrice = days * 40;
+
+  switch (true) {
```
ArtemTopolskyi, line 33: don't use switch (true), lets' use if construction instead

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+    case days > 6:
+      finalPrice -= 50;
```
ArtemTopolskyi, line 35: create separate variable for every number in the program

ArtemTopolskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  } else if (days >= 3) {
+    finalPrice = priceBeforeDiscont - discontWithThreeDays;
+
+    return finalPrice;
+  } else {
+    finalPrice = priceBeforeDiscont;
+
+    return finalPrice;
+  }
 }
```
valerii-derkach, line 50:
```suggestion
    return priceBeforeDiscont - discontWithSevenDays;

  } 
   if (days >= 3) {

    return priceBeforeDiscont - discontWithThreeDays;
  } 

    return priceBeforeDiscont;
  }
}
```

valerii-derkach (changes requested):
You should return the value immediately in if statements. And don't use `else `after `return `statement. If `return` is performed - code after it won't be executed anyway.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  const pricePerDayRent = 40;
+  const priceBeforeDiscont = days * pricePerDayRent;
+  const discontWithThreeDays = 20;
+  const discontWithSevenDays = 50;
+  let finalPrice;
+
+  if (days >= 7) {
+    finalPrice = priceBeforeDiscont - discontWithSevenDays;
+
+    return finalPrice;
```
marina-tilniak, line 40:
```suggestion
     return priceBeforeDiscont - discontWithSevenDays;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  }
+
+  if (days >= 3) {
+    finalPrice = priceBeforeDiscont - discontWithThreeDays;
+
+    return finalPrice;
```
marina-tilniak, line 46: here

marina-tilniak (changes requested):

valerii-derkach (approved):

PR #1576, opened 2022-07-22
Solution

denys-sadovyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
-}
+  const amount = 0;
+  const pricePerDay = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days <= 2) {
+    return days * pricePerDay;
+  } else if (days <= 6) {
```
ArtemTopolskyi, line 38: you don't need else, you anyway return value if prev condition

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const amount = 0;
+  const pricePerDay = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days <= 2) {
```
ArtemTopolskyi, line 36: move 2, 6, 7 to variables too

ArtemTopolskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+
+  if (days <= twoDays) {
+    return days * pricePerDay;
+  }
+
+  if (days <= sixDays) {
+    return (days * pricePerDay) - smallDiscount;
+  }
 
+  if (days >= sevenDays) {
```
alexandra-protyanova, line 47: u can return it as a default value

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1582, opened 2022-07-25
Calculate rental cost has been created

antonenko-code:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costsPerDay = 40;
+  const discount = (days >= 7) ? 50 : (days >= 3) ? 20 : 0;
```
alexandra-protyanova, line 32: don't nested ternaries, I like the idea but it is not a good practice

alexandra-protyanova (changes requested):

valerii-derkach (approved):

PR #1588, opened 2022-07-28
Solution

Walintigo:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
alexandra-protyanova, line 32: add variables for discounts

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3 && days < 7) {
```
alexandra-protyanova, line 39: you can write this code without if

alexandra-protyanova (changes requested):

Bohdan-Maksymiuk (approved):

PR #1589, opened 2022-07-28
Solution

YuliaDzisiak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  } else if (days < 7) {
+    rentalCost = (costPerDay * days) - discaunt3Days;
+  } else {
+    rentalCost = (costPerDay * days) - discaunt7Days;
+  }
+
+  return rentalCost;
 }
 
 module.exports = calculateRentalCost;
```
valerii-derkach, line 47: Avoid reassigning variables values. It's better to use `const` instead of `let`.
You should` return `the value immediately in `if `statements. And don't use `else `after `return` statement. If `return `is performed - code after it won't be executed anyway

valerii-derkach (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const discaunt7Days = 50;
+
+  if (days < 3) {
+    return costPerDay * days;
+  } else if (days < 7) {
+    return costPerDay * days - discaunt3Days;
+  } else {
+    return costPerDay * days - discaunt7Days;
+  }
 }
```
valerii-derkach, line 42:
```suggestion
  if (days < 3) {
    return costPerDay * days;
  } 
   if (days < 7) {
    return costPerDay * days - discaunt3Days;
  } 
    return costPerDay * days - discaunt7Days;
  
}
```

valerii-derkach (changes requested):

Bohdan-Maksymiuk (approved):

PR #1590, opened 2022-07-29
Solution

diakovMax:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+
+  if ((days >= 3) && (days < 7)) {
+    amount = (40 * days) - 20;
+  }
+
+  if (days >= 7) {
+    amount = (40 * days) - 50;
+  }
+
+  return amount;
```
valerii-derkach, line 46: Create variables for the values so that you can reuse them and give them descriptive names.
Avoid reassigning variables values. It's better to use `const` instead of `let`
You should return the value immediately in if statements. And don't use `else` after `return `statement.

valerii-derkach (changes requested):

src/calculateRentalCost.js
diakovMax: If I understood you correctly, it should be correct now

src/calculateRentalCost.js
```diff
@@ -29,6 +29,27 @@
...
   // write code here
+  const carCost = 40;
+  const discountForThreeDays = 20;
+  const discountForSevenDays = 50;
+
+  if (days < 3) {
+    const amount = (carCost * days);
+
+    return amount;
+  }
```
valerii-derkach, line 40:
```suggestion
  const discountForSevenDays = 50;
  const amount = (carCost * days);

  if (days < 3) {
     return amount;
  }
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,27 @@
...
+
+  if ((days >= 3) && (days < 7)) {
+    const amount = (carCost * days) - discountForThreeDays;
+
+    return amount;
+  }
```
valerii-derkach, line 46:
```suggestion
  if ((days >= 3) && (days < 7)) {
    return amount  - discountForThreeDays;
  }
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,27 @@
...
+
+  if (days >= 7) {
+    const amount = (carCost * days) - discountForSevenDays;
+
+    return amount;
+  }
 }
```
valerii-derkach, line 53:
```suggestion
  if (days >= 7) {
    return amount  - discountForSevenDays;
  }
}
```

valerii-derkach (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+  const discountForSevenDays = 50;
+  const amount = carCost * days;
+
+  if (days < 3) {
+    return amount;
+  }
+
+  if ((days >= 3) && (days < 7)) {
+    return amount - discountForThreeDays;
+  }
```
valerii-derkach, line 43:
```suggestion
  if (days < 7) {
    return amount - discountForThreeDays;
  }
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+
+  if (days >= 7) {
+    return amount - discountForSevenDays;
+  }
```
valerii-derkach, line 47:
```suggestion

    return amount - discountForSevenDays;

```

valerii-derkach (approved):

PR #1591, opened 2022-07-29
calculateRentalCost solution

yurii-kozlov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+      sum -= 20;
+
+      return sum;
+
+    case days >= 7:
+      sum -= 50;
+
+      return sum;
+  }
 }
```
valerii-derkach, line 47: let's create variables for all these numbers, to understand what is going on here
image

valerii-derkach (changes requested):

marina-tilniak (approved):

PR #1599, opened 2022-08-01
Solution

artem-kodesnikov:


dellachaise:
You could ask one of students which already did this task for students review. And after fixes if needed set it for mentor review.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
sergii-nosachenko, line 31:
```suggestion
  let result = days * 40;
```

You can perform this repeated calculation only once before checking discount.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    result = days * 40;
+  } else if (days < 7) {
+    result = (days * 40) - 20;
+  } else {
+    result = (days * 40) - 50;
+  }
```
sergii-nosachenko, line 39: This part can be simplified to only 2 options: first check if `days >= 7`, after it if `days >= 3`. Remember, you can immediately return result from inside `if (days >= 7) {return}` and stop the execution.

sergii-nosachenko (changes requested):
Hi, Artem! ✌
You can improve your code just a bit to make it more consistent and clear.

src/calculateRentalCost.js
sergii-nosachenko, line 31: This can be named as `cost` for example.

sergii-nosachenko (approved):
I like it! 🔥

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = days * 40;
+
+  if (days >= 7) {
+    return price - 50;
+  } else if (days >= 3) {
+    return price - 20;
```
danylo-onishchuk, line 36: what is 50 or 20?

danylo-onishchuk (changes requested):

danylo-onishchuk (approved):

PR #1600, opened 2022-08-01
Solution

Dmytro-Matsehora:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  switch (true) {
+    case (days < 3):
+      return 40 * days;
+
+    case (days >= 3 && days < 7):
+      return 40 * days - 20;
+
+    case (days >= 7):
+      return 40 * days - 50;
+  }
```
tiserett, line 40:
```suggestion
  switch (true) {
    case (days < 3):
      return 40 * days;

    case (days >= 3 && days < 7):
      return 40 * days - 20;

    case (days >= 7):
      return 40 * days - 50;
  }
```

you should create variable for discount, adjust it's value in each case, and then return value from function, you should also create variable for daily rent price

tiserett (changes requested):
a few little changes

src/calculateRentalCost.js
```diff
@@ -28,15 +28,18 @@
...
-      return 40 * days;
+      return totalAmount;
 
     case (days >= 3 && days < 7):
-      return 40 * days - 20;
+      return totalAmount - discount;
 
     case (days >= 7):
-      return 40 * days - 50;
+      return totalAmount - discount * 2.5;
```
tiserett, line 42: you should set discount default value as 0, change it's value in each case, and then return dailyRent * days - discoun

tiserett (changes requested):
still not how I expect it to be, but much better

tiserett (approved):
good job! ;)

danylo-onishchuk (approved):

PR #1608, opened 2022-08-03
Solution

kir8mir:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,15 @@
...
+function calculateRentalCost(day) {
+  switch (true) {
+    case (day >= 3 && day < 7):
+      return 40 * day - 20;
+    case (day >= 7):
+      return 40 * day - 50;
+    default:
+      return 40 * day;
+  }
 }
```
tiserett, line 39:
```suggestion
function calculateRentalCost(day) {
  switch (true) {
    case (day >= 3 && day < 7):
      return 40 * day - 20;
    case (day >= 7):
      return 40 * day - 50;
    default:
      return 40 * day;
  }
}
```
solution with 'switch statement' looks complicated, use 'ternary' operator or 'if/else' statement instead

tiserett (changes requested):
few little fixes required

src/calculateRentalCost.js
kir8mir: I would like to train my switch skill, so could this code be like this?

tiserett (approved):

src/calculateRentalCost.js
```diff
@@ -27,8 +27,15 @@
...
-  // write code here
+function calculateRentalCost(day) {
+  switch (true) {
+    case (day >= 3 && day < 7):
+      return 40 * day - 20;
+    case (day >= 7):
+      return 40 * day - 50;
+    default:
+      return 40 * day;
+  }
```
danylo-onishchuk, line 38: solution with 'switch statement' looks overcomplicated

danylo-onishchuk: better way is if

danylo-onishchuk (changes requested):

danylo-onishchuk (approved):

PR #1610, opened 2022-08-03
Solution

zhenija:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let total = days * 40;
+
+  if (days >= 7) {
+    total = total - 50;
+  }
+
+  if (days >= 3 && days < 7) {
+    total = total - 20;
+  }
```
Anna-Tkachenko, line 39: 40, 50, 20 - magic numbers for now
Will be better to create variables to store these values and use these variables in code

Anna-Tkachenko: point 1 in [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
+
+  if (days >= 7) {
+    total = total - 50;
```
Anna-Tkachenko, line 34: You should also return value immediately in if statements (point 3 in [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md))

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -26,9 +26,23 @@
  *
  * @return {number}
  */
+const rentalCost = 40;
+const smallOff = 20;
+const bigOff = 50;
```
Anna-Tkachenko, line 31: You can move it inside calculateRentalCost function

Anna-Tkachenko (approved):

PR #1617, opened 2022-08-04
Solution

Roman-Artiukh:


oksana-logos-frontend (approved):
good)

Roman-Artiukh:
thanks))

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
   // write code here
+
+  switch (true) {
+    case days >= 3 && days < 7:
+      return days * 40 - 20;
+    case days >= 7:
+      return days * 40 - 50;
+    default:
+      return days * 40;
+  }
```
MaksimusCRS, line 40: U can use simple if block, switch redundant here

MaksimusCRS (changes requested):

MaksimusCRS (approved):
And add variables, get rid of "magic numbers"

PR #1619, opened 2022-08-04
Solution

illia-kots:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const total = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return total - 20;
+  }
+
+  if (days > 6) {
+    return total - 50;
+  }
```
Bohdan-Maksymiuk, line 39: what a magic numbers (3, 7, 20, 50)
lets pass them into variables and name them properly what they mean

Bohdan-Maksymiuk (changes requested):

alexandra-protyanova (approved):

PR #1622, opened 2022-08-04
Solution

anna-kolmakova:


max-kravchenko (approved):
Great!

dellachaise (changes requested):
Well done! 👍

1. But better create  variables for 40, 20, 50 , to understand what does them mean.
2. `days * 40` you use 3 times, better also create a variable.
3. Better 2 last `if` rewrite to `else if`.

dellachaise (approved):
Well done 👍 

But, better rewrite without using variable assigned with `let`, and return value in each `if staitment` - according to  checklist. 
Screenshot 2022-08-06 at 00 18 19

PR #1623, opened 2022-08-04
solution

anyakrainiuk:


zinchenko-anastasiia (approved):

dellachaise (changes requested):
Great job 👍 

But better create meaningful variables for 40,`days * 40`, 50, 20 and use it.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const sumCost = days * 40;
```
dellachaise, line 31: better rename more meaningful way.
```suggestion
  const pricePerDay = 40;
  const totalPrice = pricePerDay * days;
```

dellachaise (approved):
Well done 👍

src/calculateRentalCost.js
anyakrainiuk: Fixed!

PR #1624, opened 2022-08-05
solution

lizapolokhina:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
   // write code here
+  let cost = 0;
+
+  if (days >= 3 && days < 7) {
+    cost = days * 40 - 20;
+  } else if (days >= 7) {
+    cost = (days * 40) - 50;
+  } else {
+    cost = days * 40;
+  }
```
plemiannik, line 40: not sure if else need here

plemiannik (approved):
pretty nice 👍

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let cost = 0;
+
+  if (days >= 3 && days < 7) {
```
AndriiRut, line 34: checklist #1

AndriiRut (changes requested):

src/calculateRentalCost.js
```diff
@@ -31,12 +31,12 @@ function calculateRentalCost(days) {
   // write code here
   let cost = 0;
 
-  if (days >= 3 && days < 7) {
+  if (days >= 0 && days <= 2) {
+    cost = days * 40;
```
AndriiRut, line 35: this action `days * 40` you can write once on the 32th line and use only a variable inside your IFs. If you use the same action or number twice this is a reason to reuse it  as a variable.

the cost without any discounts can be returned by default if any conditions are `false`. 

In addition, you can use only 2 `if` if in the first one you'll check the biggest number of days, in the second you'll check some mild numbers of days and if these two `if` will be false you return the default cost.

AndriiRut (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const cost = days * 40;
+
+  if (days >= 7) {
+    return cost - 50;
+  } else if (days >= 3 && days <= 6) {
```
AndriiRut, line 36:
```suggestion
  }
  
  if (days >= 3) {
```

you don't need this extra check. you already checked this in first if and return(stop function) if `days >= 7`.

AndriiRut (approved):
🔥

PR #1625, opened 2022-08-05
Solution

Desertiks:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rent = 0;
+
+  for (let i = 0; i < days; i++) {
+    rent += 40;
+  }
```
plemiannik, line 35: u can just days*40

plemiannik (approved):
good work 👍

src/calculateRentalCost.js
Desertiks: Agreed)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  if (days >= 3) {
+    rent -= 20;
+
+    return rent;
+  }
+
+  return rent;
 }
 
+calculateRentalCost(5);
```
AndriiRut, line 50: why do you need to call this function? 🤔

AndriiRut (changes requested):

src/calculateRentalCost.js
Desertiks: Forgot to remove(

AndriiRut (approved):

PR #1627, opened 2022-08-05
Solution

vladgusakovskiy:


OlgaOchichenko (approved):
Good Job!

MaksimusCRS (approved):
You can use just if block without else

PR #1634, opened 2022-08-05
task solution code

rboiar:


Roman-Artiukh (changes requested):
try not to use a lot of else,  recommend using only if or switch.
http://joxi.ru/l2Zzj9Kclxklg2

rboiar:
> try not to use a lot of else, recommend using only if or switch. http://joxi.ru/l2Zzj9Kclxklg2

Ok, thanks

MaksimusCRS (approved):

PR #1636, opened 2022-08-05
1task Solution

Anastasiia-Khmelovska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
```
YuriiPronin, line 30: your solution is good. But if i look at this code in few weeks, i think i cant understand all these numbers and why we substraction them. maybe you can do more variables?

YuriiPronin (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
-  // write code here
-}
+  const basicPrice = 40;
+  const basicDiscount = 50;
+  const minDiscount = 20;
+  let rental = 0;
```
danylo-onishchuk, line 34:
```suggestion
  let rental;
```

danylo-onishchuk (approved):

PR #1637, opened 2022-08-05
rental cost task done

Andrii-Dunets:


Anatolii-Bukoros (approved):
It would be nice to put price and discount in variables

MaksimusCRS (approved):

PR #1646, opened 2022-08-07
solution

YuriiPronin:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let fullCoast = 0;
+  const coastOneDay = 40;
+  const coastWithoutOffer = days * coastOneDay;
+
+  if (days < 3) {
+    fullCoast = coastWithoutOffer;
+  }
```
plemiannik, line 37: Maybe, no need in const coastWithoutOffer?
  
```
  const coastOneDay = 40; 
  let fullCoast = days * coastOneDay;

  if (days >= 3 && days < 7) {
    fullCoast -= 20;
  }

  if (days >= 7) {
    fullCoast -= 50;
  }
```

but i not sure. and in that case, if u don`t have const or let 4 -20 and -50, than why u have one 4 price per day? so, maybe, u should get consts 4 day price and discounts or do not create any.

plemiannik (approved):
Good job!👍

AndriiRut (approved):

PR #1647, opened 2022-08-07
1-st commit

nikolay-li:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const rent = 40;
+
+  if (days < 3) {
+    return days * rent;
+  } else if (days < 7) {
```
plemiannik, line 36: not sure is there is a reason in else here

plemiannik (approved):
nice job 👍

danylo-onishchuk (approved):

PR #1649, opened 2022-08-07
Solution

Mejeezz:


ibondrk (approved):
Goog job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  if (days >= 7) {
+    return rentPrice * days - bigDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
+    return rentPrice * days - smallDiscount;
+  }
+
+  return rentPrice * days;
 }
```
tiserett, line 44:
```suggestion
  const rentPrice = 40;
  let discount = 0;

  if (days >= 7) {
    discount = 50;
  } else if (days >= 3) {
    discount = 20;
  }

  return rentPrice * days - discount;
}
```

if there's gonna be more discounts you'll have to create different variables.
You can create one variable - discount, and adjust it's value in each if case, then return value from function.
With my variation of code you can simply add different values to discount, which allows you to get rid of unnecessary variables.

tiserett (changes requested):
would be nice to create 1 variable for discount and change it's value in different cases

tiserett (approved):
good job!

MaksimusCRS (approved):

PR #1652, opened 2022-08-07
solution

RonnieFixxer:


Alzay007 (approved):
i think you can use in this task only 'if' . When we use construction if else. last child must be else (but i not sure, better check information)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * 40;
+
+  if (days >= 7) {
+    return result - 50;
+  } else if (days >= 3) {
```
plemiannik, line 35: why u use else here

plemiannik (approved):
nice 👍

AndriiRut (changes requested):

AndriiRut (approved):

PR #1655, opened 2022-08-08
solution

vadym-zahazei:


plemiannik (approved):
wow, looking tricky 👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  switch (true) {
+    case days < 3:
+      return 40 * days;
```
AndriiRut, line 33: good! but check the checklist#1

AndriiRut (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalDayCost = 40;
+  const basicDiscont = 20;
+  const extraDiscont = 50;
+
+  switch (true) {
+    case days < 3:
+      return rentalDayCost * days;
```
AndriiRut, line 37: good but why you don't use a variable for this multiplication?

```suggestion
      return rentalDayCost * days; // const staticCost = rentalDayCost * days; (use your own naming for it)
```

AndriiRut (changes requested):

src/calculateRentalCost.js
vadym-zahazei: > good but why you don't use a variable for this multiplication?

didn't think of it(

AndriiRut: remember to use variables for any action or dials which you used 2 or more times)

AndriiRut (approved):

PR #1660, opened 2022-08-08
Solution

David-Shahmatov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const bigDiscount = 50;
+
+  if (days >= 7) {
+    amount += rentalCost - bigDiscount;
+  } else if (days >= 3) {
+    amount += rentalCost - smallDiscount;
+  } else {
+    amount += rentalCost;
+  };
+
```
plemiannik, line 43: why u use ellse instead just if + u can return imidiatly

plemiannik (approved):
nice work 👍

AndriiRut (approved):
🚀

PR #1663, opened 2022-08-09
solution added

volodymir-tymtsias:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
alexandra-protyanova, line 32: let's move calculations to variables to everyone could understand what is happening

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1664, opened 2022-08-09
Solution

miamarichka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
```
alexandra-protyanova, line 31: move all discounts to variables to make your code more understandable

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
GJ!

PR #1665, opened 2022-08-09
 js_task-transportation-on-vacation done

EhorSaprykin:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = 0;
+  const dayCost = 40;
+
+  if (days < 3) {
+    cost = days * dayCost;
+  } else if (days >= 3 && days < 7) {
+    cost = days * dayCost - 20;
```
alexandra-protyanova, line 37: move discounts to variables as well

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
EhorSaprykin: @alexandra-protyanova done

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  let cost = 0;
+  const dayCost = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    cost = days * dayCost;
+  } else if (days >= 3 && days < 7) {
+    cost = days * dayCost - smallDiscount;
+  } else {
```
alexandra-protyanova, line 40: u can write one of part of code in `if` without condition

alexandra-protyanova (approved):

PR #1669, opened 2022-08-10
Solution

mary-kucher:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const allCount = days * 40;
+  let result = 0;
+
+  if (days < 3) {
+    result = allCount;
+  } else if (days >= 3 && days <= 6) {
+    result = allCount - 20;
+  } else {
```
alexandra-protyanova, line 38: u can set it as a default value without else

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1671, opened 2022-08-11
Solution

MaxManis:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let res = 0;
+
+  for (let i = 1; i <= days; i++) {
+    res += 40;
+  }
+
+  return days >= 7 ? res - 50 : days >= 3 ? res - 20 : res;
```
Anna-Tkachenko, line 37: you should write clear and understandable code 🙂
will be better to use ternary operator only in case when you need to store one of two values depending on some condition

in this task will be more understandable to use if / else statement

and for now 50, 40, 20 - magic numbers
create variables for them

each of this rules also are on checklist, don't forget to check it)

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  let result = 0;
+
+  for (let i = 1; i <= days; i++) {
+    result += 40;
+  }
+
+  if (days >= 7) {
+    result -= 50;
+  } else if (days >= 3) {
+    result -= 20;
```
Denchepornyuk, line 40: 40, 50, and 20 still are magic numbers
try to keep all numbers like these in variables for more readable code

denys-cheporniuk (approved):

PR #1678, opened 2022-08-15
Solution

dashaprotsko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  }
+
+  if (days < 7) {
+    return 40 * days - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
murat-yesimbietov, line 39: It's better to not use else block

murat-yesimbietov (approved):
Do not add the "package-lock.json" file to the pull request

PR #1682, opened 2022-08-15
Solution

illia-sosnytskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = 40 * days;
+
+  if (days >= 3 && days < 7) {
+    return result - 20;
```
Andr1yK, line 34: why 20? and what is 20?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days >= 7) {
+    return result - 50;
```
Andr1yK, line 36: same here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = 40 * days;
+
+  if (days >= 3 && days < 7) {
+    return result - 20;
+  } else if (days >= 7) {
```
Andr1yK, line 35: Can u do this without `else if`?)

Andr1yK (changes requested):

Andr1yK (changes requested):
[point 5](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

silchyn (changes requested):
By changing "3" to "three" you don't solve the problem

naziamoff (approved):

PR #1689, opened 2022-08-15
Solution

LenaPatrashok:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
```
oablokhina, line 32: it would be better to create a variable with this value because you use it several times

oablokhina (changes requested):

oablokhina (approved):

PR #1690, opened 2022-08-15
Solution

AndrewGolovnykh:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days >= 7) {
+    return (days * 40 - 50);
```
oablokhina, line 33: brackets are not required here, remove them.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
+  }
+
+  if (days >= 3) {
+    return (days * 40 - 20);
+  } else {
+    return (days * 40);
```
oablokhina, line 39: create a variable with this value, because you use it several times

oablokhina (changes requested):

src/calculateRentalCost.js
AndrewGolovnykh: Changes were made.

src/calculateRentalCost.js
AndrewGolovnykh: Done.

oablokhina (approved):

PR #1691, opened 2022-08-15
Solution

alinashults:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-}
+  if (days >= 7) {
+    return days * 40 - 50;
+  };
+
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  };
+
+  return days * 40;
```
oablokhina, line 39: create variable with this value, because you use it several times

oablokhina (changes requested):

oablokhina (approved):

PR #1693, opened 2022-08-15
Solution

Anna-Kononuchenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days === 1 || days === 2) {
```
murat-yesimbietov, line 31: Use <> operators instead of enumerating all possible options
```suggestion
  if (days < 3) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+    return 40 * days;
+  }
+
+  if (days === 3 || days === 4 || days === 5 || days === 6) {
```
murat-yesimbietov, line 35: since after the `return` operator the function finishes its work, you can use  `days < 7`

murat-yesimbietov (changes requested):
Do not add 'package-lock.json' and 'node-modules' to the PR

PR #1694, opened 2022-08-15
Solution Elina Nechaieva

ElinaNechaieva:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return +`${days * 40}`;
```
oablokhina, line 32: why do you use interpolation? you don't need it in this case

oablokhina: create a variable with a value days*40, because you use it several times

oablokhina (changes requested):

oablokhina (approved):

PR #1695, opened 2022-08-15
Solution

Olha-Moroziuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  if (days >= 7) {
+    return (40 * days - 50);
+  }
+
+  if (days >= 3) {
+    return (40 * days - 20);
+  }
+
+  if (days < 3) {
+    return (40 * days);
```
oablokhina, line 40: create a variable with this value, because you use it several times

oablokhina (changes requested):

oablokhina (approved):

PR #1696, opened 2022-08-15
Solution

YehorVoroniuk:


oablokhina:
your tests are failed. try again

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+  }
+
+  if (days >= 3 && days <= 6) {
+    totalCost = dayCost * days - 20;
+
+    return totalCost;
+  }
+
+  if (days >= 7) {
+    totalCost = (dayCost * days) - 50;
```
oablokhina, line 47: It would be better to create a variable with a value days * 40, because you use it several times

oablokhina (changes requested):

oablokhina (approved):

PR #1697, opened 2022-08-15
Solution

nicatino:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
oablokhina, line 32: add a variable with this value, because you use it several times

oablokhina (changes requested):

src/calculateRentalCost.js
oablokhina: add days * 40 value to a variable, and then use it.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let answer;
+
+  if (days < 3) {
+    answer = days * 40;
```
oablokhina, line 34: add days * 40 value to a variable, and then use it.

oablokhina (changes requested):

djtyrf312 (approved):

PR #1699, opened 2022-08-15
solution

BogdanHubQA:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const Rental = 40;
+  let totalAmount = Rental * days;
+
+  if (days >= 3) {
+    totalAmount = Rental * days - 20;
```
oablokhina, line 35: it would be better to create a variable like 'discountAmount' and use your totalAmount for counting it

oablokhina (changes requested):

PR #1700, opened 2022-08-15
solution

PetrekoIvan:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const Rental = 40;
+  let totalAmount = Rental * days;
+
+  if (days >= 3) {
+    totalAmount = Rental * days - 20;
```
oablokhina, line 35: Add a variable like 'discountAmount' and use your variable 'totalAmount' for counting

oablokhina (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const Rental = 40;
+  let totalAmount = Rental * days;
```
oablokhina, line 32: Add a variable like 'discountAmount' and use your variable 'totalAmount' for counting

oablokhina (changes requested):

PR #1702, opened 2022-08-15
Solution

InsultedCaramel:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = 0;
+
+  for (let i = 0; i < days; i++) {
```
oablokhina, line 33: You don't need a loop for this task. it's not optimized.

oablokhina (changes requested):

PR #1705, opened 2022-08-16
Solution

serhii-serenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
+    return days * 40 - 20;
+  }
+
+  return days * 40;
```
oablokhina, line 39: add a variable with this value, because you use it several times

oablokhina (changes requested):

oablokhina (approved):

PR #1706, opened 2022-08-16
Solution

Valerii-R:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 1 && days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
murat-yesimbietov, line 37: Let's do the same without the `else` block

murat-yesimbietov (changes requested):
Need some changes and do not add package-lock.json and node_modules to the future PR

PR #1709, opened 2022-08-16
Solution

anastasiya-gutsal:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,12 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = days * 40;
```
oablokhina, line 31: use 'const'. This variable doesn’t change

oablokhina (changes requested):

PR #1710, opened 2022-08-17
Solution

Anastasiia-Lisova:


oablokhina:
Your tests are failed. try again

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (3 <= days <= 6) {
+    return days * 40 - 20
+  }
+  if (days => 7) {
+    return days * 40 - 50
+  }
+  return days * 40
```
oablokhina, line 37: create a variable with this value, because you use it several times

oablokhina (changes requested):

Anastasiia-Lisova:
What do I need to do?

src/calculateRentalCost.js
oablokhina: you don't change it)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days > 2 && days < 7) {
+    return days * 40 - 20;
+  }
+
+  if (days > 6) {
+    return days * 40 - 50;
+  }
+
+  return days * 40;
```
oablokhina, line 39: you don't change it) pay attention to my second comment

oablokhina (changes requested):

oablokhina (approved):

PR #1713, opened 2022-08-18
calculateRental added

Yaroslav-Radchuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
+    return days * 40 - 20;
+  }
```
Bohdan-Maksymiuk, line 41: what a magic numbers 🪄 

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const count = days * 40;
+
+  if (days >= 7) {
+    return (count - 50);
+  }
+
+  if (days >= 3) {
+    return (count - 20);
+  }
```
Bohdan-Maksymiuk, line 39: same for discount amounts and days count

Bohdan-Maksymiuk (changes requested):

alexandra-protyanova (approved):

PR #1719, opened 2022-08-21
Solution

vitalii-borzov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayRentCost = 40;
+  let rentPrice = oneDayRentCost * days;
+
+  if (days < 3) {
+    rentPrice = (oneDayRentCost * days);
+  } else if (days < 7) {
+    rentPrice = (oneDayRentCost * days) - 20;
```
SanyaBratashchuk, line 37: You can return directly from the if block and get rid of the else if. Also It will be much better if you create external variables for discounts, It will be much cleaner)

SanyaBratashchuk (changes requested):
Good start)

src/calculateRentalCost.js
vitalii-borzov: "You can return directly from the if block and get rid of the else if" - what you means?

alexandra-protyanova: he meant u can move the logic of one of the ifs to the main code & nothing will change

alexandra-protyanova (changes requested):
fix previous comments

TarasHoliuk (approved):
GJ! Approved

PR #1722, opened 2022-08-23
Solution

offsset:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const day = 40;
```
SanyaBratashchuk, line 31:
```suggestion
  const costPerDay = 40;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  switch (true) {
+    case (days < 3):
+      return day * days;
+
+    case (days < 7):
+      return day * days - 20;
+
+    case (days >= 7):
+      return day * days - 50;
```
SanyaBratashchuk, line 41: create variables for discounts

SanyaBratashchuk (changes requested):

alexandra-protyanova (approved):
Good job, but `switch` is overkilling here)

PR #1723, opened 2022-08-23
Solution

Oleksii-Blyzniuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  let total = days * dayPrice;
+
+  if (days >= 3 && days < 7) {
+    total -= 20;
+  }
+
+  if (days >= 7) {
```
alexandra-protyanova, line 38: u don't need this if here, this case can be default

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1725, opened 2022-08-24
Solution

AMahalias:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rentOfDay = 40;
+  let total = days * rentOfDay;
+
+  if (days >= 3 && days <= 6) {
+    total = total - 20;
+  }
+
+  if (days >= 7) {
```
alexandra-protyanova, line 38: u can not to use one of the `if` just write one of the code without it

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1728, opened 2022-08-26
solution implemented

Naveed-Shoukat:
Required JS solution implemented and it's working for all test cases.

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -35,11 +35,13 @@ function calculateRentalCost(days) {
 
   if (days >= 7) {
     return totalCarRent - maxDiscount;
-  } else if (days >= 3) {
-    return totalCarRent - minDiscount;
-  } else {
-    return totalCarRent;
   }
+
+  if (days >= 3) {
+    return totalCarRent - minDiscount;
+  }
+
+  return totalCarRent;
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -30,11 +30,13 @@
 function calculateRentalCost(days) {
   const dailyCarRent = 40;
   const totalCarRent = days * dailyCarRent;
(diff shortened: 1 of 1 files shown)


Quelaan1 (approved):
Good job

src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dailyCarRent = 40;
+  const totalCarRent = days * dailyCarRent;
+
+  return days >= 7
+    ? totalCarRent - 50
+    : days >= 3
+      ? totalCarRent - 20
+      : totalCarRent;
```
VinnieJ-2k20, line 38: Don't write nested ternary operators.
This way it wirks, but extremely hard to read.
Replace with `if` structure, and don't forget to read checklist.

VinnieJ-2k20 (changes requested):

src/calculateRentalCost.js
Naveed-Shoukat: ternary replaced with if else block.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const maxDiscount = 50;
+  const minDiscount = 20;
+
+  if (days >= 7) {
+    return totalCarRent - maxDiscount;
+  } else if (days >= 3) {
+    return totalCarRent - minDiscount;
+  } else {
+    return totalCarRent;
+  }
```
VinnieJ-2k20, line 42: Checklist item 3

VinnieJ-2k20 (changes requested):
Almost done.

src/calculateRentalCost.js
Naveed-Shoukat: if block changed

VinnieJ-2k20 (approved):
Good job 👍

PR #1737, opened 2022-08-31
Solution

devWizHub:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+    number = days * 40;
+  } else if
+  (days < 7) {
+    number = (days * 40) - 20;
+  } else {
+    number = (days * 40) - 50;
+  }
 
+  return number;
+}
```
amitpatiljc, line 44: let's not have magic numbers in the code. at the moment i dont understand what 20, 40 and 50 are
You have repeated `days * 40` a few times. maybe put it in a variable?

amitpatiljc (changes requested):
Good Job! just one change needed

src/calculateRentalCost.js
devWizHub: implemented

amitpatiljc: and yes one variable for this `days * dailycost`
`let totalCost = days * dailycost`

this looks good right 
 `actualCost = (totalCost) - disountForThreeDays `
 (may be change `number` variable as well)

amitpatiljc (changes requested):

src/calculateRentalCost.js
devWizHub: implemented

amitpatiljc (approved):
Good Job!

PR #1738, opened 2022-09-01
Solution

omerfarukbalik:


ghost (changes requested):
Read the checklist, you need to declare variables and use them in the if statement

VinnieJ-2k20 (changes requested):
Almost done.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const price = 40;
+  const totalPrice = days * price;
+  const smallDiscont = 20;
+  const bigDiscount = 50;
+
+  if (days < 3 && days > 0) {
+    return totalPrice;
+  } else if (days >= 3 && days < 7) {
+    return totalPrice - smallDiscont;
+  }
```
VinnieJ-2k20, line 40: 1. All else statements should be removed.
2. Find a way to use single conditions instead of double, e.g. `days >= 7`, `days >= 3` etc.

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):

PR #1740, opened 2022-09-02
Task calculate Rental Cost

ArtemKravcov85:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  let result = 0;
+  const costs = 40;
+
+  if (days < 3) {
+    result = days * costs;
+  } else if (days < 7) {
+    result = days * costs - 20;
+  } else {
+    result = days * costs - 50;
+  }
```
Bohdan-Maksymiuk, line 40: avoid such nested `if else` conditions

Bohdan-Maksymiuk: and what a magic numbers 🪄

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const costs = 40;
+  let result = days * costs;
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
+
+  if (days >= 3 && days < 7) {
+    result -= discountThreeDays;
+  }
+
+  if (days >= 7) {
```
Bohdan-Maksymiuk, line 40: pass these numbers into variables too

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costs = 40;
+  let result = days * costs;
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
```
Bohdan-Maksymiuk, line 34: what if we will change these days numbers - we should rename variables then - so fix names in some way

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 3 && days < 7) {
+    result -= discountThreeDays;
```
Bohdan-Maksymiuk, line 37: you can directly return into condition

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
ArtemKravcov85: По последнему замечанию не понял, что имеется ввиду

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costs = 40;
+  let result = days * costs;
+  const startOfRent = 3;
+  const endOfRent = 7;
+  const discountFerst = 20;
+  const discountSecond = 50;
```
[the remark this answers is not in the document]
Bohdan-Maksymiuk: minDaysForSmallDiscount

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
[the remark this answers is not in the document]
ArtemKravcov85: thanks!

alexandra-protyanova (approved):

PR #1742, opened 2022-09-02
working

dmitriyn3679:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    case days >= 7:
+      total = days * 40 - 50;
+      break;
+
+    case days >= 3:
+      total = days * 40 - 20;
+      break;
+
+    default:
+      total = days * 40;
```
Bohdan-Maksymiuk, line 43: what a magic numbers 🪄

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  const oneDayRentCost = 40;
+  const cashbackFromThreeDays = 20;
+  const cashbackFromSevenDays = 50;
+
+  switch (true) {
+    case days >= 7:
+      total = days * oneDayRentCost - cashbackFromSevenDays;
+      break;
+
+    case days >= 3:
```
Bohdan-Maksymiuk, line 41: make variables for these days too

Bohdan-Maksymiuk (changes requested):
good job, fix comment

Bohdan-Maksymiuk (approved):

PR #1746, opened 2022-09-05
solution

bugsworld38:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+
+  let total = days * dayPrice;
+
+  if (days >= 7) {
+    total -= bigDiscount;
+  } else if (days >= 3) {
+    total -= smallDiscount;
+  }
```
naziamoff, line 41: Avoid `else if `when it is not necessary. 
And you can return the value right away after you've found the right condition

```suggestion
  if (days >= 7) {
    return total - bigDiscount;
  }  
  
  if (days >= 3) {
    return total - smallDiscount;
  }
```

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #1748, opened 2022-09-05
made function calculateRentalCost

Hlazman:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  let sum = 0;
+  const price = 40;
+
+  if (days > 2 && days < 7) {
+    discount = 20;
+  }
+
+  if (days >= 7) {
+    discount = 50;
+  }
```

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,29 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const stock = {
+    firstDiscount: 20,
+    secondDiscount: 50,
+  };
+  const term = {
+    twoDays: 2,
+    moreWeek: 7,
```
alexandra-protyanova, line 37:
```suggestion
    week: 7,
```

alexandra-protyanova (approved):
I guess objects here a bit overkilling

Hlazman:
> I guess objects here a bit overkilling

Thanks. I got it

PR #1749, opened 2022-09-05
Solution

SergIats:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
alexandra-protyanova, line 32: move numbers to variables to everyone could understand what is it

alexandra-protyanova (changes requested):

alexandra-protyanova (changes requested):
I meant, move amount of discount to variable, and `3, 7`

alexandra-protyanova (approved):

PR #1750, opened 2022-09-06
Solution

anton-martyniuk1:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalPrice = days * 40;
+
+  if (days >= 7) {
+    totalPrice -= 50;
+  } else if (days >= 3) {
```
alexandra-protyanova, line 35: u can write this code without else construction

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1752, opened 2022-09-07
Solution

Irina-Kulish:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
+  const price = 40;
+
+  if (days <= 2) {
+    count = price * days;
+  } else if (days === 3) {
+    count = price * days - 20;
```
alexandra-protyanova, line 37: create more variables to everyone could understand where this numbers came from

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  } else if (days >= 7) {
+    count = price * days - 50;
+  } else if (days === 6) {
```
alexandra-protyanova, line 40: too many `else if`s u need maximally 2 of them, combine rules

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
```
marina-tilniak, line 31:
```suggestion
  let count = amount;
```

marina-tilniak: and now you can delete the last if

marina-tilniak (changes requested):

marina-tilniak (approved):

PR #1756, opened 2022-09-08
RentalCost

oleksii-bieliaiev:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else {
```
alexandra-protyanova, line 33: let us avoid nested else if structure

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+    if (days < 7 && days >= 3) {
+      return (days * 40) - 20;
```
alexandra-protyanova, line 35: move to variables to everyone could understand where these numbers came from and what is it

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
oleksii-bieliaiev: did as you said

alexandra-protyanova (approved):

PR #1761, opened 2022-09-09
feat: add function calculateRentalCost

ihorandrianov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,10 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = 40 * days;
+
+  return (days >= 3) ? (days >= 7)
```
korniienkokarina, line 33: looks good, but perhaps for the sake of readability it would be better not to use a ternary expression

Denchepornyuk (approved):

PR #1764, opened 2022-09-10
Solution

AniaHna:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
Denchepornyuk, line 37: 40, 50, and 20 are magic numbers
try to keep all numbers like these in variables for more readable code

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,12 +28,16 @@
  */
 
 function calculateRentalCost(days) {
+  const rent = 40;
+  const off3 = 20;
+  const off7 = 50;
```
Denchepornyuk, line 33: don't use numbers in variables name
off3 and off7 are not a good name for the variables

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1768, opened 2022-09-11
Solution

Andrew-Skoryk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
+  const cost = days * dayCost;
+  const basicDiscount = 20;
+  const additionalDiscount = 30;
+
+  switch (true) {
```
Andr1yK, line 36: I don't recommend to use switch (true), but up to u :)

Andr1yK (approved):

PR #1769, opened 2022-09-11
Solution

AntonVaida:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
+  } else if (days >= 3) {
+    return (days * 40) - 20;
+  } else {
+    return days * 40;
+  }
```
imaksweb, line 37: You should return the value immediately in if statements. And don't use else after return statement.
```suggestion
  if (days >= 7) {
    return (days * 40) - 50;
  } 
  
  if (days >= 3) {
    return (days * 40) - 20;
  }
  
  return days * 40;
```

imaksweb, line 37: You use `days * 40` in all cases, I think it's better to pre-calculate this value.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
```
imaksweb, line 32: Also it's better to add variable for rent price per day

imaksweb (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  let sum = days * pricePerDay;
+
+  if (days >= 7) {
+    sum -= 50;
+  } else if (days >= 3) {
+    sum -= 20;
+  }
```
imaksweb, line 38: You should return value right away after you've found the right condition. It makes the code simpler to read and helps to avoid unnecessary `else if`
```suggestion
  if (days >= 7) {
    return sum - 50;
  } 
  
  if (days >= 3) {
    return sum - 20;
  }
```

imaksweb (changes requested):

imaksweb (approved):
🔥

src/calculateRentalCost.js
Andr1yK: \+

Andr1yK: \+

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  const sum = days * pricePerDay;
+
+  if (days >= 7) {
+    return sum - 50;
```
Andr1yK, line 35: why 50?) and what is 50?

Andr1yK (changes requested):
Avoid magic nums

Andr1yK (approved):

PR #1770, opened 2022-09-11
solution

myroslav-diiak:


vladimir-dublya (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,13 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days >= 3 && days < 7) {
+    return (40 * days) - 20;
+  } else if (days >= 7) {
+    return (40 * days) - 50;
+  }
```
Denchepornyuk, line 36: 50, 40 and 20 are the magic numbers, create variables for them

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
myroslav-diiak: You don't like magic?!

Denchepornyuk: oh sure, code isn't place for magic

denys-cheporniuk (approved):

PR #1771, opened 2022-09-11
solution

Dikey945:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+
+  if (days >= 3 && days < 7) {
+    return days * 40 - threeDaysBonus;
+  }
+
+  if (days >= 7) {
+    return days * 40 - sevenDaysBonus;
+  } else {
+    return days * 40;
+  }
```
illia-sosnytskyi, line 43: you can implement this part without `else`

illia-sosnytskyi (changes requested):

src/calculateRentalCost.js
naziamoff: I agree with illia

naziamoff (changes requested):

naziamoff (approved):
Great 🚀

PR #1773, opened 2022-09-12
complete the task

barsuchellos:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days >= 7 ? days * 40 - 50 : days >= 3 ? days * 40 - 20 : days * 40;
```
Denchepornyuk, line 31: you should write clear and understandable code 
will be better to use ternary operator only in case when you need to store one of two values depending on some condition

in this task will be more understandable to use if / else statement
and for now 50, 40, 20 - magic numbers
create variables for them

each of this rules also are on checklist, don't forget to check it)

Denchepornyuk (changes requested):

src/calculateRentalCost.js
barsuchellos: ok , i will rewrite it)

Denchepornyuk (approved):

Andr1yK (approved):

PR #1774, opened 2022-09-12
Solution

ZoyaCanepa:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
-}
+  const dayRentCost = 40;
+  const rebateOfSevenDays = 50;
+  const rebateOfThreeDays = 20;
```
imaksweb, line 34: Don't hardcode some values that may changed in the future in variable names.

BAD EXAMPLE:
`const discountForThreeDays = 50;`

GOOD EXAMPLE:
`const basicDiscount = 50;`

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  const totalRentalCost = days * dayRentCost;
+
+  if (days >= 7) {
+    return totalRentalCost - rebateOfSevenDays;
+  } else if (days >= 3) {
+    return totalRentalCost - rebateOfThreeDays;
+  }
```
imaksweb, line 41: There's no need of `else if` statement
```suggestion
  if (days >= 7) {
    return totalRentalCost - rebateOfSevenDays;
  }
  
  if (days >= 3) {
    return totalRentalCost - rebateOfThreeDays;
  }
```

imaksweb (changes requested):

imaksweb (approved):
Well done 👍

Denchepornyuk (approved):

PR #1775, opened 2022-09-12
Solution

anton-iskryk:


artem-nasadchuk (approved):
gj

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const noDiscount = days * 40;
```
naziamoff, line 31: Good job! 🔥 
But it is better to rename noDiscount, it is not clear what it means now. 
I would make it fullPrice

naziamoff (approved):

PR #1777, opened 2022-09-12
Solution

rstarovoit:


Anna-Borodaienko (changes requested):
Add variables for dayCost and discounts

Andr1yK (changes requested):
[point 5](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const firstDiscount = 20;
+  const secondDiscount = 50;
+
+  if (days <= 2) {
+    return days * costPerDay;
+  }
+
+  if (days >= 3 && days <= 6) {
+    return days * costPerDay - firstDiscount;
+  }
```
silchyn, line 41:
```suggestion
  if (days <= 6) {
    return days * costPerDay - firstDiscount;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 7) {
+    return days * costPerDay - secondDiscount;
+  }
```
silchyn, line 45:
```suggestion
  return days * costPerDay - secondDiscount;
```

silchyn (changes requested):

src/calculateRentalCost.js
rstarovoit: Wow!)

silchyn (approved):
Cool!

PR #1779, opened 2022-09-12
calculateRentalCost

Vladosik1337:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
naziamoff, line 37: Good job! But we can make it a bit better

```suggestion
  if (days < 3) {
    return days * 40;
  }  
  
  if (days >= 3 && days < 7) {
    return days * 40 - 20;
  } 
  
  return days * 40 - 50;
 
```

naziamoff (changes requested):

naziamoff (changes requested):
Create variables for discounts. The code should be clear and understandable, you definitely don't want some random numbers in it

naziamoff (approved):
Gj 🚀

PR #1781, opened 2022-09-12
Solution

a-nahaitsev:


jojnajbude (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days < 3) {
+    totalCosts = carRentalPrice * days;
+  } else if (days < 7) {
+    totalCosts = carRentalPrice * days - discount1;
+  } else {
+    totalCosts = carRentalPrice * days - discount2;
+  }
+
+  return totalCosts;
```
naziamoff, line 44: You should not check for all conditions if you need only one of them to return a value

```suggestion
  if (days < 3) {
    return carRentalPrice * days;
  }  
  
  if (days < 7) {
   return carRentalPrice * days - discount1;
  } 

  return carRentalPrice * days - discount2;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCosts = 0;
+  const carRentalPrice = 40;
+  const discount1 = 20;
+  const discount2 = 50;
```
naziamoff, line 34: You should rename your variables. They have to describe what they are and why we need them
You may try sth like minDiscount and maxDiscount.

naziamoff (changes requested):

naziamoff (approved):
Nice! 🔥

PR #1783, opened 2022-09-12
Solution

jojnajbude:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,10 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = days * 40;
+  const getOff = days >= 3 ? (days >= 7 ? 50 : 20) : 0;
```
ArturNagaycev, line 32: It's better to write it with if else operator or write it in several lines

a-nahaitsev (approved):
Maybe for this task it's clear. But mentors don't advise to use this short form of condition operator

src/calculateRentalCost.js
Denchepornyuk, line 32: don't write ternary operator inside ternary operator

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const totalCost = days * 40;
+  let getOff = 0;
+
+  if (days >= 3) {
+    getOff = 20;
+  }
+
+  if (days >= 7) {
+    getOff = 50;
+  }
```
Denchepornyuk, line 40: use `else if` instead of two `if`

denys-cheporniuk (changes requested):
you didn't fix this one
https://github.com/mate-academy/js_task-transportation-on-vacation/pull/1783#discussion_r969631823

denys-cheporniuk (approved):

Anna-Tkachenko (approved):

PR #1784, opened 2022-09-12
Solution

imaksweb:


grafit-off (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const dayRent = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  let total = dayRent * days;
+
+  if (days >= 7) {
+    total -= bigDiscount;
+  } else if (days >= 3) {
+    total -= smallDiscount;
+  }
```
naziamoff, line 40: You should return value right away after you've found the right condition. It makes the code simpler to read and helps to avoid unnecessary `else if`

```suggestion
  if (days >= 7) {
    return total - bigDiscount;
  } 
  
  if (days >= 3) {
    return total - smallDiscount;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRent = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  let total = dayRent * days;
```
naziamoff, line 34: better to rename the variable. And if you'll return values as I've shown you below - you don't have to declare it with let, const will do perfect

```suggestion
  const fullPrice = dayRent * days;
```

naziamoff (changes requested):

naziamoff (approved):
🔥 🚀

PR #1785, opened 2022-09-12
solution

artem-nasadchuk:


anton-iskryk (changes requested):
Looks good. I'd recommend making some variables for such values as 'days * 40' and separately for 20 and 50 because they look like magical numbers.

anton-iskryk (approved):
GJ!

naziamoff (approved):
Gj!

PR #1786, opened 2022-09-12
Task was solved

denyskleimenov:


Yuriy-Komolov (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDiscount = 50;
+  const liteDiscount = 20;
+  const pricePerDay = 40;
```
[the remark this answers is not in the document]
naziamoff: @Yuriy-Komolov basicDiscount is better in order to program should be easy to maintain
If we have basicDiscount - its value can be easily changed in future. If we have sth like threeDaysDiscount - we will have to modify its name too.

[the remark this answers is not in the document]
naziamoff: but your point about the variable should be easy to understand - that is correct, I totally agree.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+  let totalPrice = 0;
+
+  for (let i = 0; i < days; i++) {
+    totalPrice += pricePerDay;
+  }
```
naziamoff, line 38: No need to use loop here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+
+  if (days >= 3 && days < 7) {
+    totalPrice -= liteDiscount;
+
+    return totalPrice;
+  } else if (days >= 7) {
+    totalPrice -= basicDiscount;
+
+    return totalPrice;
```
naziamoff, line 47: you should avoid using else when it is not necessary. You need to check only two `if's` in this task

naziamoff (changes requested):
You need but two if's in

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDiscount = 50;
+  const liteDiscount = 20;
+  const pricePerDay = 40;
+  let totalPrice = 0;
+
+  totalPrice += pricePerDay * days;
```
naziamoff, line 36: You may set expression as a value of the variable
```suggestion
  const totalPrice = pricePerDay * days;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+
+  if (days >= 3 && days < 7) {
+    totalPrice -= liteDiscount;
```
naziamoff, line 39: You should not change totalPrice in that case, you may return expression
```suggestion
    return totalPrice - liteDiscount;
```

naziamoff (changes requested):

naziamoff (approved):

PR #1789, opened 2022-09-12
Solution

OleksandrGlulhovskiy:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  if (days >= 7) {
+    return (sum += days * 40 - 50);
+  }
+
+  if (days >= 3) {
+    return (sum += days * 40 - 20);
+  }
+
+  if (days > 0) {
+    return (sum += days * 40);
```
Andrew-Skoryk, line 42: It might be a good idea not to use the same code in several places. Otherwise, everything is good 👍

Andrew-Skoryk (approved):
GJ!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  let sum = 0;
+
+  if (days >= 7) {
+    return (sum += days * 40 - 50);
+  }
+
+  if (days >= 3) {
+    return (sum += days * 40 - 20);
+  }
```
Denchepornyuk, line 39: 50, 40 and 20 are the magic numbers, create variables for them

src/calculateRentalCost.js
Denchepornyuk, line 42: I believe you don't need the last `if` you can remove it

src/calculateRentalCost.js
Denchepornyuk, line 39: use `else if` instead of two `if`

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const price = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  let sum = days * price;
+
+  if (days >= 7) {
+    return (sum -= bigDiscount);
+  } else if (days >= 3) {
+    return (sum -= smallDiscount);
+  }
```
Denchepornyuk, line 40: check your code with checklist
Screenshot 2022-09-15 at 12 48 18

denys-cheporniuk (approved):

Andr1yK (approved):

PR #1790, opened 2022-09-12
Solution

roman-zherebetskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * 40;
+
+  if (days > 2 && days < 7) {
+    return result - 20;
```
Andr1yK, line 34: what is 20, 50... or 40? I don't know🥲

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * 40;
+
+  if (days > 2 && days < 7) {
```
Andr1yK, line 33: u can check >=7 days and avoid &&

Andr1yK (changes requested):

src/calculateRentalCost.js
roman-zherebetskyi: okay I'll create a variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+  if (days < 7) {
+    return result - lowDiscount;
+  }
+
+  if (days >= 7) {
+    return result - bigDiscount;
+  }
+
+  return result;
```
Andr1yK, line 49: u really need 4 returns? Mb, we can decrease their count?)

Andr1yK (changes requested):

src/calculateRentalCost.js
roman-zherebetskyi: Andreeeeeeeew!!!!

Andr1yK: oops😂

Andr1yK (approved):
GJ

PR #1791, opened 2022-09-12
Solution

andrew-satyr:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return 40 * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    return 40 * days - 20;
+  }
+
+  if (days >= 7) {
+    return 40 * days - 50;
+  }
```
OleksandrGlulhovskiy, line 41: use else if, if you have more then one if.

OleksandrGlulhovskiy (commented):
nice job))

varkovroma (approved):

src/calculateRentalCost.js
Andr1yK: Is it really needed if u immediately return the value? As for me - no

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  } else if (days >= 3 && days < 7) {
+    return 40 * days - 20;
+  } else if (days >= 7) {
+    return 40 * days - 50;
+  }
```
Andr1yK, line 37: If u immediately return the value, u can write simple `if`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
```
Andr1yK, line 32: Duplicated in 3 places, mb, move to const

```suggestion
    return 40 * days;
```

Andr1yK (changes requested):
20, 40, 50 - some magic number 🥲
move to const

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const dayRent = 40;
+  const rentWithoutBonus = dayRent * days;
+
+  if (days < 3) {
+    return rentWithoutBonus;
+  } else if (days >= 3 && days < 7) {
+    return rentWithoutBonus - littleBonus;
+  } else if (days >= 7) {
+    return rentWithoutBonus - bigBonus;
+  }
```
silchyn, line 42:
```suggestion
  if (days < 3) {
    return rentWithoutBonus;
  }
  
  if (days < 7) {
    return rentWithoutBonus - littleBonus;
  }
  
  return rentWithoutBonus - bigBonus;
```

silchyn (changes requested):
Don't ignore comments

silchyn (approved):
Cool!

PR #1792, opened 2022-09-12
Solution

SanyaKovalenko:


bugsworld38 (approved):
nothing to suggest
gj

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  let cashBack = 0;
```
naziamoff, line 32: I would better divide this one into two different discount variables (declared with const)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+  if (days > 0 && days < 3) {
+    return rent * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    cashBack = 20;
```
naziamoff, line 39: if you create discount variables at the very beginning you won't have to set different value for cashback here, just use the existing variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+    return rent * days - cashBack;
+  }
+
+  if (days >= 7) {
+    cashBack = 50;
+
+    return rent * days - cashBack;
+  }
```
naziamoff, line 48: You don't need this if statement, because previous statements cover every other scenario.
If you divide cashback into discounts - you can just return it like that

```suggestion
    return rent * days - maxDiscount;
```

naziamoff (changes requested):
Gj, added some tips to improve it

naziamoff (changes requested):
Try to remove those, it should work as expected
The point is - we return a value if days < 3. So function stops right there anyway and we don't have to check whether it is >= 3

naziamoff (approved):
GJ!

PR #1793, opened 2022-09-12
Solution

anastasiia-nahaiska:


avkovlit (approved):
GJ!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const threeDaysDiscount = 20;
+  const sevenDaysDiscount = 50;
+  const dayPrice = 40;
+  let cost = 0;
+
+  if (days >= 7) {
+    cost = dayPrice * days - sevenDaysDiscount;
+  }
```
naziamoff, line 38: you should make a return statement if you don't need further checks
So if days >= 7 you don't have to check whether it is < 7 or is < 3

naziamoff (changes requested):

src/calculateRentalCost.js
naziamoff: it is applicable for all of your if's btw. If you figure out how to do it - you won't even have to use 'cost' variable

naziamoff (approved):
GJ 🔥

PR #1795, opened 2022-09-12
Solution

avkovlit:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const basicDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days > 2 && days < 7) {
+    total = total - basicDiscount;
+  }
+
+  if (days >= 7) {
+    total = total - maxDiscount;
+  }
```
imaksweb, line 42: You should return value right after you've found the right condition. Also there's no need to use such a complex condition like `days > 2 && days < 7`
```suggestion
  if (days >= 7) {
    return total - maxDiscount;
  }

  if (days >= 3) {
    return total - basicDiscount;
  }
  
  return total;
```

imaksweb (changes requested):

xiimiik (approved):
All right 👍

src/calculateRentalCost.js
naziamoff: I agree with Illia 👍

avkovlit: done!

naziamoff (approved):
Gj!

PR #1801, opened 2022-09-12
solution

poznianski:


tatyana22 (approved):
Good!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const sevenDaysOff = 50;
+  const threeDaysOff = 20;
+
+  if (days >= 7) {
+    totalCost = baseRate * days - sevenDaysOff;
+  } else if (days >= 3) {
+    totalCost = baseRate * days - threeDaysOff;
+  } else if (days < 3) {
+    totalCost = baseRate * days;
+  }
```
naziamoff, line 42: try to do it without `else`, it is unnecessary here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
-  // write code here
+  let totalCost = 0;
+  const baseRate = 40;
+  const sevenDaysOff = 50;
+  const threeDaysOff = 20;
+
+  if (days >= 7) {
+    totalCost = baseRate * days - sevenDaysOff;
+  } else if (days >= 3) {
+    totalCost = baseRate * days - threeDaysOff;
```
naziamoff, line 39: Feel free to return right away if you've found the proper condition
```suggestion
    return baseRate * days - threeDaysOff;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = 0;
+  const baseRate = 40;
+  const sevenDaysOff = 50;
+  const threeDaysOff = 20;
```
naziamoff, line 34: you should take a look at the checklist, there is a point about how you should name your variables

naziamoff (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const baseRate = 40;
+  const sevenDaysDiscount = 50;
+  const threeDaysDiscount = 20;
+
+  if (days >= 7) {
+    return baseRate * days - sevenDaysDiscount;
```
Anna-Tkachenko, line 36: don't repeat yourself: create separate variable for baseRate * days value

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

src/calculateRentalCost.js
tatyana22: I see. ok!

src/calculateRentalCost.js
tatyana22: ok

PR #1805, opened 2022-09-12
Solution

korniienkokarina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
```
imaksweb, line 33: Don't hardcode some values that may change in the future in variable names (numbers etc.)
BAD EXAMPLE:
```js
```
GOOD EXAMPLE:
```js
const basicDiscount = 50;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  let sum = days * costPerDay;
+
+  if (days >= 7) {
+    sum -= discountSevenDays;
+  } else if (days >= 3) {
+    sum -= discountThreeDays;
+  }
```
imaksweb, line 41: You should return value right after you've found the right condition. It makes the code simpler to read and helps to avoid unnecessary `else if`. And using this approach you can use a `sum` as a `const`, cause there's no need to change it anymore.
```suggestion
  if (days >= 7) {
    return sum - discountSevenDays;
  }
  
  if (days >= 3) {
    return sum - discountThreeDays;
  }
```

imaksweb (changes requested):

src/calculateRentalCost.js
korniienkokarina: thank you, I actually forgot about checklist🙈

imaksweb (approved):
Great job 🔥

naziamoff (approved):
Good job 🔥

PR #1808, opened 2022-09-12
solution

igor-1987:
initial commit for task

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -32,13 +32,13 @@ function calculateRentalCost(days) {
   const longTimeRent = 7;
   const middleTimeRent = 3;
   const baseRentPrice = 40;
-  const longTimeCiscount = 50;
-  const threeDaysCiscount = 20;
+  const longTimeDiscount = 50;
+  const threeDaysDiscount = 20;
 
   if (days >= longTimeRent) {
-    total = days * baseRentPrice - longTimeCiscount;
+    total = days * baseRentPrice - longTimeDiscount;
   } else if (days >= middleTimeRent) {
-    total = days * baseRentPrice - threeDaysCiscount;
+    total = days * baseRentPrice - threeDaysDiscount;
   } else {
     total = days * baseRentPrice;
   }
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total;
+  const rentPrice = 40;
+
```
Yeliseiev, line 33: 4. [REUSABILITY] - Give variables abstract names - it's will be easiar for another person to read your code
5. [VARIABLES] - Don't hardcode some values that may change in the future in variable names (numbers etc.)

Name the variables of discount. Look examples in checklist.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  if (days >= 7) {
+    total = days * rentPrice - 50;
+  } else if (days >= 3 && days <= 6) {
```
Yeliseiev, line 36: You don't need check if days less then 6.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+    total = days * rentPrice - 20;
+  } else {
+    total = days * rentPrice;
```
Yeliseiev, line 39: Maybe count total outside the statement 'if'. To remove the unnecessary 'else'.

Yeliseiev (changes requested):

src/calculateRentalCost.js
igor-1987: Yeah, all of it make sense, I'll fix these issues. Thank you for remarks!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total;
+  const longTimeRent = 7;
+  const middleTimeRent = 3;
+  const baseRentPrice = 40;
+  const longTimeCiscount = 50;
+  const threeDaysCiscount = 20;
```
Denchepornyuk, line 36:
```suggestion
  const longTimeDiscount = 50;
  const threeDaysDiscount = 20;
```

denys-cheporniuk (approved):

src/calculateRentalCost.js
igor-1987: Yeah, looks and sounds weird.

PR #1811, opened 2022-09-12
Solution

vallerii:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return count;
+  }
+
+  if (days < 7) {
+    return count - firstDiscount;
+  }
+
+  if (days >= 7) {
+    return count - secondDiscount;
+  }
```
imaksweb, line 47: For better readability of your code you can use only 2 `if` statements here. There's no need to use `if (days < 3)` statement:
```js
  if (days >= 7) {
    return count - secondDiscount;
  }

  if (days >= 3) {
    return count - firstDiscount;
  }

return count;
```

imaksweb (changes requested):
GJ 🔥 Just one minor thing you need to do if you agree, but it's up to you)

src/calculateRentalCost.js
```diff
@@ -34,17 +34,15 @@ function calculateRentalCost(days) {
 
   const count = days * price;
 
-  if (days < 3) {
-    return count;
-  }
-
-  if (days < 7) {
+  if (days >= 3 && days < 7) {
```
imaksweb: There's no need to use such a complex condition. You can check first `if (days >= 7)` and after that you can use `if (days >= 3)` condition because all numbers greater than `7` don't pass through first condition

imaksweb (changes requested):

imaksweb (approved):
Great job 👍

src/calculateRentalCost.js
vallerii: i have understand this after push first fixing commit, so check the second fixed commit, please

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const firstDiscount = 20;
+  const secondDiscount = 50;
```
naziamoff, line 33: those variables can be more clear
Would be better to make it like smallDiscount/bigDiscount to make variable names more understandable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  const count = days * price;
```
naziamoff, line 35: `Count` should be renamed too. Let's say
```suggestion
  const fullPrice = days * price;
```

naziamoff (changes requested):
Well done 🔥

naziamoff (approved):
Nice!

PR #1812, opened 2022-09-12
Solution

La-La-Op:


Anna-Borodaienko (approved):
You can add one more variable: days * rentPerDay

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPerDay = 40;
+  const moreThenSevenDiscount = 50;
+  const moreThenThreeDiscount = 20;
+
+  if (days < 3) {
```
OleksiiBlezniuk, line 35: Better to store such numbers in variables, now it looks like magic numbers

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return days * rentPerDay;
+  }
+
+  if (days >= 7) {
+    return days * rentPerDay - moreThenSevenDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * rentPerDay - moreThenThreeDiscount;
+  }
```
OleksiiBlezniuk, line 45: You've duplicated logic of previous `if's` in this one

No need to create it, your `days` value is already fits this condition (`days` are already more or equal than 3 after first `if` and less than 7 after second)

OleksiiBlezniuk (approved):

OleksiiBlezniuk (approved):

PR #1813, opened 2022-09-12
add variables

volodymyrklymkiv:


bugsworld38 (approved):

naziamoff (changes requested):
Try to remove variable `count`
Hint: you don't have to check all the possible conditions each time. In that case you can return value right after you've found the needed condition

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+    return days * rentPrice;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * rentPrice) - smallDiscount;
+  }
+
+  if (days >= 7) {
+    return (days * rentPrice) - bigDiscount;
+  }
```
naziamoff, line 46: You don't need if statement here.
Both previous options cover every possible different path, so feel free to return without `if` after them

naziamoff (changes requested):

naziamoff (approved):
Good job 🚀

PR #1814, opened 2022-09-12
Solution

oleksandr-korobko:


imaksweb (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const costOfDayRent = 40;
+  let totalPrice = costOfDayRent * days;
+
+  switch (true) {
+    case days >= 7:
+      totalPrice -= 50;
+      break;
+
+    case days >= 3:
+      totalPrice -= 20;
```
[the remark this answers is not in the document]
oleksandr-korobko: Maybe you're right)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  }
```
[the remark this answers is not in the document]
oleksandr-korobko: Especially for me, I think that when we calculate 1 logical process it is better to use "switch case" instead "if"

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costOfDayRent = 40;
+  let totalPrice = costOfDayRent * days;
+  const threeDaysDiscaunt = 20;
+  const sevenDaysDiccaunt = 50;
```
naziamoff, line 34: check 5th point in the cheklist

naziamoff (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const rentCost = 40;
+  const minDiscaunt = 20;
+  const maxDiscaunt = 50;
+
+  let totalPrice = rentCost * days;
+
+  switch (true) {
+    case days >= 7:
+      totalPrice -= maxDiscaunt;
+      break;
```
naziamoff, line 40:
```suggestion
      return totalPrice - maxDiscount;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+    case days >= 3:
+      totalPrice -= minDiscaunt;
+      break;
+  }
```
naziamoff, line 45:
```suggestion
    case days >= 3:
      return totalPrice - minDiscount;
  }
```

naziamoff (approved):

src/calculateRentalCost.js
naziamoff: In that case you should return right away after you've found your condition

PR #1815, opened 2022-09-12
Develop

Ihor-1993:


NickBrodyuk:
GJ

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+
+  if (days < 3) {
+    return 40 * days;
+  }
+
+  if (days >= 3 && days < 7) {
```
Denchepornyuk, line 36: you checked `days < 3` in the first `if` so in the second `if` you don't need `days >= 3` here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+    return 40 * days - 20;
+  }
+
+  if (days >= 7) {
+    return 40 * days - 50;
+  }
```
Denchepornyuk, line 42: you can remove if here and leave only return

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  }
+
+  if (days < 7) {
+    return 40 * days - 20;
+  }
+
+  return 40 * days - 50;
```
Denchepornyuk, line 39: 50, 40 and 20 are the magic numbers, create variables for them

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1817, opened 2022-09-12
Car rental

Yuriy-Komolov:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let totalAmount = 0;
+  const rentPerDay = 40;
+  const sevenDaysDiscount = 50;
+  const threeDaysDiscount = 20;
```
DenysKleimenov, line 35: Don't hardcode some values that may change in the future in variable names (numbers etc.)
Something like basicDiscount and so on would be good.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+
+  for (let i = 0; i < days; i++) {
+    totalAmount += rentPerDay;
+  }
+
+  if (days >= 3 && days < 7) {
+    totalAmount -= threeDaysDiscount;
+  } else if (days >= 7) {
+    totalAmount -= sevenDaysDiscount;
+  }
```
DenysKleimenov, line 45: You should return the value immediately in if statements.

```suggestion
  if (days >= 3 && days < 7) {
    totalAmount -= threeDaysDiscount;
    
    return totalAmount;
  } else if (days >= 7) {
    totalAmount -= sevenDaysDiscount;
    
    return totalAmount;
  }
```

denyskleimenov (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,26 @@
...
 function calculateRentalCost(days) {
   // write code here
+  let totalAmount = 0;
+  const rentPerDay = 40;
+  const sevenDaysDiscount = 50;
+  const threeDaysDiscount = 20;
+
+  for (let i = 0; i < days; i++) {
+    totalAmount += rentPerDay;
+  }
```
naziamoff, line 39: You don't need loop in this task

src/calculateRentalCost.js
naziamoff: I agree with Denys, you should take a look at the checklist again

src/calculateRentalCost.js
```diff
@@ -29,6 +29,26 @@
...
+
+  if (days >= 3 && days < 7) {
+    totalAmount -= threeDaysDiscount;
```
naziamoff, line 42: figure out how to return right after the program has found proper condition

naziamoff (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const rentPerDay = 40;
+  const totalAmount = days * rentPerDay;
+  const basicDiscount = 50;
+  const minorDiscount = 20;
+
+  if (days >= 3 && days < 7) {
+    return totalAmount - minorDiscount;
+  } else if (days >= 7) {
```
naziamoff, line 39: remove `else` here

naziamoff (changes requested):
Not bad at all, but fix comment

naziamoff (approved):

PR #1821, opened 2022-09-13
Solution

oleksusov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
+
+  if (days >= 7) {
+    total -= 50;
+  } else if (days >= 3) {
+    total -= 20;
+  }
```
ArturErmolenko, line 37: [CODE STYLE] - You should return the value immediately in if statements. And don't use else after return statement. If return is performed - code after it won't be executed anyway

ArturErmolenko (changes requested):
VARIABLES] - use variables for the main values so that you can reuse them and give them descriptive names.

[Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

valerii-derkach (approved):

PR #1823, opened 2022-09-13
Solution

hajnaloltyan:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRate = 40 * days;
```
Bibicutu, line 31: You should declare a variable for price per day, according to checklist item 5.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const twentyOff = 20;
+  const fiftyOff = 50;
```
Bibicutu, line 33: Change the names for this two variables - checklist item 5.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  switch (days) {
```
Bibicutu, line 35: According to checklist item 3, you should use `if` instead of `while`.

Bibicutu (changes requested):
Make the changes below and send to mentor review. Good job!

src/calculateRentalCost.js
hajnaloltyan: I mean you're looking at it, I don't understand what the issue is?

src/calculateRentalCost.js
hajnaloltyan: These are in the tasks description....

src/calculateRentalCost.js
hajnaloltyan: [CODE STYLE] - You should return the value immediately in if statements. And don't use else after return statement. If return is performed - code after it won't be executed anyway.    

I'm sorry but I don't see that part. What do you mean?

src/calculateRentalCost.js
Bibicutu: I mean don't use value 40, use a variable instead.

src/calculateRentalCost.js
Bibicutu: `twentyOff`:

> Don't hardcode some values that may change in the future in variable names (numbers etc.)

the value twenty may change.

VinnieJ-2k20 (changes requested):
Good job with variables.
Need to rewrite using `if`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const smallDiscountDays = 3;
+  const bigDiscountDays = 7;
+
+  if (days < smallDiscountDays) {
+    return dailyRate;
+  } else if (days >= smallDiscountDays && days < bigDiscountDays) {
+    return dailyRate - smallDiscount;
+  } else {
+    return dailyRate - bigDiscount;
+  }
```
VinnieJ-2k20, line 44: You can simplify this complex structure

1. Get rid of `else`. If there's `return` inside `if` block code would stop executing anyway on return
2. Avoid using double conditions. Choose direction 0...7 days or 7...0 days and follow it.

Check the checklist and code examples there once again.

VinnieJ-2k20 (changes requested):
Good job with variables.
But conditions can be simplified.

VinnieJ-2k20 (approved):

PR #1824, opened 2022-09-13
solution

bradleymatjie:


MeshackDev (changes requested):
You calculate cost * days three times. Try using a variable.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
   // write code here
+  const cost = 40 * days;
+  let amount = 0;
+
+  if (days < 3) {
+    amount = cost;
+  }
+
+  if (days >= 3) {
+    amount = cost - 20;
```
VinnieJ-2k20, line 40: Introduce variables  for discounts.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const cost = 40 * days;
+  let amount = 0;
+
+  if (days < 3) {
+    amount = cost;
+  }
```
VinnieJ-2k20, line 37: Find a way to return value from inside if block instead of using `let`.

VinnieJ-2k20 (changes requested):
Some changes needed.

ihor-jpeg (approved):

ihor-jpeg (changes requested):
test

VinnieJ-2k20 (approved):

PR #1826, opened 2022-09-14
Solution

MAKSOti:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
```
rstarovoit, line 30: Add variables for CostPerDay and DiscountsFrom...

rstarovoit (changes requested):

naziamoff (changes requested):
This would look clearer with `if`. And you should create separate variables for discounts. There shouldn't be any magic numbers in your code

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const rentPrice = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return days * rentPrice;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * rentPrice) - smallDiscount;
```
Anna-Tkachenko, line 40: don't repeat yourself: create separate variable for days * rentPrice value

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1829, opened 2022-09-14
solution for calculate rental

sapnachoudhary06:


BZOzyurt (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const dayTariff = 40;
+  const total = days * dayTariff;
+
+  if (days >= 3 && days < 7) {
+    return total - 20;
+  } else if (days >= 7) {
+    return total - 50;
+  }
```
VinnieJ-2k20, line 39: 1. Read checklist item 3 about `if/else` and `return`
2. Introduce variables for discount values.

VinnieJ-2k20 (changes requested):

src/calculateRentalCost.js
sapnachoudhary06: @VinnieJ-2k20 , Addressed your comments. Please check again.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+  const discount50 = 50;
+  const totalTariff = days * dayTariff;
+
+  if (days >= 3 && days < 7) {
+    return totalTariff - discount20;
+  }
+
+  if (days >= 7) {
+    return totalTariff - discount50;
+  }
```
VinnieJ-2k20, line 43: You can simplify this.
Avoid double conditions.
Switch these blocks and use single condition inside if statement.

VinnieJ-2k20 (changes requested):

src/calculateRentalCost.js
sapnachoudhary06: @VinnieJ-2k20  Simplified as per your comment.

VinnieJ-2k20 (approved):

PR #1833, opened 2022-09-14
solution

metinbicaksiz:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
...
-}
+  let total = 0;
+
+  if (days >= 7) {
+    total = (days * 40) - 50;
+  } else if (days > 2 && days < 7) {
+    total = (days * 40) - 20;
+  } else {
+    total = days * 40;
+  }
```
amitpatiljc, line 40: let's not have magic numbers in the code. at the moment i dont understand what 20, 40 and 50 are
let's use variables

amitpatiljc (changes requested):

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job !

PR #1835, opened 2022-09-15
Solution

bokorzita:


ghost (changes requested):
Declare and use variables for repetitive syntaxes, also, rewrite it to avoid `else` as per the checklist

VinnieJ-2k20 (changes requested):
Check the checklist please and follow advice mentioned there.

amitpatiljc (approved):
Nicely Done!

PR #1840, opened 2022-09-15
commit

OleksChernikov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const result = (days * 40);
+
+  if (days > 6) {
+    return result - 50;
+  }
+
+  if (days > 2) {
+    return result - 20;
```
SanyaBratashchuk, line 38: Move all 'magic' numbers to the variables

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = (days * 40);
+  const magicNumOne = 2;
+  const magicNumTwo = 6;
```
Bohdan-Maksymiuk, line 33: 😄

Bohdan-Maksymiuk: use more meaningful names
for example - daysForSmallDiscount
and pass amount of discount to variable too

Bohdan-Maksymiuk (changes requested):

AlisaVlasova (approved):

PR #1841, opened 2022-09-16
car rent

kalmykov-mykhailo:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const amountNotDiscount = days * 40;
+
+  switch (true) {
```
alexandra-protyanova, line 33: u don't need to use switch, if its condition true

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+    case (days >= 7):
+      return amountNotDiscount - 50;
+
+    case (days >= 3):
+      return amountNotDiscount - 20;
```
alexandra-protyanova, line 38: move numbers to variables, to everyone could understand what is it

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1843, opened 2022-09-17
commit1

Gintako666:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else {
+    if (days >= 3 && days < 7) {
+      return days * 40 - 20;
+    } else {
+      return days * 40 - 50;
+    }
+  }
```
SanyaBratashchuk, line 39: move all 'magic' numbers to the variables and try to get rid of `else `

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const money = days * 40;
+
+  switch (true) {
```
alexandra-protyanova, line 33: u don't need swith, if its condition true

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const money = days * 40;
+
+  if (days < 3) {
+    return money;
+  }
+
+  if (days >= 3 && days < 7) {
+    return money - 20;
```
alexandra-protyanova, line 38: move this numbers to variables, to everyone be able to understand where they came from

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1848, opened 2022-09-21
Solution

mykolahuryn:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let result;
+
+  if (days < 3) {
+    result = days * 40;
+  } else if (days >= 3 & days < 7) {
+    result = (days * 40) - 20;
+  } else if (days >= 7) {
+    result = (days * 40) - 50;
+  }
```

Bohdan-Maksymiuk: avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  const cashBack3MoreDays = 20;
+  const cashBack7MoreDays = 50;
+
+  if (days < 3) {
+    return rentCost;
+  }
+
+  if (days >= 3 & days < 7) {
+    result = rentCost - cashBack3MoreDays;
+  }
```
Bohdan-Maksymiuk, line 42:
```suggestion
  if (days < 7) {
    result = rentCost - cashBack3MoreDays;
  }
```
you returned value from function in previous if, so no need to check it here

Bohdan-Maksymiuk (changes requested):

AlisaVlasova (approved):

PR #1851, opened 2022-09-24
Solution

bagibest:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
+  switch (true) {
+    case (days >= 7):
+      return sum - 50;
+    case (days >= 3):
+      return sum - 20;
+    default:
+      return sum;
+  }
+}
```
SanyaBratashchuk, line 41: switch true is useless, use if and move all 'magic' numbers to the variables

SanyaBratashchuk (changes requested):

stanislav-topikha (changes requested):
You misunderstood previous comment
Move all number to the variables with corresponding names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+  const sum = pricePerDay * days;
+
+  if (days >= 7) {
+    return sum - maxDiscount;
+  } else if (days >= 3) {
```
stanislav-topikha, line 38:
```suggestion
  if (days >= bigPeriod) {
    return sum - maxDiscount;
  } else if (days >= smallPeriod) {
```
2 more vars

stanislav-topikha (changes requested):

SanyaBratashchuk (approved):

PR #1855, opened 2022-09-26
solution

Prashanth-svg:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const d = 40;
```
amitpatiljc, line 31: use proper variable names like `costPerDay`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days >= 1 && days <= 2) {
+    return d * days;
+  } else if (days >= 3 && days <= 6) {
+    return (d * days) - 20;
+  } else if (days >= 7) {
+    return (d * days) - 50;
+  }
 }
```
amitpatiljc, line 40: also let's not have magic numbers in the code. at the moment i dont understand what 20, 40 and 50 are

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const d = 40;
+
+  if (days >= 1 && days <= 2) {
+    return d * days;
+  } else if (days >= 3 && days <= 6) {
+    return (d * days) - 20;
+  } else if (days >= 7) {
+    return (d * days) - 50;
+  }
```
amitpatiljc, line 39: either use if else and return directly only at the end or
only use 2 if statements and return after every if

amitpatiljc (changes requested):

PR #1858, opened 2022-09-28
Solution

Artem193:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return days * 40 - 20;
+  }
+
+  return days * 40 - 50;
```
SanyaBratashchuk, line 43: move all 'magic' numebrs to the variables

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
Artem193: fixed

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days === 0) {
+    return 0;
+  }
+
+  if (days < 3) {
+    return days * priceRental;
+  }
+
+  if (days < 7) {
```
Bohdan-Maksymiuk, line 43: move dayNumbers into variables too

Bohdan-Maksymiuk (changes requested):
Good job - one more comment

src/calculateRentalCost.js
Artem193: fixed

polosanya (approved):
Great job!

PR #1860, opened 2022-09-28
Task was done

Vlad-Fedorishchev:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  let sum = 0;
+  const costPerDay = 40;
+
+  if (days >= 7) {
+    sum = (costPerDay * days) - 50;
+  } else if (days >= 3) {
+    sum = (costPerDay * days) - 20;
+  } else {
+    sum = costPerDay * days;
+  }
```
SanyaBratashchuk, line 40: move all 'magic' numbers to the variables and get rid of `else if` and `else`

SanyaBratashchuk (changes requested):

valerii-derkach (changes requested):

liliia-livandovskaya (approved):
Good job 🔥

PR #1864, opened 2022-10-01
Solution calculateRentalCost

Sasha-Green:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days < 3) {
+    return amount;
+  }
+
+  if (days < 7) {
+    return amount - 20;
+  }
+
+  return amount - 50;
```
SanyaBratashchuk, line 42: move all 'magic' numebers to the variables

SanyaBratashchuk (changes requested):

liliia-livandovskaya (approved):
Great job 🔥

PR #1868, opened 2022-10-04
solution

andriibilous4:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const price = 40;
+  const total = days * price;
+
+  if (days >= 7) {
+    return total - 50;
+  }
+
+  if (days >= 3) {
+    return total - 20;
+  }
```
Bohdan-Maksymiuk, line 40: what a magic numbers 🪄 (days and amountOfDiscount)

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (changes requested):
new test reject

PR #1869, opened 2022-10-04
Solution

natalia-maryshko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    amount = rent;
+  }
+
+  if (days >= 3 && days < 7) {
+    amount = rent - 20;
+  }
+
+  if (days >= 7) {
+    amount = rent - 50;
+  }
```
Bohdan-Maksymiuk, line 44: what a magic numbers 🪄

Bohdan-Maksymiuk (changes requested):

liliia-livandovskaya (approved):
👍

PR #1870, opened 2022-10-04
solution

mykola8426:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
```
ArtemTopolskyi, line 36: what do all this magic numbers mean? Move all numbers to separate variables

ArtemTopolskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discountAfterTwoDays = 20;
+  const discountAfterSixDays = 50;
```
ArtemTopolskyi, line 32:
```suggestion
  const smallDiscount = 20;
  const largeDiscount = 50;
```
why after two or six? what if we change our business requirements and there will be 3 days and 5 days? Then we should update it everywhere in code, it's not ok

ArtemTopolskyi (approved):

PR #1872, opened 2022-10-05
Solution

bonchuk:


src/calculateRentalCost.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+// const { Linter } = require('eslint');
+
```
TarasHoliuk, line 4: delete it, please

TarasHoliuk (approved):
Approved with 1 warning: don't leave comments

PR #1883, opened 2022-10-13
task was done

Sergey-Mironenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let amount = 0;
+
+  if (days < 3) {
+    amount = days * 40;
+  } else if ((days >= 3) && (days < 7)) {
+    amount = (days * 40) - 20;
+  } else {
+    amount = (days * 40) - 50;
```
TarasHoliuk, line 38: To make the code more clear and readable better to create variables for all magic (for now) numbers, like:
`const discount = 20` etc.

TarasHoliuk (changes requested):

loralevitska (approved):
good job)

PR #1884, opened 2022-10-13
Jtransportation-on-vacation-service

divertitoS:


Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const rentCostPerDay = 40;
+  const threeDaysDiscount = 20;
+  const weeklyDiscount = 50;
+  const total = days * rentCostPerDay;
+
+  if (days >= 7) {
+    return total - weeklyDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
```
Bohdan-Maksymiuk, line 40: lets move day numbers into variables too
if they change, we should find every place in code and change there

Bohdan-Maksymiuk, line 40:
```suggestion
  if (days >= 7) {
    return total - weeklyDiscount;
  }

  if (days >= 3) {
```
you have returned from function in previous if for this condition

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCostPerDay = 40;
+  const threeDaysDiscount = 20;
+  const weeklyDiscount = 50;
```
Bohdan-Maksymiuk, line 33: do not use numbers in names - what if we change 3 to 4?
You will change the variables names then?

Bohdan-Maksymiuk (changes requested):

PR #1887, opened 2022-10-16
Solution

FallenMAD:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const sum = 40;
```
polosanya, line 31: What is sum? I think it's better to give it more informative name

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  let daysCost = sum * days;
```
polosanya, line 33: daysCost is also not the best name. Customer pays not just for days, but for some service. Something like rentalCost might sound better

polosanya (changes requested):
Good job! Just please pay attention to the naming style of variables. Another developer should clearly understand what is going on in your code)

polosanya (approved):
Well done!

PR #1888, opened 2022-10-17
firstJS_Git

Kukharuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return amount;
+  }
+
+  if (days < 7) {
+    amount = (days * 40) - 20;
+
+    return amount;
+  }
+
+  return (days * 40) - 50;
```
Bohdan-Maksymiuk, line 45: what a magic numbers 🪄

Bohdan-Maksymiuk (changes requested):
Look throughout [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) again, pls

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days < 3) {
+    return baseAmount;
+  }
+
+  if (days < 7) {
+    return baseAmount - 20;
+  }
+
+  return baseAmount - 50;
```
Bohdan-Maksymiuk, line 42: Let's move numberOfDaysToSmallDiscount, same as for large, and amount of discounts

Bohdan-Maksymiuk (changes requested):
Lets fix one more comment

polosanya (approved):
Well done!

PR #1895, opened 2022-10-17
Solution

andrudenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
```
alexandra-protyanova, line 31: move calculations to everyone could understand where did they came from

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return (days * 40 - 20);
+  }
+
+  if (days >= 7) {
```
alexandra-protyanova, line 39: on of case can be without if

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1896, opened 2022-10-18
commit

AndreyChernysh:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  switch (true) {
+    case (days >= 3 && days < 7):
+      return days * 40 - 20;
```
Largoxoxo, line 35:
```suggestion
      return days * carRent - discountForThreeDays;
```
Checklist #1, you should use variables and give them descriptive names.

vadymbaranov (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  switch (true) {
```
Olexandr6, line 33: Don't use `switch` with `true`. Better to use `if` clauses instead

src/calculateRentalCost.js
Olexandr6: +1, let's fix that

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+    case (days >= 3 && days < 7):
+      return days * 40 - 20;
+
+    case (days >= 7):
+      return days * 40 - 50;
+
+    default:
+      return days * 40;
```
Olexandr6, line 41: `days * 40` - you use this expression in the several places, so makes sense to create a variable for that

G-YaSha (changes requested):
Pls, read the [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

G-YaSha (approved):

PR #1899, opened 2022-10-18
Solution

Anatolii-Voloshyn:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 6) {
+    return (days * 40) - 50;
```
Anastasiia-Smirnova, line 32: Good job! I have almost the same solution:)
Although you don't really need brackets, because multiplication will be performed before subtraction even without them.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  };
+
+  if (days > 2) {
+    return (days * 40) - 20;
```
Anastasiia-Smirnova, line 36: Also you should add a variable with the rent cost per day (40), because you use it multiple times and it will be easier to change it if needed.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 6) {
+    return (days * 40) - 50;
+  };
```
Anastasiia-Smirnova, line 33: And you shouldn't put semicolon after curly brackets {}

Anastasiia-Smirnova (approved):

mkyloa (approved):

PR #1900, opened 2022-10-18
solution

mitattt:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
```
Olexandr6, line 31: `price` of what?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return days * price;
+  }
+
+  if (days <= 6) {
+    return days * price - littleDiscount;
+  }
+
+  if (days >= 7) {
+    return days * price - bigDiscount;
+  }
```
Olexandr6, line 45: Let's think about how we can implement this logic using only 2 `if` conditions0

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const littleDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days <= 2) {
+    return days * price;
```
Olexandr6, line 36: `days * price` - you use this expression in several places, so makes sense to move it to a variable

Olexandr6 (changes requested):
Your solution is good, but not great🙂 Let's work on improvements👇

Olexandr6 (approved):

PR #1902, opened 2022-10-18
Solution

iryna-lelyk:


Illia-Skladnik (approved):
Good job and interesting solution. I've made it in another way)

mkyloa (changes requested):
checklist 2,3

mkyloa (approved):

PR #1904, opened 2022-10-18
complete

spiritkyiv:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  const threeDaysDiscont = 20;
```
mykola-koval, line 32: use more abstract var names like `smallDiscount` `largeDiscount`

mkyloa (approved):

PR #1906, opened 2022-10-18
solution

sergosdv1:


spiritkyiv (changes requested):
Maybe you should use variable to discont? what is 20 and 50 if i haven`t read task

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const tax = 40;
+  const discSmall = 20;
+  const discBig = 50;
+
+  switch (true) {
+    case days >= 7:
```
mykola-koval, line 36: you don't need switch here
and switch (true) is almost always a bad idea

mkyloa (changes requested):

mkyloa (approved):

PR #1908, opened 2022-10-18
Solution

SerhiiVovchenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = days * 40;
+
+  if (days >= 7) {
+    sum -= 50;
+  } else if (days >= 3) {
+    return (sum - 20);
+  }
```
Bohdan-Maksymiuk, line 37: avoid such nesting `else if` conditions

Bohdan-Maksymiuk: what a magic numbers 🪄

Bohdan-Maksymiuk (changes requested):
Good job, fix comments

polosanya (changes requested):
Almost perfect! Please pay attention to your varibles naming style

polosanya (approved):
Well done!

PR #1910, opened 2022-10-18
transportation-on-vacation solution

Svitlana-Yudina:


Stepico (approved):
Good job, interesting solution to rewrite the basicDiscount instead of creating a second variable as I did

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  let basicDiscount = 50;
```
mykola-koval, line 32: i'd use two variables and avoid rewriting this one
it is not a basic discount if it changes after all

mkyloa (approved):

PR #1912, opened 2022-10-18
Solution

sofiia-dovhan:


Anna-Shapovalova (commented):
GJ!) Attagirl) :fire: I do like your solution with the switch) But U need to use var-s due to the checklist)

Anna-Shapovalova (approved):

G-YaSha (approved):
Gj!

PR #1917, opened 2022-10-19
Solution

Oleh-Burii:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const oneDay = 40;
```
polosanya, line 31: What does `oneDay` mean? I think it's better to give more informative name

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
+  const sum = days * oneDay;
```
polosanya, line 34: The same

polosanya (changes requested):
Nice job! Please pay attention to your variable naming style

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40 * days;
```
polosanya, line 31: It would be great to add variable 40, as it is important value

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const minRentDiscount = 20;
+  const maxRentDiscount = 50;
+
+  if (days >= 7) {
+    return rentCost - maxRentDiscount;
+  }
+
+  if (days >= 3) {
```
polosanya, line 39: The same for days. Better to clarify that these are days for different discounts

polosanya (changes requested):
Almost perfect! Please provide one small fix

polosanya (approved):
Good job!

PR #1921, opened 2022-10-19
Solution

Anna-Shapovalova:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  const rent = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return rent * days;
+  }
+
+  if (days >= 3 && days < 7) {
```
Oleksii-Siviolov, line 39: I think, that it will still work like this. Try it.
```suggestion
  if (days < 7) {
```

Oleksii-Siviolov (approved):
Nice, I have similar solution. But my variable names are cooler ;) (hope they won't reject my job because of them XD)

src/calculateRentalCost.js
mykola-koval, line 39: if code runs past the previous `if` it means that `days` always >= 3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return rent * days - smallDiscount;
+  }
+
+  return rent * days - bigDiscount;
```
mykola-koval, line 43: checklist 3

mkyloa (changes requested):

src/calculateRentalCost.js
Anna-Shapovalova: You are totally right)

mkyloa (approved):
gj 👍

PR #1922, opened 2022-10-19
Solution

MikhailKucherina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return days * dayCost;
+  }
+
+  if (days < 7) {
+    return days * dayCost - 20;
+  }
+
+  if (days >= 7) {
+    return days * dayCost - 50;
+  }
```
MXMUK, line 43: last check is not needed, just
`return days * dayCost - 50;`

MXMUK (approved):
gj 🚀

src/calculateRentalCost.js
Olexandr6: +1

G-YaSha (approved):

PR #1924, opened 2022-10-19
Solution

Oleksandr-Stryzhavchuk:


masha-girya (approved):
I think everything is good!) but if + else if is it ok about Cheacklist?

mkyloa (approved):

Oleksandr-Stryzhavchuk:
> I think everything is good!) but if + else if is it ok about Cheacklist?

if + else if is it ok, but mentor said that it would be better through two separate if

PR #1926, opened 2022-10-19
Solution

alpopkov:


Anna-Shapovalova (approved):
GJ!) But maybe it will be better to name the discount for 7 and more days more clearly (it is not ex value actually)?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const discount = 20;
+  const exDiscount = 50;
```
mykola-koval, line 33: bad var name

mkyloa (approved):

PR #1928, opened 2022-10-19
Task completed

vadymbaranov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  }
+
+  if (days < 7) {
+    return days * rent - discountHalfWeek;
+  }
+
+  if (days >= 7) {
+    return days * rent - discountWeek;
+  }
 }
```
Oleksii-Siviolov, line 46: can save two lines here )

```suggestion
  return days * rent - discountWeek;
}
```

Oleksii-Siviolov (approved):
solution is good, but I have a small suggestion.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
```
Olexandr6, line 31: `rent` of what? Give more descriptive name

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const discountWeek = 50;
+  const discountHalfWeek = 20;
```
Olexandr6, line 33: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) the 5ht point

src/calculateRentalCost.js
Olexandr6: +1, Let's do it

G-YaSha (approved):
Your solution is good👍, but can be improved

PR #1930, opened 2022-10-19

MXMUK:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const sum = days * 40;
```
referee1988, line 32: You use "magic numbers" in your code. I think it's better to assign all constant values to variables, and use variables in formulas. But the checklist says nothing about it. Therefore, approve.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+
+  if (days < 3) {
+    return sum;
+  }
+
+  if (days < 7) {
+    return sum - 20;
+  }
+
+  return sum - (Math.floor(days / 7) * 50);
```
referee1988, line 42: According to the checklist, everything is correct - that is approved.
But I would not do calculations in return. It is better to make them earlier, and return a ready-made value. This makes the code easier to read.

VikNikolaiev (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const dayCost = 40;
+  const sevenOff = 50;
+  const threeOff = 20;
```
mykola-koval, line 34: bad variable names, don't attach them to days, use more abstract entities like `small` `large`

mkyloa (approved):

PR #1934, opened 2022-10-20
Solution calcRentalCost

sssolomia:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
```
Andriy-Shymkiv, line 33: i think better to use this names for this variables

`const threeDaysDiscount
const sevenDaysDiscount`

andriy-shymkiv (approved):
GJ

src/calculateRentalCost.js
sssolomia: maybe it will be better

G-YaSha (approved):

PR #1938, opened 2022-10-22
Solution

TarasV123:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    amount = (days * dayRentPrice) - 20;
+
+    return amount;
+  }
+
+  if (days >= 7) {
+    amount = (days * dayRentPrice) - 50;
```
AlisaVlasova, line 47: move 'magic' numbers into variables

AlisaVlasova (changes requested):

nazarmatsevych (changes requested):
You still need to remove 'magic' numbers,
You also can move "days * dayRentPrice" to variable,
And you can immediately return the result instead of creating "amount"

If you have any questions, feel free to ask them in fe_chat2

src/calculateRentalCost.js
```diff
@@ -28,25 +28,18 @@
...
-
-    return amount;
+    return (days * dayRentPrice);
   }
 
   if (days >= 3 && days < 7) {
-    amount = (days * dayRentPrice) - 20;
-
-    return amount;
+    return (days * dayRentPrice) - 20;
```
AlisaVlasova, line 38: 20 and 50 are also magic)

AlisaVlasova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return (days * dayRentPrice);
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * dayRentPrice) - 20;
+  }
+
+  if (days) {
+    return (days * dayRentPrice) - 50;
+  }
```
TarasHoliuk, line 43:
```suggestion
  return (days * dayRentPrice) - 50;
```
if is redundant here, since `days` is an obligate  argument in the function (always defined)

TarasHoliuk (changes requested):

src/calculateRentalCost.js
TarasV123: why??? I don't understand

TarasV123: what i need to change else?

tetiana-ivanchenko (changes requested):
I see that you don't understand some of the comments, you didn't fix it and just ask for approval again. In this case, when you do not understand something, feel free to ask about it in fe_chat2, this will help us unblock you faster, and will help you move faster, because: if someone rejected your work, and you did not correct the comment, then your work will be rejected again with some comments.

and about "magic" numbers - if you use some numbers that are not explicit, just move them to a variable with an explicit name, then if someone looks at your code, they will understand that your number is some discount and it will help make your code more readable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
-  // write code here
+  const dayRentPrice = 40;
+  let discount = 50;
+
+  if (days < 3) {
+    return (days * dayRentPrice);
+  }
+
+  if (days >= 3 && days < 7) {
+    discount = 20;
```
AlisaVlasova, line 39: let`s not reassign discount, init two constants with different discount type

AlisaVlasova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return (days * dayRentPrice);
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * dayRentPrice) - discount1;
+  }
+
+  return (days * dayRentPrice) - discount2;
```
nazarmatsevych, line 43: It's not a good idea to use numbers in variable names, what about firstDiscount and secondDiscount, or bigDiscount and smallDiscount

nazarmatsevych (changes requested):
Almost done,
check comment ;)

tetiana-ivanchenko (approved):

PR #1939, opened 2022-10-24
Solution

petrolozynskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
```
ivanchuk-yurii, line 31: set `total = 40 * days` here and then inside `if` only subtract

ivanchuk-yurii (changes requested):

stanislav-topikha (changes requested):
fix comment
image

SanyaBratashchuk (approved):

PR #1955, opened 2022-10-30
Solution

jaxxxon78:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    rentalCost = days * 40;
+  }
+
+  if (days >= 3 && days <= 6) {
+    rentalCost = (days * 40) - 20;
+  } else if (days >= 7) {
+    rentalCost = (days * 40) - 50;
+  }
```
SanyaBratashchuk, line 41: move all 'magic' numbers to the variables and get rid of `else if` statement

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,10 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const perDayCost = 40;
+  const discount = (days >= 7 ? 50 : (days >= 3 ? 20 : 0));
```
alexandra-protyanova, line 32: it is really bad practice to use nested ternaries

alexandra-protyanova (changes requested):

ivanchuk-yurii (approved):

PR #1958, opened 2022-10-31
solution calculateRentalCost

vadiimvooo:


VladimirEdelmann (approved):
You don't need to add package.json file

valerii-derkach (approved):

PR #1960, opened 2022-10-31
Solution

KrisMakarovska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return 40 * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (40 * days) - 20;
+  }
+
+  if (days >= 7) {
+    return (40 * days) - 50;
+  }
```

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
KrisMakarovska: Done)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return dayPrice * days;
```
Bohdan-Maksymiuk, line 36:
```suggestion
    return dayPrice * days;
```
pass `dayPrice * days` into variable too

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    return (dayPrice * days) - smallDiscount;
+  }
+
+  if (days >= 7) {
+    return (dayPrice * days) - bigDiscount;
+  }
```
Bohdan-Maksymiuk, line 45: pass days count needed for discount into variables too

Bohdan-Maksymiuk (changes requested):
Good, the last comments)

src/calculateRentalCost.js
KrisMakarovska: +

src/calculateRentalCost.js
KrisMakarovska: All right) done as well

Bohdan-Maksymiuk (approved):

PR #1964, opened 2022-11-03
Solution

andriibilous4:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const total = days * price;
+
+  if (days >= 7) {
+    return total - 50;
```
loralevitska, line 35: `avoid magic numbers` - fix it now in the whole solution and for the future. If you meet such a comment, it means that all numbers that do not carry anything semantic in the code should be made into constants whose name will explain what these numbers mean

E.g `const threeDayRentalDiscount = 20;`

loralevitska (changes requested):
almost done)

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
  * @return {number}
  */
 
+const DAYS_WITH_LOW_DISCOUNT = 3;
+const DAYS_WITH_HIGH_DISCOUNT = 7;
+const LOW_DISCOUNT = 20;
+const HIGH_DISCOUNT = 50;
```
liliia-livandovskaya, line 33: it's better to use camelCase there

liliia-livandovskaya (approved):
GJ! Pay attention to comments

PR #1965, opened 2022-11-03
Added the solution

zynkro:


jstmpelowycz (commented):
what is 40? what is 20? what is 50?
when such values appear in our code, we call it ✨Magic Numbers✨ 
let's put these values into variables so that it's clear for a reader what's going on
```
const basicDiscount = 20;
const greatDiscount = 50
...
```
also, leaving else-if statement as the last one is not a good practice, you can handle the task by doing the following
```
if (days >= 7) {
  ...
}

if (days >= 3) {
  ...
}

...
```

jstmpelowycz (changes requested):
let's apply some fixes so that your code looks great!

zbyn3k:
Hi Dmytro, thanks for your feedback! I've updated the code following your suggestions.

jstmpelowycz (approved):
a-w-e-s-o-m-e! keep moving on!

PR #1966, opened 2022-11-04
solution

Tsvigunova:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const total = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return total - 20;
+  } else if (days >= 7) {
+    return total - 50;
+  } else {
+    return total;
+  }
```

Bohdan-Maksymiuk: Also avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):

liliia-livandovskaya (approved):
Great job! 🔥

PR #1968, opened 2022-11-04
Solution

RomaSheva1987:


nazarmatsevych (changes requested):
What is 40, 20 and 50?
Move this ✨Magic Numbers✨into variables so it will become clear to others 
who will look at your code

tetiana-ivanchenko (approved):

PR #1970, opened 2022-11-05
Solution

kuzhelek:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
+
+  if (days >= 7) {
+    return total - 50;
```
qoa11a, line 34: Move number 50 to a variable so it's understandable what this number represents.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days >= 3) {
+    return total - 20;
```
qoa11a, line 36: Same here.

qoa11a (changes requested):

qoa11a (approved):

PR #1971, opened 2022-11-08
Solution

anton-sosonskyi:


stanislav-topikha (changes requested):
lets put every number that u use to the var, and give them explicit names

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const carCost = 40;
+  const off20 = 20;
+  const off50 = 50;
+  const trial = 3;
+  const week = 7;
```
stanislav-topikha, line 36:
```suggestion
  const carCost = 40;
  const smallDiscount = 20;
  const bigDiscount = 50;
  const smallPeriod = 3;
  const bigPeriod = 7;
```
_explicit_ names

stanislav-topikha (changes requested):

stanislav-topikha (approved):

PR #1978, opened 2022-11-10
Solution

constantweb:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const cost = days;
+
+  if (days >= 7) {
+    return cost * 40 - 50;
+  } else if (days >= 3) {
+    return cost * 40 - 20;
+  } else {
+    return cost * 40;
+  }
```
tetiana-ivanchenko, line 39: move "magic" numbers to variables. I mean: if someone looks at your code, it's implicit, what is 40, 20, 50?
create some variables that will explain the meaning of the numbers.

tetiana-ivanchenko (changes requested):

AlisaVlasova (approved):

PR #1979, opened 2022-11-11
Solution

IVxVI:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    return (days * 40) - 50;
+  }
```
SanyaBratashchuk, line 41: move all magic numbers to the variables and get rid of 1 if

SanyaBratashchuk (changes requested):

ivanchuk-yurii (approved):

PR #1982, opened 2022-11-12
Solution

Malva37:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
nzrii-kysel, line 32: could use a variable for `days * 40`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
```
nzrii-kysel, line 35: `days` is always >= 3 at this point
```suggestion
  if (days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
```
nzrii-kysel, line 41: redundant `if` statement

0xNazarii (changes requested):
well done

src/calculateRentalCost.js
alexandra-protyanova: agree

alexandra-protyanova (approved):
Well done 👍

PR #1983, opened 2022-11-13
Solution calculateRentalCost

panianchuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+      rentCostTotal = (rentCostOneDay * days) - biggerDiscount;
+      break;
+
+    case days >= 3 && days < 7:
+      rentCostTotal = (rentCostOneDay * days) - smallerDiscount;
+      break;
+
+    default:
+      rentCostTotal = rentCostOneDay * days;
+  }
```
SanyaBratashchuk, line 47: There is no need in switch(true), use if

SanyaBratashchuk (changes requested):

ivanchuk-yurii (approved):

PR #1984, opened 2022-11-13
Solution

Dmytro-Litinskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return oneDayRent * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    return oneDayRent * days - threeDayDiscount;
+  } else {
+    return oneDayRent * days - sevenDayDiscount;
+  }
```
tetiana-ivanchenko, line 43:
```suggestion

  return oneDayRent * days - sevenDayDiscount;
 
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  const oneDayRent = 40;
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
+
+  if (days < 3) {
+    return oneDayRent * days;
+  }
+
+  if (days >= 3 && days < 7) {
```
tetiana-ivanchenko, line 39: also, it you put condition for 7 days upper, than for 3 dsys, you don't need two conditions here

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1988, opened 2022-11-14
Solution

OleksandrOse:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
...
+  const price = 40;
+
+  switch (true) {
+    case days < 3:
+      return price * days;
+    case days < 7:
+      return price * days - 20;
+    default:
+      return price * days - 50;
+  }
```

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const price = 40;
+  const threeDays = 3;
+  const sevenDays = 7;
```
Bohdan-Maksymiuk, line 34: not the best names for days
what if we will change number? You will need to rename variable
better smth like `daysForSmallDiscount`

Bohdan-Maksymiuk (approved):
Good job, take my comment into consideration

Bohdan-Maksymiuk (approved):

Bohdan-Maksymiuk (approved):
no need to re-request review after approve)

PR #1989, opened 2022-11-15
Solution

fallowlone:


valerii-derkach (changes requested):
you repeat this code three times
days * priceForDay

let's create a variable

valerii-derkach (approved):

PR #1994, opened 2022-11-15
solution

lefteris-L:


IoghiStudio (approved):
Awesome!

lefteris-L:
> Awesome!

Thanks 😃

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
+
+  if (days < 3) {
+    return cost;
+  }
+
+  return cost - (days < 7 ? 20 : 50);
```
meforoma, line 37: might sound like a tiny thing, BUT you have a pack of `magic numbers`:
3, 7, 20 and 50 are driving the logic
another developer who has no clue about the task readme - would have hard times thinking of what do they stand for.

Something like this will be story-like and easily understandable:
```
const daysNoDiscount = 3;
const discountUnderSevenDays = 20;
const discountAboveSevenDays = 50;

if (days < daysNoDiscount) {
  return cost;
}

const discountPerStayDuration = days < 7
  ? discountUnderSevenDays
  : discountAboveSevenDays

return cost - discountPerStayDuration;
```

meforoma (changes requested):
let's make our code self-explanatory

meforoma (approved):
🚀

PR #2001, opened 2022-11-19
develop

volodymyr4501:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
```
loralevitska, line 41:
```suggestion
    return days * 40 - 50;
```

loralevitska (changes requested):
Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content (40, 20, 50)

PR #2003, opened 2022-11-19
Solution

halynarydasova:


jstmpelowycz (changes requested):
what is 40? what is 20? what is 50?
when such values appear in our code, we call it ✨Magic Numbers✨ 
let's put these values into variables so that it's clear to a reader what's going on
```
const basicDiscount = 20;
const greatDiscount = 50
...
```
also, leaving else-if statement as the last one is not a good practice, you can handle the task by doing the following
```
if (days >= 7) {
  ...
}

if (days >= 3) {
  ...
}

...
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    totalAmount -= greatDiscount;
+
+    return totalAmount;
+  }
+
+  if (days >= 3) {
+    totalAmount -= basicDiscount;
+
+    return totalAmount;
+  }
```
jstmpelowycz, line 46:
```suggestion
  if (days >= 7) {
    return totalAmount - greatDiscount;
  }

  if (days >= 3) {
    return totalAmount - basicDiscount;
  }
```

jstmpelowycz (approved):
nicely done!

PR #2004, opened 2022-11-19
Transportation on vacation

OlexandraKorol:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    sum += rentalConst;
+  }
+
+  if (days >= 3 && days < 7) {
+    sum -= 20;
+  }
+
+  if (days >= 7) {
+    sum -= 50
+  }
```
GabrielMaksimovich, line 44: p5, lets also name 20,50

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
+  const rentalConst = 40;
```
GabrielMaksimovich, line 32: p2(let)

MaksymBaturinskyi (changes requested):

src/calculateRentalCost.js
OlexandraKorol: Fixed!

src/calculateRentalCost.js
OlexandraKorol: "sum" it's a value thats will change, so I can't use const here. Or u mean something else?

GabrielMaksimovich: I mean that we can do it without let, but probably I though wrong about this. Lets see what menors say)

MaksymBaturinskyi (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  const basicDiscount = 20;
+  const bigDiscount = 50;
+
+  for (let i = 1; i <= days; i++) {
+    sum += rentalConst;
+  }
```
solaryasha, line 38: Do you really need a loop here?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+  if (days >= 3 && days < 7) {
+    sum -= basicDiscount;
+  }
+
+  if (days >= 7) {
```
solaryasha, line 44: The number of days could also change, let's move them to the variables

solaryasha (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalConst = 40;
```
ivan-kadykalo, line 31:
```suggestion
  const dayRentalConst = 40;
```

ivan-kadykalo (approved):

PR #2005, opened 2022-11-19
Solution

romanblyashuk94:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,10 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceOfDay = 40;
+  const discount = (days >= 7) ? 50 : (days >= 3) ? 20 : 0;
+
```
GabrielMaksimovich, line 33: You may also name variables for 50, 20, cause it's look like "magic numbers". p5

MaksymBaturinskyi (changes requested):

MaksymBaturinskyi (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceOfDay = 40;
+  const discount20 = 20;
+  const discount50 = 50;
+  const currentDiscount = (days >= 7) ? discount50
+    : (days >= 3) ? discount20 : 0;
```
solaryasha, line 35: Don't use nested ternary operators, let's use basic if-else instructions

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceOfDay = 40;
+  const discount20 = 20;
+  const discount50 = 50;
```
solaryasha, line 33: What if the value of the discount will change?

src/calculateRentalCost.js
solaryasha, line 35: The number of days could also change, let's move it to the variables

solaryasha (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const daysForSmallDiscount = 3;
+  const daysForBigDiscount = 7;
+
+  let currentDiscount = 0;
+
+  if (days >= daysForBigDiscount) {
+    currentDiscount = bigDiscount;
+  } else if (days >= daysForSmallDiscount) {
+    currentDiscount = smallDiscount;
+  }
```
solaryasha, line 43: Avoid using the `let ` keyword. Let's return value immediately instead

solaryasha (changes requested):

ivan-kadykalo (approved):
🔥

PR #2012, opened 2022-11-20
Solution

Anastasia4sik:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let amount = 40 * days;
+
+  if (days >= 3 && days < 7) {
+    amount -= 20;
+  } else if (days >= 7) {
+    amount -= 50;
+  }
```
GabrielMaksimovich, line 37: p1,p3 checklist

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let amount = 40 * days;
+
```
GabrielMaksimovich, line 32: p2

MaksymBaturinskyi (changes requested):

src/calculateRentalCost.js
Anastasia4sik: I can`t use here const variable, cause it changes in the if

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const amount = 40 * days;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return amount - minDiscount;
+  } else if (days >= 7) {
+    return amount - maxDiscount;
+  }
```
GabrielMaksimovich, line 39: p3, we don't need use 'else'

MaksymBaturinskyi (changes requested):

MaksymBaturinskyi (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  const amount = 40 * days;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return amount - minDiscount;
+  }
+
+  if (days >= 7) {
```
solaryasha, line 39: The number of days could also change, let's move it to the variables

solaryasha (changes requested):

solaryasha (approved):
Nice!

PR #2018, opened 2022-11-21
Solution

Deniskakaka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const cost = 40;
+
+  if (days >= 3 && days < 7) {
+    return days * cost - smallDiscount;
+  } else if (days >= 7) {
```
Viktor-Kost, line 37: Try to avoid else if conditions to make your code clear. 
You can use `switch` statement or even just a few of `if` statements

Viktor-Kost (changes requested):
Good job!

Let's optimize your code

liliia-livandovskaya (approved):
Great job! 👍

PR #2032, opened 2022-11-21
Solution

4ebupel:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const smallDiscount = 20;
+  const hugeDiscount = 50;
+
+  if (days < 3) {
+    return 40 * days;
```
AntonMartynes, line 35: Put the daily rent in a variable

AntonMartynes (changes requested):

src/calculateRentalCost.js
4ebupel: Fixed

AntonMartynes (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < 3) {
+    return dailyRent * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (dailyRent * days) - smallDiscount;
+  }
+
+  if (days >= 7) {
```
solaryasha, line 43: Let's move the days count to variables. They could be changed too.

solaryasha (changes requested):

solaryasha (approved):

PR #2036, opened 2022-11-21
solution

oleksandrnikora:


vrima1221 (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const amount = days * 40;
+  const basicDiscount = 50;
+  const smallDiscount = 20;
+
+  return days >= 7
+    ? amount - basicDiscount : days >= 3 && days < 7
+      ? amount - smallDiscount : amount;
 }
```
ivan-kadykalo, line 38: Let's move the days count to variables. They could be changed too.
And also don't nest ternary operators into ternary operators. It will be better to use if(){} statement here.

ivan-kadykalo (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const amount = days * 40;
+  const basicDiscount = 50;
+  const smallDiscount = 20;
+  const lengthSmallDiscount = 3;
+  const lengthBasicDiscount = 7;
```
solaryasha, line 35:
```suggestion
  const daysForSmallDiscount = 3;
  const daysForBasicDiscount = 7;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const amount = days * 40;
```
solaryasha, line 31:
```suggestion
  const basicCost = days * 40;
```

solaryasha (changes requested):
Let's use proper namings for our variables

solaryasha (approved):
👍

ivan-kadykalo (approved):

PR #2040, opened 2022-11-22
Solution

sercansimsek:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let cost = 40;
+
+  if (days >= 7) {
+    return (cost = cost * days - 50);
+  }
+
+  if (days >= 3) {
+    return (cost = cost * days - 20);
+  }
```
meforoma, line 39: 50 and 20 are better be named - to make code more explanatory
please provide variables for both

meforoma (changes requested):
let's add variable to avoid 'magic numbers'

grslckr (approved):
Cool!

PR #2041, opened 2022-11-22
solution

Alireza-Mohammadhossein:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
```
mgrinko, line 35: no need to check that `days >= 3` because we would have already returned the value if the condition was true in the prev `if`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40 - 20;
```
mgrinko, line 36: you repeated `days * 40` 3 times

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  if (days >= 7) {
```
mgrinko, line 39: this condition is always true at this point

mgrinko (changes requested):

src/calculateRentalCost.js
Alireza-Mohammadhossein: I fix it, can you check it again?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const cost = days * 40;
+
+  if (days >= 7) {
+    return cost - 50;
+  }
+
+  if (days >= 3) {
+    return cost - 20;
+  }
```
meforoma, line 39: you're done logic-wise 👍 
you can also improve the code readability by telling what numbers 20 and 50 are:
```
const discountLower = 20;
const discountHigher = 20;
```

meforoma (approved):

PR #2042, opened 2022-11-22
Solution

UmizDemud:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,7 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days * 40 - ((days >= 7) ? 50 : (days >= 3) ? 20 : 0);
```
mgrinko, line 31: please never use more that 1 ?: operator in one expression
it is not readable at all

mgrinko (changes requested):

src/calculateRentalCost.js
UmizDemud: Now that I have to re-read it, feeling the pain.. Correcting

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = days * 40;
+
+  if (days >= 7) { // discount is $50 for 7 or more days
+    cost -= 50;
+  } else if (days >= 3) { // discount is $20 for 3 or more days
+    cost -= 20;
+  }
```
meforoma, line 37: fine, but could be even simpler
```
if (days >= 7) {
  return cost - 50
}

if (days >= 3) {
  return cost - 20
}
```

and please make 20 and 50 meaningful - i.e. assign to variables with explanatory names

meforoma (changes requested):
see the comment)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const cost = days * 40;
+  const discountForSeven = 50;
+  const discountForThree = 20;
+
+  if (days >= 7) { // discount is $50 for 7 or more days
+    return cost - discountForSeven;
+  } else if (days >= 3) { // discount is $20 for 3 or more days
+    return cost - discountForThree;
+  }
```
meforoma, line 39: no need for `else if` when if statement returns something:

```suggestion
  if (days >= 7) { // discount is $50 for 7 or more days
    return cost - discountForSeven;
  }
  
  if (days >= 3) { // discount is $20 for 3 or more days
    return cost - discountForThree;
  }
```

meforoma (approved):

PR #2043, opened 2022-11-22
Finished Tasks

JzSol:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+
+  if (days >= 7) {
+    return days * price - 50;
+  } else if (days >= 3) {
```
mgrinko, line 35: checklist

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+    return days * price - 20;
+  } else {
+    return days * price;
```
mgrinko, line 38: `days * price` is written 3 times

mgrinko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const standart = days * 40;
+
+  if (days >= 7) {
+    return standart - 50;
+  } else if (days >= 3) {
+    return standart - 20;
+  }
+  return standart;
```
mgrinko, line 38:
```suggestion
  if (days >= 7) {
    return standart - 50;
  }
  
  if (days >= 3) {
    return standart - 20;
  }

  return standart;
```

mgrinko (approved):

PR #2046, opened 2022-11-22
Solution

CharAnagn:


UmizDemud (approved):
Algorithm fast for reusability 👍

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  const basicDiscount = 20;
+  const extendedDiscount = 50;
+
+  const rentalCost = days * dayCost;
+
+  return days > 6
+    ? rentalCost - extendedDiscount
+    : days > 2
+      ? rentalCost - basicDiscount
+      : rentalCost;
```
meforoma, line 42: nested ternaries are not considered good practice
better replace with `if (...) { return ... }`

meforoma (changes requested):
let's not use nested ternaries

meforoma (approved):

PR #2047, opened 2022-11-22
solution

Grayfox147:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+  let total;
+
+  if (days < 3) {
+    total = dailyRent * days;
```
UmizDemud, line 35: These can be return statements for slightly faster code execution

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  } else {
+    if (days >= 3 && days < 7) {
+      total = (dailyRent * days) - 20;
```
UmizDemud, line 38: When the code comes to this else, we know that ```!(days < 3) or (days >= 3) same thing``` so you can omit that first condition on line 37:

```js
  } else if (days < 7) { // lline36 and line37 combined

```

UmizDemud (changes requested):
Good job, maybe you can consider these comments, I'll check again most of the day

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
```
JzSol, line 31: dailyRent can also include days, so you don't have to use days in your if statements!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  let total;
```
JzSol, line 32: You can just use return instead of making a new variable.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days < 3) {
+    total = dailyRent * days;
+  } else {
+    if (days >= 3 && days < 7) {
+      total = (dailyRent * days) - 20;
+    } else {
+      if (days >= 7) {
+        total = (dailyRent * days) - 50;
```
JzSol, line 41: Check out the checklist!

JzSol (changes requested):

src/calculateRentalCost.js
Grayfox147: Done!

src/calculateRentalCost.js
Grayfox147: Done!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+
+  if (days < 3) {
+    return dailyRent * days;
+  } else if (days < 7) {
+    return (dailyRent * days) - 20;
+  } else {
```
NgineZZ, line 37: I like the idea of declaring variable 'dailyRent' and the use of if/else if/else statement is perfect. One recommandation from myside you can remove the last else statement and can only return "(dailyRent * days) - 50" this. It will works the same.

NgineZZ (approved):
Nicely done.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+
+  if (days < 3) {
+    return dailyRent * days;
```
mgrinko, line 34: please do this calculation once 
currently it is repeated 3 times

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days < 7) {
```
mgrinko, line 35: you don't need else after if with return

mgrinko (changes requested):

src/calculateRentalCost.js
Grayfox147: Done!

src/calculateRentalCost.js
Grayfox147: Done!

naziamoff (changes requested):
- let's try to use only 2 if's here
- let's create separate variables for discounts

src/calculateRentalCost.js
```diff
@@ -28,20 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  const dailyRent = 40;
-  const dailyCost = (dailyRent * days);
+  const dailyCost = (40 * days);
+  const weeklyDiscount = dailyCost - 20;
+  const monthlyDiscount = dailyCost - 50;
```
naziamoff, line 33: that's not the discount 😄 
Take a deeper look at what is weekly/monthly discount

naziamoff (changes requested):

meforoma (approved):
🏅

naziamoff (approved):

PR #2048, opened 2022-11-22
Solution

YakovlewDan:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+  const weekDiscount = 50;
+
+  const regularMinDays = 3;
+  const weekMinDays = 7;
+
+  if (days < regularMinDays) {
+    return fullCost;
+  } else if (days < weekMinDays) {
+    return fullCost - regularDiscount;
+  }
```
illiashcherbak, line 44: I would use "if" condition several times than "else if" for better readability. but it's ok

illiashcherbak (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const basicDayCost = 40;
+  const fullCost = days * basicDayCost;
+  const regularDiscount = 20;
+  const weekDiscount = 50;
+
+  const regularMinDays = 3;
+  const weekMinDays = 7;
```
solaryasha, line 38: What if we change the number of discount days? (for example, for 30days-month)

You should use abstract variable naming

solaryasha (changes requested):

solaryasha (approved):

PR #2049, opened 2022-11-22

Mikhaylo-Orlov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 0 && days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
+  } else {
```
vlad-demchuk, line 35: checklist
image

vlad-demchuk (changes requested):

loralevitska (changes requested):
hi! almost done). Let's move magic numbers to variables. For example 
```
const SEVEN_DAYS_DISCOUNT = 50
```

AlisaVlasova (approved):

PR #2052, opened 2022-11-22
solved

alaaashkar:


jackelashkar (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+
+  if (days < 3) {
+    total = 40 * days;
```
meforoma, line 34: you use `40 * days` three times - deserves for a variable already

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days >= 3 && days < 7) {
+    total = (40 * days) - 20;
+  } else {
+    total = (40 * days) - 50;
```
meforoma, line 38: 20 and 50 are currently what's called a `magic numbers` - noone knows what they are
assign them to a variable to let your fellow-devs know what it is

meforoma (changes requested):
you can avoid chaining if..elseif 
```
const cost = days * 40;

if (days >= 7) {
  return cost - bigDiscount;
}

if (... >= 3) {
  return cost - smallDiscount;
}

return cost;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days < 3) {
+    total = price * days;
+  } else if (days >= smallDiscountDays && days < bigDiscountDays) {
+    total = (price * days) - smallDiscount;
+  } else {
+    total = (price * days) - bigDiscount;
+  }
+
+  return total;
```
meforoma, line 46: and what about unnecessary `else if` chaining?
```
const cost = days * 40;

if (days >= 7) {
  return cost - bigDiscount;
}

if (... >= 3) {
  return cost - smallDiscount;
}

return cost;
```

meforoma (changes requested):
and now let's make it concise

meforoma (approved):
🎉

PR #2055, opened 2022-11-22
solution

mmahfuzi:


lefteris-L (approved):
You could use variables here, but great job still.

mmahfuzi:
> You could use variables here, but great job still.

You mean for the discounts? 
Thanks :)

meforoma (changes requested):
see the [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md#:~:text=%7B%0A%20%20return%20y%3B%0A%7D-,GOOD%20EXAMPLE,-%3A) in regard to using `else`

and the reversed logic will simplify your conditionals
```
if (days >= 7) {return}
if (days >= 3) {return}
return
```

and yes, 20 and 50 should be meaningful -> assign to variables with explanatory names

lefteris-L:
> > You could use variables here, but great job still.
> 
> You mean for the discounts? Thanks :)

Exactly, as well as `days * 40` since it's used 3 times. You're very welcome

meforoma (approved):
code does the job, yet could be a tiny bit shorter 🙂 
```
const totalCost = days * dailyRent;

if (days >= 7) {
  return totalCost - moreDiscount;
}

if (days >= 3) {
  return totalCost - lessDiscount;
}

return totalCost;
```

PR #2057, opened 2022-11-23
Solution

Rijo-J:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = days * 40;
+  const withDiscount = {
+    [days > 2]: price - 20,
+    [days > 6]: price - 50,
```
sercansimsek, line 34: It could be better if you create variables for the discounts.

sercansimsek (changes requested):

sercansimsek (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const price = 40;
+  const total = days * price;
+  const lowerDiscount = 20;
+  const highDiscount = 50;
+
+  const withDiscount = {
+    [days > 2]: total - lowerDiscount,
+    [days > 6]: total - highDiscount,
+  };
```
meforoma, line 39: allright, that's already a deeper level of objects understanding 😎 👍

meforoma (approved):
advanced object usage 🚀

PR #2058, opened 2022-11-23
Rental cost is calculation

kateryna-stemblivska:


Kavster92 (commented):
compact and clean .js nice work

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dayRent = 40;
+
+  if (days >= 7) {
+    return (dayRent * days - 50);
+  }
+
+  if (days >= 3) {
+    return (dayRent * days - 20);
```
liliia-livandovskaya, line 38: add variables instead of numbers and use them 
`const basicDiscount = 20;
  const extendedDiscount = 50;`

liliia-livandovskaya (changes requested):
Pay attention to the comments

polosanya (approved):
Good job!

PR #2061, opened 2022-11-23
Solution

NgineZZ:


Kavster92 (commented):
Clean efficient .js, good work

mmahfuzi (changes requested):
Good job, just based on the check list you need to use variables. (instead of using 40 which has been repeated 3 times in your code use a meaningful variable for it, the same for 20 and 50)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const discountMax = 50;
+  const discountMin = 20;
+
+  if (days >= 7) {
+    return dailyRent * days - discountMax;
+  } else if (days >= 3) {
+    return dailyRent * days - discountMin;
+  }
+
+  return dailyRent * days;
```
mgrinko, line 41: this calculation is done 3 times
please do it once

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+  const discountMax = 50;
+  const discountMin = 20;
+
+  if (days >= 7) {
+    return dailyRent * days - discountMax;
+  } else if (days >= 3) {
```
mgrinko, line 37: you don't need else here

mgrinko (changes requested):

mgrinko (approved):

PR #2062, opened 2022-11-23
solution

dolnaro:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let price = 0;
+
+  if (days >= 7) {
+    price = 40 * days - 50;
+  } else if (days >= 3) {
+    price = 40 * days - 20;
```
grslckr, line 36: 50 and 20 should assigned to variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days < 3) {
+    price = 40 * days;
+  }
```
grslckr, line 39: i think this else if condition is unnecessary you can leave it -price = 40 * days- outside the block

grslckr (changes requested):

dolnaro:
refreshed.

dolnaro:
variables changed.

grslckr (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = days * 40;
+  const priceLong = days * 40 - 50;
+  const priceShort = days * 40 - 20;
```
mgrinko, line 33: This calculations should be done only when conditions are met
Not need to calculate them beforehand
```suggestion
  const priceLong = price - 50;
  const priceShort = price - 20;
```

mgrinko (approved):

PR #2063, opened 2022-11-23
Rental Agreement Discount Calculator

Kavster92:
Turned discounts into variables and implemented solution

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRate = 40;
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return (days * dayRate) - threeDayDiscount;
+  }
+
+  if (days >= 7) {
+    return (days * dayRate) - sevenDayDiscount;
+  }
+
+  if (days < 3) {
+    return days * dayRate;
+  }
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -31,17 +31,16 @@ function calculateRentalCost(days) {
   const dayRate = 40;
   const threeDayDiscount = 20;
   const sevenDayDiscount = 50;
-
-  if (days >= 3 && days < 7) {
-    return (days * dayRate) - threeDayDiscount;
-  }
-
-  if (days >= 7) {
-    return (days * dayRate) - sevenDayDiscount;
-  }
+  const cost = (days * dayRate);
 
   if (days < 3) {
-    return days * dayRate;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return (days * dayRate) - threeDayDiscount;
+  }
+
+  if (days >= 7) {
+    return (days * dayRate) - sevenDayDiscount;
+  }
+
+  if (days < 3) {
+    return days * dayRate;
+  }
```
Mohameddahi, line 45: Every is great , but i think here it is better to use return for the rest options without the  IF  condition 
```suggestion
    return days * dayRate;
```

dahimohamed (approved):
Well done !

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRate = 40;
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
+
+  if (days >= 3 && days < 7) {
```
mgrinko, line 35: if you start from the smallest values you could simplify your conditions 

```
if < 3 return
if < 7 return 

return (it is for >= 7)

File: src/calculateRentalCost.js [hunk 3]
```diff
@@ -28,7 +28,21 @@
...
+    return (days * dayRate) - threeDayDiscount;
+  }
+
+  if (days >= 7) {
+    return (days * dayRate) - sevenDayDiscount;
+  }
+
+  if (days < 3) {
+    return days * dayRate;
```
mgrinko, line 44: this calculation is written 3 time 
please do it once

mgrinko (changes requested):

src/calculateRentalCost.js
```diff
@@ -31,17 +31,16 @@ function calculateRentalCost(days) {
...
-    return (days * dayRate) - sevenDayDiscount;
+  if (days < 3) {
+    return cost;
   }
 
-  if (days < 3) {
-    return days * dayRate;
+  if (days >= 3 && days < 7) {
+    return cost - threeDayDiscount;
+  } else {
```
mgrinko, line 42: else if redundant because of return in the prev if

src/calculateRentalCost.js
```diff
@@ -31,17 +31,16 @@ function calculateRentalCost(days) {
...
-
-  if (days >= 7) {
-    return (days * dayRate) - sevenDayDiscount;
+  if (days < 3) {
+    return cost;
   }
 
-  if (days < 3) {
-    return days * dayRate;
+  if (days >= 3 && days < 7) {
```
mgrinko, line 40: `days >= 3` this condition is always true here
because you would have already returned a value in the prev `if`

mgrinko (approved):

PR #2064, opened 2022-11-23
DiDemkiv solution

DiDemkiv:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  }
+
+  if (days >= 7) {
+    finalCost -= 50;
+  } else if (days >= 3) {
+    finalCost -= 20;
+  }
+
+  return finalCost;
 }
```
batstolya, line 44: Your code works, but to be honest it is very difficult to understand it)) It seems to me that we can do the task without a cycle. And then our code will be understandable.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let finalCost = 0;
+
+  for (let day = 0; day < days; day++) {
+    finalCost += 40;
```
batstolya, line 34: CL-1 [VARIABLES] - use variables for the main values so that you can reuse them and give them descriptive names.
how to understand what is 40?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  }
+
+  if (days >= 7) {
+    finalCost -= 50;
```
batstolya, line 38: CL-1 
Also, what is 50? 
for example you can write variable  = discountMax or something different.

batstolya: and in the future it will be easier for you to reuse variables or change their value)

batstolya:
Good job🔥

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
🔥

PR #2065, opened 2022-11-23
solution

dahimohamed:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
```
JzSol, line 31: You can make a variable for days * 40!  It will be much more efficient!

JzSol (changes requested):
Look at the comments I left, everything else looks fine!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    return rent * days - 20;
+  }
+
+  if (days >= 7) {
+    return rent * days - 50;
+  }
+
+  return rent * days;
```
mgrinko, line 41: this calculation is repeated 3 times
please do it once because it must be the same in all 3 places

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+
+  if (days >= 3 && days < 7) {
```
mgrinko, line 33: if you start from < 3 it will simplify your condition

mgrinko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  const rentalCost = rent * days;
+
+  if (days < 3) {
+    return rentalCost;
+  }
+
+  if (days >= 3 && days < 7) {
```
mgrinko, line 38: This condition `days >= 3` is always true
Because you already checked the opposite before and returned if it was true

mgrinko (approved):

PR #2069, opened 2022-11-24
Solution

alyonkaradchuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const price = days * 40;
+
+  if (days >= 3 && days <= 6) {
+    return price - 20;
+  }
+
+  if (days >= 7) {
+    return price - 50;
+  }
```
solaryasha, line 39: Use variables for the main values so that you can reuse them and give them descriptive names

solaryasha (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const daysForMinDiscount = 3;
+  const daysForMaxDiscount = 7;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+  const price = days * 40;
```
solaryasha, line 35: The value of 40 we could store in the variable

solaryasha (approved):

PR #2071, opened 2022-11-24
Solution for transportation on vacation

katerinalex:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 40 * days;
+
+  if (days >= 7) {
+    total = total - 50;
+  } else if (days >= 3) {
+    total = total - 20;
```
tetiana-ivanchenko, line 36: move 40, 20 and 50 in the variables

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
Anastasiia-Svintsova: Seems that you didn't fix the previous comment( Please, check it one more time

Anastasiia-Svintsova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costOfDay = 40;
+  const off7 = 50;
+  const off3 = 20;
```
```suggestion
  const off7 = 50;
  const off3 = 20;
```
What does it mean? Let's use full, descriptive names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  let total = costOfDay * days;
+
+  if (days >= 7) {
+    total = total - off7;
+  } else if (days >= 3) {
+    total = total - off3;
+  }
```
```suggestion
  if (days >= 7) {
    return total - off7;
  } 
  
  if (days >= 3) {
    return total - off3;
  }
```
As for me, it looks better. Btw, you can also add variables for 3 and 7, as we can change the required amount of days for the discount later.

polosanya (changes requested):
Almost perfect! Please refactor your code a bit

polosanya (approved):
Well done!

PR #2078, opened 2022-11-26
Solution

vladyslav-krainyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalFeeForDay = 40;
+  const rentalFee = rentalFeeForDay * days;
+  let discountedRent = 0;
+
+  switch (true) {
```
ivanchuk-yurii, line 35: it's better to use `if else` here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    case days < 3:
+      return rentalFee;
+
+    case days >= 3 && days < 7:
+      discountedRent = rentalFee - 20;
+
+      return discountedRent;
```
ivanchuk-yurii, line 42:
```suggestion
      return rentalFee - 20;
```

ivanchuk-yurii (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rentalFeeForDay = 40;
+  const rentalFee = rentalFeeForDay * days;
+
+  if (days < 3) {
+    return rentalFee;
+  } else if (days >= 3 && days < 7) {
+    return rentalFee - 20;
+  } else if (days >= 7) {
```
ivanchuk-yurii, line 38: 'else' is redundant here

ivanchuk-yurii (approved):

PR #2079, opened 2022-11-27
Solution

dissolveyo:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  switch (true) {
+    case days >= 7:
+      return (days * 40) - 50;
+    case days >= 3:
+      return (days * 40) - 20;
+    default:
+      return days * 40;
+  }
```
SanyaBratashchuk, line 38: move all `magic` numbers to the variable

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const longerThanThree = days >= 3;
+
+  switch (true) {
+    case longerThanSix:
+      return (days * moneyPerDay) - sevenDaysDiscount;
+    case longerThanThree:
+      return (days * moneyPerDay) - threeDaysDiscount;
+    default:
+      return days * moneyPerDay;
+  }
```
ivanchuk-yurii, line 44:
```suggestion
  const rentalCost = 40 * days;
  const sevenDaysDiscount = 50;
  const threeDaysDiscount = 20;

  switch (true) {
    case days >= 7:
      return rentalCost- sevenDaysDiscount;
    case  days >= 3:
      return rentalCost - threeDaysDiscount;
    default:
      return rentalCost;
  }
```
too much variables)

ivanchuk-yurii (approved):

PR #2081, opened 2022-11-28

rafmamedov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
-  // write code here
+  const rent = 40;
+  let sum = days * rent;
+
+  if (days >= 3) {
+    sum -= 20;
+  }
+
+  if (days >= 7) {
+    sum -= 30;
```
DarkMistyRoom, line 39: Move all these numbers to variables

DarkMistyRoom (approved):
Good just try to make variables for each "magic" number

src/calculateRentalCost.js
rafmamedov: Got it, thanks!

PR #2082, opened 2022-11-28
Solution

Ukrainiane-panda:


src/calculateRentalCost.js
```diff
@@ -28,6 +28,18 @@
...
 function calculateRentalCost(days) {
+  const rent = 40;
+  let sum = 0;
+
+  if (days >= 7) {
+    sum = (rent * days) - 50;
+  } else if (days < 3) {
+    sum = (rent * days);
+  } else if (days >= 3 && days < 7) {
+    sum = (days * rent) - 20;
```

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,6 +28,19 @@
...
+
+  switch (true) {
+    case days >= 7:
+      return sum - 50;
+
+    case days < 3:
+      return sum;
+
+    default:
+      return sum - 20;
```
tetiana-ivanchenko, line 42: move 50 and 20 in the variables too

tetiana-ivanchenko (changes requested):

Bohdan-Maksymiuk (approved):

PR #2086, opened 2022-12-01
Solution

RomanTsvindin:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let off = 0;
```
Bohdan-Maksymiuk, line 31: use more specific variable names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    off = 20;
+  }
+
+  if (days >= 7) {
+    off = 50;
+  }
```

Bohdan-Maksymiuk (changes requested):

loralevitska (changes requested):
Much better, keep it up!
 
 1. Output blank lines between variable declarations. Empty lines should be left between different logical blocks (for example, between the part where variables are declared, and then there is a cycle, or data processing, or we return a value from a function)
 2. remove comments, variable names should explain what they contain

loralevitska (approved):
good job!

PR #2087, opened 2022-12-02
Calculate Rental Cost Solution

anna-akhalmosulishvili:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const dayRent = 40;
+  let total = days * dayRent;
+
+  if (days < 3) {
+    total = days * dayRent;
+  } else if ((days >= 3) && (days < 7)) {
+    total -= 20;
+  } else {
+    total -= 50;
+  }
```
nazarmatsevych, line 40: move all this "magic numbers" to variables

nazarmatsevych (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const threeDaysDiscount = 20;
+  let total = days * dayRent;
+
+  if (days < 3) {
+    total = days * dayRent;
+  } else if ((days >= 3) && (days < 7)) {
+    total -= threeDaysDiscount;
+  } else {
+    total -= weekDiscount;
+  }
```
stanislav-topikha, line 42: move remaining numbers to vars 
use early return instead of `if else`

stanislav-topikha (changes requested):

SanyaBratashchuk (approved):

PR #2088, opened 2022-12-02
Solution

tania-boiar:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const rentalCost = days * costPerDay;
+
+  if (days < 3) {
+    return rentalCost;
+  }
+
+  if (days >= 3 && days < 7) {
+    return rentalCost - smallDiscount;
```
Anastasiia-Khudych, line 41: Values here will always be >3, so you can remove `days >= 3`

Anastasiia-Khudych (approved):
Good job!

src/calculateRentalCost.js
tania-boiar: It makes sense. Thank you!

denys-cheporniuk (approved):

PR #2090, opened 2022-12-03
Solution

vitaliy-zviriuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let price = 0;
+
+  if (days < 3) {
+    price = days * 40;
```
TarasHoliuk, line 34: move numbers to variables with meaningful names. It'll make the code clearer. Now it's not clear what is 40, 20, 50.

TarasHoliuk (changes requested):

loralevitska (approved):
good job!

PR #2092, opened 2022-12-04
Initial commit

Dynamizx:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const cost = days * 40;
+
+  if (days >= 7) {
+    return (cost - 50);
+  } else if (days >= 3) {
+    return (cost - 20);
+  } else {
+    return cost;
+  }
```
```suggestion
  if (days >= 7) {
    return (cost - 50);
  } 
  
  if (days >= 3) {
    return (cost - 20);
  }
  
   return cost;
```
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
```
polosanya, line 31:
```suggestion
  const cost = days * 40;
```
Cost of what? Let's make it clear

polosanya (changes requested):
Almost done! Please check yourself with [the checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md).
In this task, you should use variables for all the main values. Please create them)
image

jstmpelowycz (approved):

PR #2093, opened 2022-12-05
RentalCost

Oleksandr-Maslo:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
+  if (days >= 3 && days < 7) {
+    result = result - 20;
+
+    return result;
+  } else if (days >= 7) {
+    result = result - 50;
```
nazarmatsevych, line 38: move all of this "magic numbers" into variables

nazarmatsevych (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const discountMore3Days = 20;
+  const discountMore7Days = 50;
+
+  let result = days * daycost;
+
+  if (days >= 3 && days < 7) {
+    result = result - discountMore3Days;
+
+    return result;
+  } else if (days >= 7) {
```
AlisaVlasova, line 41: you don`t need else condition when using return inside
just make it another if

AlisaVlasova (changes requested):

nazarmatsevych (approved):

PR #2095, opened 2022-12-06
Solution

vasyliev-anton:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  switch (true) {
+    case days < 3:
+      return amount;
+
+    case (days >= 3) && (days < 7):
+      return amount - 20;
+
+    default:
+      return amount - 50;
```
vlad-demchuk, line 41: let's change cases order to avoid unnecessary conditional like `&&`

vlad-demchuk (changes requested):

stanislav-topikha (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+  let amount = Rent * days;
+
+  if (days >= firstDiscountStartDay && days < secondDiscountStartDay) {
+    amount -= firstDiscount;
+  } else {
+    if (days >= secondDiscountStartDay) {
+      amount -= secondDiscount;
+    }
+  }
```
stanislav-topikha, line 47: still looks too complicated, use early return

stanislav-topikha (changes requested):

nazarmatsevych (approved):

PR #2098, opened 2022-12-07
Solution

Vadivel-Babu:


src/calculateRentalCost.js
```diff
@@ -28,6 +28,13 @@
  */
 
 function calculateRentalCost(days) {
+  if (days >= 1 && days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  }
```
qoa11a, line 37: Your solution works, but let's work on code a little bit.
I recommend you to move these numbers to variables with appropriate names. For example, number 50 could be stored in a variable with name sevenDaysDiscount.

qoa11a (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,6 +28,17 @@
...
+  const threeDaysDiscount = 20;
+  const perDayRent = 40;
+
+  if (days >= 1 && days < 3) {
+    return days * perDayRent;
+  } else if (days >= 3 && days < 7) {
+    return days * perDayRent - threeDaysDiscount;
+  } else if (days >= 7) {
+    return days * perDayRent - sevenDaysDiscount;
+  }
```
polosanya, line 41: image
Let's change your code according to these rules so that your code looks cleaner

polosanya (changes requested):
Almost perfect!

qoa11a (approved):

PR #2099, opened 2022-12-07
Solution

AnnyKoval:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
```
liliia-livandovskaya, line 31: don't forget to remove comments

AnnyKoval: Done, thank you!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const discountLarge = 50;
+  const discountSmall = 20;
+
+  if (days >= 7) {
+    return ((days * price) - discountLarge);
+  }
+
+  if (days >= 3 && days <= 6) {
+    return ((days * price) - discountSmall);
+  } else {
```
witflash, line 41: You don't have any sense to use `else` block here. You can create general `return` after all `if` blocks.

```js
if (someCondition) {
  return some
}

if (anotheCondition) {
  return another
}

return default
```

witflash (changes requested):
Hey! Good job!
Take attention to the last comments.

src/calculateRentalCost.js
AnnyKoval: Thank you! Done ✓

witflash (approved):
Cool 👍

PR #2100, opened 2022-12-07
Solution

Ssviatt:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = days * 40;
+
+  if (days >= 7) {
+    sum -= 50;
+  } else if (days >= 3) {
+    sum -= 20;
```
tetiana-ivanchenko, line 36: move 30, 40 and 50 in the variables

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
stanislav-topikha: not fixed 
image

stanislav-topikha (changes requested):

stanislav-topikha (approved):

PR #2102, opened 2022-12-09
initial commit

ihorivna007:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  let charge;
+
+  if (days < 3) {
+    charge = 0;
+  } else if (days < 7) {
+    charge = 20;
+  } else {
+    charge = 50;
+  }
```
nazarmatsevych, line 39: Move all of this 'magic numbers' into variables.
Also in this case it's better to create two separate if blocks and combine conditions with logical operators where it's possible

nazarmatsevych (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    charge = smallCharge;
+  }
+
+  if (days >= 7) {
+    charge = bigCharge;
+  }
+
+  const cost = days * 40 - charge;
+
+  return cost;
```
SanyaBratashchuk, line 45:
```suggestion
  return days * 40 - charge;
```
And move 40 to the variable too

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #2103, opened 2022-12-09
Function calculateRentalCost

ilona-romaniuk:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
```
liliia-livandovskaya, line 31: don't forget to remove comments

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
+  let amount = days * 40;
+
+  if (days >= 3 && days < 7) {
+    amount -= 20;
+  } else if (days >= 7) {
+    amount -= 50;
```
liliia-livandovskaya, line 37: Great job! I recommend you move these numbers to variables with appropriate names. For example, the number 50 could be stored in a variable with the name sevenDaysDiscount.

liliia-livandovskaya (changes requested):
Pay attention to the comments

polosanya (changes requested):
Almost perfect! Let's polish it up)

loralevitska (approved):
perfect)

PR #2106, opened 2022-12-11
Solution

Kasapulka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  if (days > 0 && days < 3) {
+    result = days * 40;
```
loralevitska, line 34: Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content. For example `const DAILY_RENT_COST = 40`
```suggestion
    result = days * DAILY_RENT_COST;
```

loralevitska (changes requested):
almost done)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+  const dailyRent = 40;
+  const threeDaysDiscount = 20;
+  const sevenDaysDiscount = 50;
```
polosanya, line 34: image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+  if (days > 0 && days < 3) {
+    result = days * dailyRent;
+  }
+
+  if (days >= 3 && days < 7) {
+    result = days * dailyRent - threeDaysDiscount;
+  }
```
```suggestion
  if (days >= 3 && days < 7) {
    return days * dailyRent - threeDaysDiscount;
  }
```
Do you really need the first check and result variable?
You can just immediately return the needed amount after checking for discounts.
So if nothing passed the requirements for discount, just return the default amount.

polosanya (changes requested):
Almost perfect! Let's polish it up)

nazarmatsevych (approved):

PR #2107, opened 2022-12-11
Solution

MrFox32:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let res = 1;
+
+  res = days * 40;
```
TarasHoliuk, line 33:
```suggestion
  let res = days * 40;
```
why not like this?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 3 && days < 7) {
+    res -= 20;
```
TarasHoliuk, line 36: let's create variables for 20, 40, and 50 with meaningful names (should explain what it is)
because now it could be unclear

TarasHoliuk (changes requested):

vlad-demchuk (approved):

PR #2108, opened 2022-12-11
Solution

bkinklab:


src/calculateRentalCost.js
```diff
@@ -26,9 +26,19 @@
  *
  * @return {number}
  */
+const pricePerDay = 40;
+let sum = 0;
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    sum = (days * pricePerDay - 50);
```
VadimMentorQA, line 34: I recommend you to move these numbers to variables with appropriate names. For example, number 50 could be stored in a variable with name sevenDaysDiscount.

src/calculateRentalCost.js
```diff
@@ -26,9 +26,19 @@
  *
  * @return {number}
  */
+const pricePerDay = 40;
```
VadimMentorQA, line 29: Make sense to declare these variables within a function

VadimMentorQA (changes requested):

VadimMentorQA (approved):
Good job

PR #2109, opened 2022-12-11
calculation script made

Ch0p3r84:


qoa11a (changes requested):
Nice! 👍 
One small improvement can be made.

polosanya (approved):
Good job!

PR #2112, opened 2022-12-12
Solution

AnastasiiaEm:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
OlEzhA0, line 37: what is the magic numbers? please move them to constants

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
```
OlEzhA0, line 33: you may don't use `else if` statement when you return something

OlEzhA0:
```
if (condition) {
  return ...
}

if (another condition) {
  return ...
}

return ...
```

OlEzhA0 (changes requested):
Hey! please check my comments below

Virass (approved):

PR #2118, opened 2022-12-13
Solution

Levi199601:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days <= 2) {
+    return 80;
+  } else if (days <= 4) {
+    return 100;
+  } else if (days === 6) {
+    return 220;
+  } else if (days > 6) {
+    return 230;
+  }
```
Bohdan-Maksymiuk, line 39: hardcode values to pass tests - it is not about programming 
try to describe algorithms

Bohdan-Maksymiuk: avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):

Virass (approved):

PR #2130, opened 2022-12-15
Solution

AllaNebozhenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days < 3) {
+    result = days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    result = (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    result = (days * 40) - 50;
```
tetiana-ivanchenko, line 42: move 20, 40 and 50 in the variables with names that explains what this numbers actually means

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):

PR #2132, opened 2022-12-15
task completed

MykolaKaradzha:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const baseValue = 40;
+
+  if (days >= 7) {
+    return days * baseValue - 50;
+  } else if (days >= 3) {
```
OlEzhA0, line 35: you may not use `else if` construction if you return something right away
```
if (condition) {
  return ...
}

if (other condition) {
  return ...
}

return ...
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const baseValue = 40;
+
+  if (days >= 7) {
+    return days * baseValue - 50;
```
OlEzhA0, line 34: what is the `50` number? add constant for it

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days >= 3) {
+    return days * baseValue - 20;
```
OlEzhA0, line 36: the same as above

OlEzhA0 (changes requested):
GJ! Please check my comments

tetiana-ivanchenko (approved):

PR #2135, opened 2022-12-18
First task v_01

DenysKolbasin:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const rentCosts = 40;
+
+  if (days >= 7) {
+    return rentCosts * days - 50;
+  } else if (days >= 3) {
+    return rentCosts * days - 20;
+  } else {
+    return rentCosts * days;
+  }
```
liliia-livandovskaya, line 39: add 50 and 20 to the variables and check your code according to [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comment

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const luxDiscount = 50;
+  const basicDiscount = 20;
+
+  if (days >= 7) {
+    return rentCosts * days - luxDiscount;
+  } else if (days >= 3) {
+    return rentCosts * days - basicDiscount;
+  } else {
+    return rentCosts * days;
+  }
```
```suggestion
  } 
  
  if (days >= 3) {
    return rentCosts * days - basicDiscount;
  } 
  
  return rentCosts * days;
```

tetiana-ivanchenko (changes requested):
if you have return in you if block - using else after it is redundant, because you actually know, even if your condition is false - next block will works, because we know: after return, function will stops.

tetiana-ivanchenko (approved):

PR #2138, opened 2022-12-19
Solution

OlhaYevstifieieva:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
nzrii-kysel, line 32: could use a variable for `days * 40`

0xNazarii (changes requested):
checklist 3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const small = 20;
```
alexandra-protyanova, line 32: what is `small` and what is `big` it is not really good names for variables

alexandra-protyanova (changes requested):
Checklist point 3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
```
alexandra-protyanova, line 35: Checklist point 3 says that your code structure is too complicated, you should avoid it

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Good job 👏

PR #2140, opened 2022-12-20
Solution

anna-inozemets:
Function calculateRentalCost was realised

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
+
+  if (days < 3) {
+    sum = days * 40;
+  }
+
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return days * price;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * price) - mediumDiscount;
+  }
+
+  if (days >= 7) {
+    return (days * price) - bigDiscount;
+  }
```
Denchepornyuk, line 45: you can move check with days >= 3 under days >= 7 and then you can remove days < 7

and you don't need `days < 3`, you can return `days * price` if all previous checks were false

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #2144, opened 2022-12-22
Solution

Pallaci0:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const daypay = 40;
+  let allsum = 0;
+  const threeDays = 20;
+  const sevenDays = 50;
+
+  if (days >= 7) {
+    allsum += (days * daypay) - sevenDays;
```
```suggestion
    return (days * daypay) - sevenDays;
```

tetiana-ivanchenko: you don't need variable allSum, and count sum, just return

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return (days * daypay) - sevenDays;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * daypay) - threeDays;
+  }
+
+  if (days < 3) {
+    return days * daypay;
+  }
```
TarasHoliuk, line 45: Since it's the last and only possible option you can avoid `if` here (just FYI - for the future):

```suggestion
  return days * daypay;
```

TarasHoliuk (approved):

PR #2145, opened 2022-12-22
Solution

ihor-halkin:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
+  const totalCost = days * dayCost;
+
+  if (days >= 3 && days < 7) {
+    return totalCost - 20;
+  } else if (days >= 7) {
+    return totalCost - 50;
```
qoa11a, line 37: Let's move 20 and 50 to variables so their names would describe its content (smallDiscount and bigDiscount for example).

qoa11a (changes requested):

qoa11a (approved):

PR #2151, opened 2022-12-23
Develop: Finished calculateRentalCost task

sonyaxent:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    case (days < 3):
+      discount = 0;
+      break;
+    case (days >= 3 && days < 7):
+      discount = 20;
+      break;
+    case (days >= 7):
+      discount = 50;
+      break;
+  }
```
SanyaBratashchuk, line 43: You don't need swith(true) here use if

SanyaBratashchuk: And move each `magic` numbers to the variables

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const smallDiscountDays = 3;
+  const bigDiscountDays = 7;
+
+  if (days >= smallDiscountDays && days < bigDiscountDays) {
+    discount = 20;
+  } else if (days >= bigDiscountDays) {
+    discount = 50;
+  }
+
+  const totalPrice = (40 * days) - discount;
```
nazarmatsevych, line 41: also move "20", "50" and "40" to variables

nazarmatsevych (changes requested):

SanyaBratashchuk (approved):

PR #2153, opened 2022-12-24
Solution

oleh-holovnykh:


chermak-oleh:
Great job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const minDiscount = 20;
+  const maxDiscount = 50;
+  const maxLine = 7;
+  const minLine = 3;
+
+  if (days >= maxLine) {
+    return carCost * days - maxDiscount;
+  }
+
+  if (days < maxLine && days >= minLine) {
```
Denchepornyuk, line 41: you don't need `days < maxLine` check here, if `days >= maxLine` will be false then `days < maxLine` always will be true

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
oleh-holovnykh: oh, right

denys-cheporniuk (approved):

PR #2154, opened 2022-12-24
solution

Bir3791:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 3 && days < 7) {
+    discount = 20;
+  }
+
+  if (days >= 7) {
+    discount = 50;
+  }
+
+  totalCoast = (days * 40) - discount;
```
ViktorHavryliuk, line 42: Better move 40 to it's own const, like costPerDay or any name you consider suitable

ViktorHavryliuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const costPerDay = 40;
+  let totalCoast = 0;
+  let discount = 0;
+
+  if (days >= 3 && days < 7) {
+    discount = 20;
+  }
+
+  if (days >= 7) {
+    discount = 50;
```
SanyaBratashchuk, line 40: move 20 and 50 to the variables

SanyaBratashchuk (changes requested):

ViktorHavryliuk (approved):

PR #2156, opened 2022-12-24
Solution

dmytro-klepikov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
```
TarasHoliuk, line 36: do not use `else if` when you return value
use a few common `if` blocks (for the last condition you can return without `if` block because it's the last variant that is left)

also it would be nice to move numbers 40, 20, 50 to variables with meaningful names

TarasHoliuk (changes requested):

PR #2159, opened 2022-12-24
add solution

Anastasiia-Khudych:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalSum = days * 40;
+
+  if (days < 3) {
+    return totalSum;
+  }
+
+  if (days >= 3 && days < 7) {
```
nzrii-kysel, line 37: `days` will always be >= 3 at this point
```suggestion
  if (days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return totalSum - 20;
+  }
+
+  if (days >= 7) {
+    return totalSum - 50;
+  }
```
nzrii-kysel, line 43: redundant `if` statement
```suggestion
  return totalSum - 50;
```

0xNazarii (changes requested):
good thinking

src/calculateRentalCost.js
Anastasiia-Khudych: You are right, thanks for reviewing.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    return totalSum;
+  }
+
+  if (days < 7) {
+    return totalSum - 20;
+  }
+
+  return totalSum - 50;
```
alexandra-protyanova, line 41: move 50 & 20 to variables as well

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #2160, opened 2022-12-24
Solution

Rio1391:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    return sum - 20;
+  }
+
+  if (days >= 7) {
+    return sum - 50;
+  }
+
+  return sum;
```
SanyaBratashchuk, line 41: Move all `magic` numbers to the variables

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
Rio1391: done

tetiana-ivanchenko (approved):

PR #2161, opened 2022-12-25
Result

svitjojo:


0xNazarii (approved):
👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const rent = 40;
+  const discountFromThree = 20;
+  const discountFromSeven = 50;
+  let total = rent * days;
+
+  if (days >= 7) {
+    total -= discountFromSeven;
+  }
+
+  if (days >= 3 && days < 7) {
```
alexandra-protyanova, line 40:
```suggestion
  if (days >= 3) {
```

alexandra-protyanova (approved):
Good job 👍

src/calculateRentalCost.js
svitjojo: If days will be more than 7 function will count discount twice, it will pass both if statement, that's why I put && operator.

PR #2162, opened 2022-12-25
solution

imariash86:
hello:) my solution is here

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -29,15 +29,18 @@
 
 function calculateRentalCost(days) {
   // write code here
-  let totalRent = days * 40;
+  const discountFor3Days = 20;
+  const discountFor7Days = 50;
+  const priceForDay = 40;
+  let basicPrice = days * priceForDay;
 
   if (days >= 7) {
-    return (totalRent -= 50);
+    return (basicPrice -= discountFor7Days);
   } else if (days >= 3) {
-    return (totalRent -= 20);
+    return (basicPrice -= discountFor3Days);
   }
 
-  return totalRent;
+  return basicPrice;
 }
 
 module.exports = calculateRentalCost;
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let totalRent = days * 40;
```
anna-inozemets, line 32: I think it will be better to create variables for discount and basic price (Cheklist)

anna-inozemets (changes requested):

src/calculateRentalCost.js
imariash86: > I think it will be better to create variables for discount and basic price (Cheklist)

thanks for your review :)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const discountFor3Days = 20;
+  const discountFor7Days = 50;
+  const priceForDay = 40;
+  let basicPrice = days * priceForDay;
+
+  if (days >= 7) {
+    return (basicPrice -= discountFor7Days);
```
alexandra-protyanova, line 38:
```suggestion
    return basicPrice - discountFor7Days;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+  } else if (days >= 3) {
+    return (basicPrice -= discountFor3Days);
```
alexandra-protyanova, line 40:
```suggestion
    return basicPrice - discountFor3Days;
```

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
imariash86: sorry, I'll fix it

alexandra-protyanova (approved):
It isn't a good practice to use numbers in variables naming)

PR #2163, opened 2022-12-25
Solution

Kardash-Oleksandr:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
```
andrieieva-yuliia, line 31: You need use variables for the main values - checklist point 1
And don't use else after return statement - checklist point 3
🙂

andrieieva-yuliia (changes requested):
Nice work 👍  just small fixes according to the checklist

src/calculateRentalCost.js
Kardash-Oleksandr: fixed

alexandra-protyanova (changes requested):
Your code structure is really complicated, try to make it easier. Use 2 cases and third in main code.

Kardash-Oleksandr:
> Your code structure is really complicated, try to make it easier. Use 2 cases and third in main code.

fixed

alexandra-protyanova (approved):
Well done 👍

PR #2164, opened 2022-12-25
Solution

Avichez:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const defaultCost = days * 40;
+
+  if (days >= 7) {
+    return defaultCost - 50;
```
OlyaZaiets, line 34: It is better to create variables for discount

OlyaZaiets:
Great job!

src/calculateRentalCost.js
Denchepornyuk: agree
all these numbers should be in variables, keep your code without "magic" numbers, I don't know what numbers 40, 20, 50 mean, you must explain it to me in the code

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #2165, opened 2022-12-25
Solution

MTS13:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const carRentCosts = 40;
+
+  if (days <= 2) {
+    return carRentCosts * days;
+  } else if (days >= 7) {
+    return days * carRentCosts - 50;
```
ViktorHavryliuk, line 36: Move 50 and 20 to consts too
It's better to keep all your numbers in named variables/consts, then it's obvious what this number means

ViktorHavryliuk (changes requested):

ViktorHavryliuk (approved):

PR #2168, opened 2022-12-26
Solution

chermak-oleh:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = days * 40;
```
oleh-holovnykh, line 31: you need to define constants for discounts amounts
CL 1

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days > 6) {
```
oleh-holovnykh, line 33: also you can define constants for number of days after which the discount is applied

oleh-holovnykh (changes requested):
clear solution, well done! just need to define some more constants

src/calculateRentalCost.js
Denchepornyuk: agree

denys-cheporniuk (approved):

PR #2169, opened 2022-12-26
solution

yaroslavna-horban:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  }
+
+  if (days >= 3 && days < 7) {
```
vladyslav78292, line 35:
```suggestion
  if (days < 7) {
```

vladyslav78292: good job, just delete highlighted code

vladyslav78292 (changes requested):

src/calculateRentalCost.js
yaroslavna-horban: thank you!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  }
+
+  if (days < 7) {
+    return (40 * days) - 20;
+  }
+
+  return (40 * days) - 50;
```
Denchepornyuk, line 39: all these numbers should be in variables, keep your code without "magic" numbers, I don't know what numbers `40, 20, 50` mean, you must explain it to me in the code

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
yaroslavna-horban: got it, sir

denys-cheporniuk (approved):

PR #2170, opened 2022-12-26
Solution

vladyslav78292:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const sum = days * 40;
+
+  if (days >= 7) {
+    return sum - 50;
```
andrii-andriushchenko, line 34: I'd like to add variables for `discount,` that will be reusable in the future) but up for you)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  if (days >= 3) {
+    return sum - 20;
```
andrii-andriushchenko, line 38: same

andrii-andriushchenko (approved):
Well done 🔥

src/calculateRentalCost.js
vladyslav78292: yep, this way is better. thank you!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const sum = days * 40;
```
Denchepornyuk, line 31: 40 also should be in the variable

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #2171, opened 2022-12-26
Develop

ann-yurkevych:


0xNazarii (approved):
very clean solution👌
I would only suggest changing the name of `discountAmount`, its not really fitting

alexandra-protyanova (changes requested):
Add variables for 50 & 20 as well

alexandra-protyanova (approved):
Good job 👍

PR #2173, opened 2022-12-27
Solution

Oleksandr-Maly:


anna-inozemets (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  let totalCost = days * costPerDay;
+
+  if (days >= 3) {
+    totalCost = days * costPerDay - discount;
+  }
+
+  if (days >= 7) {
+    totalCost = days * costPerDay - discountBig;
+  }
```
alexandra-protyanova, line 43:
```suggestion
  if (days >= 3) {
    return totalCost - discount;
  }

  if (days >= 7) {
    return totalCost  - discountBig;
  }
```

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
Oleksandr-Maly: @alexandra-protyanova  it will not work correctly. If the days will be 7 or more it will return "totalCost - discount" but not "totalCost  - discountBig" as we need

alexandra-protyanova: you can replace `ifs` and it will work correct)

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #2177, opened 2022-12-27
Solution

Kravchenko71213:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPay = 40;
+  const sum = days * rentPay;
+
+  if (days < 3) {
+    return sum;
+  } else if (days >= 3 && days < 7) {
+    return sum - 20;
```
anna-inozemets, line 37: You should create variables for discount)

anna-inozemets (changes requested):
Pretty nice, just small fix according to checklist!)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPay = 40;
+  const sum = days * rentPay;
```
alexandra-protyanova, line 32:
```suggestion
  const rent = days * rentPay;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const mediumDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return sum;
+  } else if (days >= 3 && days < 7) {
```
alexandra-protyanova, line 38:
```suggestion
  } else if (days < 7) {
```

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #2181, opened 2022-12-27
solution

Anton-Zhytonbaiev:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
```
anna-inozemets, line 31: You should use variables. (checklist)
I guess, it will be nice to create variables for 20, 40, 50

anna-inozemets (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #2182, opened 2022-12-27
Solution

victorionov:


sv-dubov (changes requested):
Use variables especially for repeated values

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
-  // write code here
+  let totalPayment = 0;
+
+  const dailyPrice = 40;
+  const maxDiscount = 50;
+  const minDiscount = 20;
+
+  if (days < 3) {
+    totalPayment = dailyPrice * days;
+  }
```
Denchepornyuk, line 39: I believe you can remove this `if` and return `dailyPrice * days` at the end of the function

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+  if (days >= 7) {
+    totalPayment = (dailyPrice * days) - maxDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
```
Denchepornyuk, line 45: you don't need `days < 7` check here, if `days >= 7` will be false then `days < 7` always will be true

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #2183, opened 2022-12-27
Solution

Alessandranitti15:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let price = 40 * days;
```
polosanya, line 31: What is 40? Let's create a variable for that value to make your code more understandable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3) {
+    price = (40 * days) - 20;
+  }
```
polosanya, line 35:
```suggestion
  if (days >= 3) {
    return price - 20;
  }
```
Don't duplicate yourself. Also, you can just immediately return the price with discount

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    price = (40 * days) - 50;
```
polosanya, line 38: Same here. Btw, would be great to create variables for `20` and `50` too.

polosanya (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
```
polosanya, line 31: Let's specify that it's a price for 1 day.
Something like `dayPrice`/`priceForDay` would be okay.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const discount3 = 20;
+  const discount7 = 50;
```
polosanya, line 33: image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 7) {
+    return finalPrice - discount7;
+  }
+
+  if (days >= 3) {
+    return finalPrice - discount3;
+  } else {
+    return finalPrice;
+  }
```
polosanya, line 44:
```suggestion
  if (days >= 3) {
    return finalPrice - discount3;
  }
  
  return finalPrice;
```
No need for else here. You code will not go further after `return` anyway

polosanya (changes requested):
Let's polish it up)

polosanya (changes requested):
Hi, I see you haven't fixed previous comments, maybe you did not push your changes? 🙂 
image
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const basicDiscount = 20;
+  const halfDiscount = 50;
```
qoa11a, line 33: What does halfDiscount mean?

qoa11a (changes requested):

src/calculateRentalCost.js
Alessandranitti15: I got confused, I thought it was 50%.

polosanya (approved):
Well done!

PR #2185, opened 2022-12-27
Solution

popeye88:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
```
polosanya, line 31: What is 40? Let's create a variable for this value to make your code more understandable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    total = total - 20;
+
+    return total;
+  } else if (days >= 7) {
+    total = total - 50;
+  }
```
```suggestion
  if (days >= 3 && days < 7) {
    return total - 20;
  }
  
  if (days >= 7) {
    return total - 50;
  }
```
Do you really need to modify total? Just make it a const and return value with discount.
Also, there is no need for `else if` as your code will not go further after `return` anyway.
P.S. would be great to create variables for `20` and `50`.

polosanya (changes requested):

src/calculateRentalCost.js
loralevitska: not fixed. Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content. 
`lowDiscount` - good
`highDiscount` - good
`????😢 ` - means 40 create variable for this number

loralevitska (changes requested):
PLease pay attention to comment which I left

popeye88:
Understood an idea. TY

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const highDiscount = 50;
+  const totalToPay = days * rentalCarCost;
+
+  if (days >= 3 && days < 7) {
+    return totalToPay - lowDiscount;
+  }
+
+  if (days >= 7) {
+    return totalToPay - highDiscount;
+  }
```
SanyaBratashchuk, line 42:
```suggestion
   if (days >= 7) {
    return totalToPay - highDiscount;
  }
  
  if (days >= 3) {
    return totalToPay - lowDiscount;
  }
```

SanyaBratashchuk (approved):

PR #2186, opened 2022-12-27
Implemented calculateRentalCost function

GarryRocksman:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPrice = 40;
+  const discountAfter3 = 20;
```
sv-dubov, line 32: ChL, p 5

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const discountAfter7 = 50;
```
sv-dubov, line 33: ChL, p 5

sv-dubov (changes requested):
Good job! But need some changes :)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const maxDiscount = 50;
+
+  switch (true) {
+    case days >= 3 && days < 7:
+      return days * rentPrice - minDiscount;
+    case days >= 7:
+      return days * rentPrice - maxDiscount;
+    default :
+      return days * rentPrice;
+  }
```
Denchepornyuk, line 42: don't use switch case with `true` it's a wrong way how to use switch case, in this case it's better to use a few `if`

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
GarryRocksman: fixed

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return days * rentPrice - minDiscount;
+  } else if (days >= 7) {
+    return days * rentPrice - maxDiscount;
+  } else {
+    return days * rentPrice;
+  }
```
Denchepornyuk, line 41: you can make code simpler with two `ifs` and without `days < 7`, make check `days >= 3` after `days >= 7`

denys-cheporniuk (changes requested):

src/calculateRentalCost.js
GarryRocksman: fixed

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const maxDiscount = 50;
+  let resultPrice = rentPrice;
+
+  if (days >= 3) {
+    resultPrice = rentPrice - minDiscount;
+  }
+
+  if (days >= 7) {
+    resultPrice = rentPrice - maxDiscount;
+  }
```
Denchepornyuk, line 42: will be better to compare numbers with `<=`

denys-cheporniuk (approved):

PR #2189, opened 2022-12-27
Solution

Dashaverz:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let price = 0;
+
+  if (days >= 7) {
+    price = days * 40 - 50;
```
ann-yurkevych, line 35: you should add variables for 50 & 20
`  const smallDiscount = 20;`
`  const bigDiscount = 50`

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  } else if (days >= 3) {
+    price = days * 40 - 20;
+  } else {
```
ann-yurkevych, line 38: I don't know if this condition is appropriate here or no, maybe it's better to write just 2 if conditions without else if

ann-yurkevych (changes requested):

src/calculateRentalCost.js
Dashaverz: thanks! fixed all mistakes

alexandra-protyanova (approved):
Well done 👍

PR #2194, opened 2022-12-27
Solution

max-rozzhalovets:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const total = 40 * days;
+
+  if (days >= 7) {
+    return total - 50;
```
OlyaZaiets, line 35: You should create variables for discount

alexandra-protyanova: agree

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #2197, opened 2022-12-27
solution

artem-sobolevskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentSum = 40 * days;
```
ann-yurkevych, line 31: in this scope you can add variables for 50 and 20
` const bigDiscount = 50;`
`const smallDiscount = 20;`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+
+  switch (true) {
```
ann-yurkevych, line 33: it's not a good idea to use switch case with the condition of 'true'. It's better to use a few if

ann-yurkevych (changes requested):

src/calculateRentalCost.js
alexandra-protyanova: If you use switch with `true` condition, you do not need `switch` at all)

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #2201, opened 2022-12-28
solution

Peti51:


qoa11a (changes requested):
Good job!
Let's refactor code a little bit.

1. Move numbers (40, 50, 20, 3, 7) to variables so it can be understood what they represent.
2. There's no need to have three 'if' blocks. You can just check if days is more than or equals to 7, as well as check if days is more than or equals to 3. It'd look like this:
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+  const sevenDayLimit = 7;
+  const fiftyDollarOffer = 50;
+  const twentyDollarOffer = 20;
+  const threeDayLimit = 3;
```
polosanya, line 35: Let's don't hardcore the variable's name for exact amounts.
Just use abstract names like basicDiscount, bigDiscount etc. 
image

polosanya (changes requested):
Almost perfect!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+  const secondDaylimit = 7;
+  const firstDiscount = 50;
+  const secondDiscount = 20;
+  const firstDaylimit = 3;
```
qoa11a, line 35: First and second describe nothing. Try to give more clear names (for example, basicDiscount, smallDiscountDays)

qoa11a (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+  const seconddiscountDay = 7;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const firstDiscountDay = 3;
```
polosanya, line 35:
```suggestion
  const daysForBigDiscount= 7;
  const bigDiscount = 50;
  const smallDiscount = 20;
  const daysForSmallDiscount = 3;
```
I think the previous comment was about words `first` and `second`, not just first and second variables))
Because it is not clear what `firstDiscount` and `secondDiscount` mean. It's better to specify what's the difference between them

polosanya (changes requested):
One little fix

qoa11a (approved):

PR #2202, opened 2022-12-28
solved task

Serghii29:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = 40;
+
```
ann-yurkevych, line 32: here you can add variables for 20 & 50 days, also it's better to create variable for storing data of final amount: 
const total = days * 40;

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+    return totalCost;
+  }
+
+  if (days < 7) {
+    totalCost = totalCost * days - 20;
+
+    return totalCost;
+  }
+
+  if (days >= 7) {
```
ann-yurkevych, line 45: you can create only 2 if conditions, (days >= 3) and (days >= 7), but it's not so necessary)

ann-yurkevych (changes requested):
You've made a good job) Excellent code, but you need to fix some parts of your code)

src/calculateRentalCost.js
alexandra-protyanova: agree

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    totalCost -= 50;
+
+    return totalCost;
+  }
+
+  if (days >= 3) {
+    totalCost -= 20;
+
+    return totalCost;
+  }
```
alexandra-protyanova, line 43:
```suggestion
  if (days >= 7) {
    return totalCost - 50;
  }

  if (days >= 3) {
    return totalCost - 20;
  }
```

alexandra-protyanova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = 40 * days;
+
+  if (days >= 7) {
+    return totalCost - 50;
+  }
+
+  if (days >= 3) {
+    return totalCost - 20;
```
alexandra-protyanova, line 38: move 20 & 50 to variables as well

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #2203, opened 2022-12-28
Develop

metalikaqq:


sv-dubov (changes requested):
Almost good. But needs some improves:
1. You have to create variables for rent, total rent and discounts (ChL p.1).
2. You don't need 3 checks with if here. Use 2 if and the last one just return.

alexandra-protyanova (approved):
Well done 👍

PR #2206, opened 2022-12-29
solution

bohdanmayor:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return (days * 40) - 20;
+  } else {
+    return (days * 40) - 50;
+  }
```
OlEzhA0, line 37: you might not use `else if` statement with return

```
if (condition) {
  return something
}

if (condition) {
  return something
}

return something
```

OlEzhA0, line 37: what is number `40`, `50` and `20` 
add constants

OlEzhA0 (changes requested):
GJ!

etojeDenys (approved):
lgtm

PR #2208, opened 2022-12-29
Solution

VadymTsyndra:


qoa11a (changes requested):
Let's refactor our code a little bit:

1. You don't need to give zero value to result right away. 
2. You can just return result right away inside if statements, then there will be no need in else..if.
3. Let's move number to variables with clear names so it's understandable what they represent.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+
+  if (days >= 7) {
+    return pricePerDay * days - 50;
```
ViktorHavryliuk, line 34: Move discounts to consts too, so that it's clear what this number means

ViktorHavryliuk (changes requested):

nazarmatsevych (approved):

PR #2214, opened 2023-01-03
the task was solved

batiukova-nataliia:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
```
Viktor-Kost, line 31: We can return value that we need without creating an extra variable
```suggestion

```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    total = days * 40;
+  } else if (days >= 3 && days <= 6) {
+    total = days * 40 - 20;
+  } else {
+    total = days * 40 - 50;
+  }
+
+  return total;
```
Viktor-Kost, line 41:
```suggestion
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let total = 0;
+
+  if (days < 3) {
+    total = days * 40;
+  } else if (days >= 3 && days <= 6) {
+    total = days * 40 - 20;
+  } else {
+    total = days * 40 - 50;
+  }
```
Viktor-Kost, line 39: We can use an early return

Ex: 
```
if (a === b) {
  return profit * percent
}

if (b === c) {
  return profit * percent * 20
}
```

Viktor-Kost (changes requested):

tetiana-ivanchenko (changes requested):
move 20, 40 and 50 in the variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const discountWeek = 50;
+
+  if (days < 3) {
+    return days * dayPrice;
+  } else if (days >= 3 && days <= 6) {
+    return days * dayPrice - discountThreeDays;
+  } else {
+    return days * dayPrice - discountWeek;
+  }
 }
```
marina-tilniak, line 42:
```suggestion
  if (days < 3) {
    return days * dayPrice;
  }
  
  if (days >= 3 && days <= 6) {
    return days * dayPrice - discountThreeDays;
  }
    
  return days * dayPrice - discountWeek;
}
```

marina-tilniak: you don't need else when you have return inside if

marina-tilniak (approved):

PR #2218, opened 2023-01-04
Solution

kateryna-babina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,8 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return (days >= 7) ? (days * 40) - 50
+    : (days >= 3) ? (days * 40) - 20 : days * 40;
```
kirill-vypirovskyi, line 32: It works, but it's harder to read and understand. In my opinion it will be better to write it using if.

kirill-vypirovskyi (approved):

src/calculateRentalCost.js
illiarerun: agree with Kirill, redo it with `if's`, nested ternaries are evil

illiarerun, line 32: [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 1st item, I don't know what 40, 50, 20, 7 and 3 are

illiarerun, line 32: you repeat `days * 40` calculation 3 times, create a variable for it

illiarerun (changes requested):

illiarerun (approved):

PR #2219, opened 2023-01-04
Solution

Zelinskyi-Serhii:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  }
+
+  if (days < 7) {
+    return days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
 }
```
Bogdan-Mykhailov, line 44:
```
 if (days < 3) {
    return days * 40;
  } else if (days < 7) {
    return days * 40 - 20;
  }
  return days * 40 - 50;
```

Bogdan-Mykhailov (approved):
Good job, this solution is easy to understand, and you can make shorter)

src/calculateRentalCost.js
Andr1yK: @Bogdan-Mykhailov why do we need to use `else if` instead of just `if`?

Andr1yK (changes requested):
What is 20, 50?
Avoid magic numbers

src/calculateRentalCost.js
Bogdan-Mykhailov: @Andr1yK if-else statements stop doing comparisons as soon as it finds one that's true.
if-if does every comparison.

Andr1yK: > if-if does every comparison

Yes, but not after return :)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const normalPrice = 40;
+  const discountAfter3days = 20;
+  const discountAfter7Days = 50;
```
Andr1yK, line 34: point 5 - https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return days * normalPrice - discountAfter3days;
+  }
+
+  if (days >= 7) {
+    return days * normalPrice - discountAfter7Days;
+  }
```
Andr1yK, line 46:
```suggestion
  return days * normalPrice - discountAfter7Days;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
```
Andr1yK, line 31: do u need this comment?

Andr1yK (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return days * normalPrice - smallDiscount;
+  }
+
+  return days * normalPrice - bigDiscount;
```
Andr1yK, line 43: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

Save same parts to const

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const normalPrice = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return days * 40;
```
Andr1yK, line 36: hmm.. mb 40 = normalPrice
```suggestion
    return days * normalPrice;
```

Andr1yK (approved):

PR #2220, opened 2023-01-04
Finish task

U-Ra-All:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
   // write code here
+  const amount = days * 40;
+
+  if (days < 3) {
+    return amount;
+  } else if (days < 7) {
+    return amount - 20;
+  } else {
+    return amount - 50;
+  }
```
Viktor-Kost, line 40: try not use else if nesting. Just use early return
```suggestion
  if (days < 3) {
    return amount;
  }
  
  if (days < 7) {
    return amount - 20;
  } 
  
   return amount - 50;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const amount = days * 40;
```
Viktor-Kost, line 32: it is not obvious naming for variable, cause it does not show amount of what

Viktor-Kost (changes requested):

src/calculateRentalCost.js
U-Ra-All: Fixed

Viktor-Kost (approved):
Nice!

PR #2222, opened 2023-01-05
solution

Spec0s:


jstmpelowycz (approved):
let's go through the checklist, you've missed many points

jstmpelowycz (changes requested):
sorry, misclicked buttons

Spec0s:
i think i got it , updated

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const totalCost = days * 40;
```
jstmpelowycz, line 32: what are 40, 50, 20? fix it according to the checklist, 5th point

jstmpelowycz (changes requested):

src/calculateRentalCost.js
Spec0s: "fix it according to the checklist, 5th point" i did not understand where t look at ? i was looking here 
https://github.com/mate-academy/style-guides/blob/master/javascript.md#23-naming-conventions

is this right place ? but at 5th point is 5. Destructuring ?

jstmpelowycz (approved):
all good now, congrats 🥳

src/calculateRentalCost.js
jstmpelowycz: https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md

Spec0s: oh man feel so stupid now, thank you!  yes i see the file now 😂😂

PR #2225, opened 2023-01-06
solution

muminDev:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < 7 && days > 2) {
+    result = (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    result = (days * 40) - 50;
+  }
+
+  return result;
```
qoa11a, line 45: Let's move numbers to variables so it's understandable what they represent. Also, you can return result right away when you go in if condition, as if days are less than 3 there's no need to go further and check other conditions.

qoa11a (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
polosanya, line 31: What is result? It's better to specify what value is it

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  const dailyPayment = 40;
+  const discountTwenty = 20;
+  const discountfifty = 50;
```
polosanya, line 34: image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    result = days * dailyPayment;
+  }
+
+  if (days < 7 && days > 2) {
+    result = (days * dailyPayment) - discountTwenty;
+  }
+
+  if (days >= 7) {
+    result = (days * dailyPayment) - discountfifty;
+  }
```
polosanya, line 46: Do you really need to save it in result variable? Can't you just return the value immediately?
You can check for bigDiscount, then for basicDiscount and if none of checks was passed, then just return default price

polosanya (changes requested):
Let's improve your code a bit

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyPayment = 40;
+  const discountBasic = 20;
+  const discountBig = 50;
+
+  if (days < 7 && days > 2) {
+    return (days * dailyPayment) - discountBasic;
```
```suggestion
    return rentalPrice - discountBasic;
```
Would be great to create a variable for that value in order not to duplicate yourself and make your code cleaner

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

PR #2228, opened 2023-01-09
solution

d-beschastnyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
```
DarkMistyRoom, line 45: you can do the same only with 2 `if`s

DarkMistyRoom (changes requested):
make variables for all 'magic' numbers

agniya-i (changes requested):

Anastasiia-Svintsova (approved):

PR #2230, opened 2023-01-09
solution

EmilSukiennik:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = days * 40;
+
+  if (days > 6) {
+    cost -= 50;
+  } else if (days > 2) {
+    cost -= 20;
```
```suggestion
  if (days >= 7) {
    return cost - 50;
  }
  
  if (days >= 3) {
    return cost - 20;
```
It's better to immediately return the price to avoid using else if. 
Also, I think it's better to use 3 and 7 as it's written in task description in order not to confuse other developers

polosanya (changes requested):
Almost perfect! Let's improve your code a bit! 
Please pay attention to this requirement:
image
Let's create variables for all main values (e.g. 40, 3, 7, 50. 20) to make your code more understandable and descriptive

src/calculateRentalCost.js
44shaggy: OK, that makes sense.

polosanya (approved):
Perfect! 🔥

PR #2232, opened 2023-01-10
Solution

dianaavem:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const price = 40;
+  let sum;
+
+  if (days >= 7) {
+    sum = (days * price) - 50;
+  } else if (days >= 3 && days < 7) {
+    sum = (days * price) - 20;
+  } else if (days < 3) {
+    sum = days * price;
+  }
```
SanyaBratashchuk, line 40: You can do it with 2 plain `if`s.

SanyaBratashchuk, line 40: move all `magic` numbers to the variables

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  let sum;
+
+  if (days >= 7) {
+    sum = (days * price) - bigDiscount;
+  } else if (days >= 3 && days < 7) {
```
DarkMistyRoom, line 38: 3 and 7 are also `magic`)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+    sum = (days * price) - smallDiscount;
+  } else {
+    sum = days * price;
+  }
```
DarkMistyRoom, line 42: You can do it without `else`s at all. Try to avoid them

DarkMistyRoom (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days >= bigDiscountDays) {
+    sum = (days * price) - bigDiscount;
+  } else if (days >= smallDiscountDays && days < bigDiscountDays) {
+    sum = (days * price) - smallDiscount;
+  } else {
+    sum = days * price;
+  }
+
+  return sum;
```
SanyaBratashchuk, line 46:
```suggestion
  if (days >= bigDiscountDays) {
    return days * price - bigDiscount;
  } 
  
  if (days >= smallDiscountDays) {
    return days * price - smallDiscount;
  } 
  
    return days * price;
```

SanyaBratashchuk: and move days * price to the variable

SanyaBratashchuk (changes requested):

tetiana-ivanchenko (approved):

PR #2233, opened 2023-01-11
Solution

yevgenbarannikov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const i = days;
```
tetiana-ivanchenko, line 31: whu do you nee this variable? (and also if you create variables - use more meanfull names)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  let totalCoast;
+
+  if (days >= 7) {
+    totalCoast = i * 40 - 50;
+  } else if (days >= 3) {
+    totalCoast = i * 40 - 20;
+  } else {
+    totalCoast = i * 40;
```
tetiana-ivanchenko, line 39: move 20, 40 and 50 in the variables

tetiana-ivanchenko (changes requested):

PR #2234, opened 2023-01-11
Solution

Volodymyr-Zahreba:


ghost (changes requested):
1. Use variables for the main values so that you can reuse them and give them descriptive names.
2. Avoid reassigning variables values. It's better to use const instead of let.
3. Rewrite your code so it returns the value immediately in if statements.

Read checklist.md again)

PR #2237, opened 2023-01-13
Solution

fisherok3452:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = 0;
```
tetiana-ivanchenko, line 31: 1. move 20, 40 and 50 in the variables
2. remove totalAmount and in if block return value instead of saving in total amount
3. in that case you don't need `else`

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
```
tetiana-ivanchenko, line 30:
```suggestion
function calculateRentalCost(days) {
```

tetiana-ivanchenko (approved):

PR #2240, opened 2023-01-13
Solution

NazariiAlieksieiev:
Or better way to use `if`?


IvanFesenko (changes requested):
If statement will be better.
No need in total variable, you can return result of calculation
`return days * price`

tetiana-ivanchenko (changes requested):

mykhalenych (approved):

PR #2241, opened 2023-01-13
Solution

OlyaNikora:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const numberOfDays = days;
```
Anastasiia-Svintsova, line 31: I don't think that you need to rewrite days to a new variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (numberOfDays < 3) {
+    return 40 * numberOfDays;
+  }
+
+  if (numberOfDays >= 3 && numberOfDays < 7) {
+    return 40 * numberOfDays - 20;
+  }
+
+  return 40 * numberOfDays - 50;
```
Anastasiia-Svintsova, line 41: Would be great to create variables for 40, 20 and 50

Anastasiia-Svintsova (changes requested):

tetiana-ivanchenko (approved):

PR #2243, opened 2023-01-15

Mykola-Borodavka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  let price = cost * days;
```
polosanya, line 32: What exactly cost and price is it? Let's specify it.
Also, can't you make it a const and just return price - discount?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days >= 3 && days < 7) {
+    price -= 20;
+  }
+
+  if (days >= 7) {
+    price -= 50;
+  }
```
polosanya, line 40: Would be great to create variables for 3, 7, 20 and 50 to make your code more descriptive

polosanya (changes requested):
Let's polish it up

src/calculateRentalCost.js
Mykola-Borodavka: Done. Changed the logic a bit. First the function checks whether days are more that 7 and then if days are more than 3.
Thank you!

Anastasiia-Svintsova (approved):
👍

PR #2245, opened 2023-01-16
Solution

nadia-hope:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = 0;
+  const totalCost = days * 40;
```
polosanya, line 32: What is 40? Let's create a variable for that value

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  switch (true) {
+    case (days >= 7):
+      cost = totalCost - 50;
+      break;
+    case (days >= 3):
+      cost = totalCost - 20;
+      break;
+    default:
+      cost = totalCost;
```
polosanya, line 42: Switching true is not the best idea. 
You can just check if days are enough for some discount and immediately return the full price.
Would be great to create variables for 3, 7, 20 and 50 to make your code more descriptive.
Also, do you really need `cost` variable?

polosanya (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPrice = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const totalCost = days * rentPrice;
+  const threeDays = 3;
+  const sevenDays = 7;
```
```suggestion
  const daysForSmallDiscount = 3;
  const daysForBigDiscount = 7;
```
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= sevenDays) {
+    return totalCost - bigDiscount;
+  } else if (days >= threeDays) {
+    return totalCost - smallDiscount;
+  } else {
+    return totalCost;
+  }
```
polosanya, line 44:
```suggestion
  if (days >= sevenDays) {
    return totalCost - bigDiscount;
  }
  
  if (days >= threeDays) {
    return totalCost - smallDiscount;
  }
  
  return totalCost;
```
Do you really need to use if...else here?
You code will not go further after `return` anyway

polosanya (changes requested):
Let's polish it up)

Anastasiia-Svintsova (approved):
👍

PR #2246, opened 2023-01-16
second commit

NikolayGit1:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  money = days * 40;
+
+  if (days > 6) {
+    money = days * 40 - 50;
+  } else if (days > 2) {
+    money = days * 40 - 20;
+  }
+
+  return money;
```
```suggestion

  if (days > 6) {
    return days * 40 - 50;
  } 
  
  if (days > 2) {
    return days * 40 - 20;
  }

  return days * 40;
```

tetiana-ivanchenko (changes requested):
also move 20, 40 and 50 in the variables

nazarmatsevych (changes requested):
You didnt move 20, 40 and 50 to the variables.
Please pay more attention to the comments before making re-request

nazarmatsevych (approved):

PR #2247, opened 2023-01-17
calculate_rental_cost_created_js

Lazarend:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    return (days * 40) - 50;
+  }
```
```suggestion
  if (days < 3) {
    return days * 40;
  }

  if (days >= 3 && days < 7) {
    return (days * 40) - 20;
  }
  
  return (days * 40) - 50;
 
```

tetiana-ivanchenko: also move 20, 40 and 50 in the variables

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
Lazarend: Hi, Tatiana.
Ok, understood and corrected the error)

ViktorHavryliuk (changes requested):
This is not implemented:
image

20, 40 and 50 should be in their own consts with descriptive names, so that it's clear what those numbers mean

Lazarend:
> This is not implemented: image
> 
> 20, 40 and 50 should be in their own consts with descriptive names, so that it's clear what those numbers mean

Hi. Ok, described.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalSum = 40 * days;
+  const moreThanThreeDays = totalSum - 20;
+  const moreThanSevenDays = totalSum - 50;
```
SanyaBratashchuk, line 33: move 20, 40, 50 to the variables

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
Lazarend: ok. I did it)

ViktorHavryliuk (approved):

src/calculateRentalCost.js
Lazarend: Hi. Please check. If everything is ok, then confirm)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicPrice = 40;
+  const firstDiscount = 20;
+  const secondDiscount = 50;
```
OleziO, line 33: Constants must be in upper case and snake case.

Example: `BASIC_PRICE` etc.

OleziO (approved):

src/calculateRentalCost.js
Lazarend: Hi, I did it)

PR #2250, opened 2023-01-17
Solution

andriy-litkovets:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const discont36 = 20;
+  const discont7 = 50;
+  const rent = 40;
+  let result = 0;
+
+  if (days >= 7) {
+    result = rent * days - discont7;
```
Anastasiia-Svintsova, line 37: Just return the value everywhere
```suggestion
    return rent * days - discont7;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  } else if (days >= 3) {
+    result = rent * days - discont36;
+  } else {
+    result = rent * days;
+  }
```
Anastasiia-Svintsova, line 42: And move rent * days to a variable
```suggestion
  } 
    return rent * days;
```

Anastasiia-Svintsova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discont36 = 20;
+  const discont7 = 50;
```
polosanya, line 32:
```suggestion
  const smallDiscount = 20;
  const bigDiscount = 50;
```
What is 36? Don't hardcore variable's name for some value. It may change in future

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const rentPerDay = 40 * days;
```
polosanya, line 33: rentPerDay is 40. This value is the rent price for all days.
Create a variable for 40

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return rentPerDay - discont7;
+  } else if (days >= 3) {
+    return rentPerDay - discont36;
+  }
```
```suggestion
  if (days >= 7) {
    return rentPerDay - discont7;
  }
  
  if (days >= 3) {
    return rentPerDay - discont36;
  }
```
Why do you need else if?
Your code will not go further after return anyway

polosanya (changes requested):
Almost done

Anastasiia-Svintsova (approved):
👍

PR #2253, opened 2023-01-19
Develop

Yevhen-Ivanchuk:


qoa11a (changes requested):
few suggestions:
1. move numbers to variables (like 20, 50)
2. you won't need else if if you return result right away. Do it that way please

src/calculateRentalCost.js
```diff
@@ -26,9 +26,19 @@
...
-}
+  const rental = 40;
+  const discountSmall = 20;
+  const discountBig = 50;
+  let price = days * rental;
+
+  if (days >= 3 && days < 7) {
+    price = (days * rental) - discountSmall;
+  } else if (days >= 7) {
+    price = (days * rental) - discountBig;
```
Anastasiia-Svintsova, line 38: You already have a variable for `days * rental`
```suggestion
    return price - discountSmall;
  } else if (days >= 7) {
    return price - discountBig;
```

src/calculateRentalCost.js
```diff
@@ -26,9 +26,19 @@
...
  * @return {number}
  */
-
 function calculateRentalCost(days) {
-  // write code here
-}
+  const rental = 40;
+  const discountSmall = 20;
+  const discountBig = 50;
+  let price = days * rental;
```
Anastasiia-Svintsova, line 33:
```suggestion
  const price = days * rental;
```

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):

PR #2256, opened 2023-01-20
Solution

Bohdan33:


ViktorHavryliuk (changes requested):
Move 20, 40 and 50 to consts with descriptive names, so that it's clear what those numbers mean

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  const discount3days = 20;
+  const discount7days = 50;
+
+  if (days >= 0 && days <= 2) {
+    return days * price;
+  } else if (days >= 3 && days < 7) {
+    return days * price - discount3days;
+  } else {
+    return days * price - discount7days;
+  }
```
DarkMistyRoom, line 42: will be better to read if you refactor it with plane `if`s

DarkMistyRoom (changes requested):

SanyaBratashchuk (approved):

PR #2257, opened 2023-01-20
done

danulyk05:


ViktorHavryliuk (changes requested):
Move 20, 40, and 50 to constants with descriptive names, so that it's clear what those numbers represent

ViktorHavryliuk (approved):
Good, but you may omit using specialOff variable

PR #2259, opened 2023-01-20
done

ghost:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const clearCost = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return clearCost - 20;
+  } else if (days >= 7) {
+    return clearCost - 50;
+  }
```
ViktorHavryliuk, line 37: Try to use else if as little as possible, it may get out of hands quickly and make your code harder to follow
```suggestion
  if (days >= 7) {
    return clearCost - 50;
  }
  
  if (days >= 3) {
    return clearCost - 20;
  }
```

ViktorHavryliuk (changes requested):
Move 20, 40 and 50 to consts with descriptive names, so that it’s clear what those numbers mean

nazarmatsevych (approved):

PR #2260, opened 2023-01-22
The function was created

YevheniiHura:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  }
```
Virass, line 37: Don't use this magic numbers. Move to variables.

Virass (changes requested):
Nice work, but correct the remark in the comment

YevheniiHura:
> Nice work, but correct the remark in the comment

src/calculateRentalCost.js
YevheniiHura: Is it ok? Or should i move the amount of days to variables too?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return days * rent;
+  } else if (days >= 3 && days < 7) {
+    return days * rent - smallDiscount;
+  } else if (days >= 7) {
+    return days * rent - bigDiscount;
+  }
```
IvanFesenko, line 41: 2023-01-24_17h56_25
No else statment needs

IvanFesenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return days * rent;
```
Anastasiia-Svintsova, line 36: Would be great to move `days * rent` to the variable as well

Anastasiia-Svintsova (approved):

src/calculateRentalCost.js
YevheniiHura: done

PR #2261, opened 2023-01-22
Develop

Vitalii-Maistrenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
+    return days * 40 - 20;
+  } else if (days <= 3) {
+    return days * 40;
```
polosanya, line 36: Do you need else...if?
Your code won't go further after return anyway.
Also, let's create descriptive variables for all the main values to make your code cleaner and more understandable

polosanya (changes requested):
Let's improve your code a bit

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+
+  const costBorday = 40
+  const retdDiscont1 = 20;
+  const retdDiscont2 = 50
```
OleziO, line 34: Use an array for data of the same type

OleziO (approved):

PR #2262, opened 2023-01-22
Develop

Mar4enkofff:


ViktorHavryliuk (changes requested):
Move other numbers to consts with descriptive name, so that it is clear what those numbers mean.
Also, try to find a way to solve it without using "else if". Feel free to ask in chat if needed

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    return price;
+  }
+
+  if (days >= 3 && days < 7) {
+    return price - smallDiscount;
+  }
+
+  if (days >= 7) {
+    return price - mediumDiscount;
+  };
```
DarkMistyRoom, line 46: 3 and 7 also looks like magic numbers)

DarkMistyRoom (approved):

PR #2263, opened 2023-01-22
Solution

oksom:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayAmount = 40;
+  let totalAmount = 0;
+  const baseDiscount = 20;
+  const addDiscount = 50;
+
+  if (days < 3) {
+    totalAmount = days * dayAmount;
```
Anastasiia-Svintsova, line 37: Move `days * dayAmoun` to variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayAmount = 40;
+  let totalAmount = 0;
```
Anastasiia-Svintsova, line 32: You don't need this variable, just return the result of the calculation each  time

src/calculateRentalCost.js
Anastasiia-Svintsova, line 37:
```suggestion
    return days * dayAmount;
```

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):
👍

PR #2265, opened 2023-01-23
Solution

Vasyl-Pavlenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const amount = days * dayPrice;
+  let totalAmount = 0;
+
+  if (days >= 3 && days < 7) {
+    totalAmount = amount - smallDiscount;
```
OlEzhA0, line 38: you can return value right here
```suggestion
    return amount - smallDiscount;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  } else if (days >= 7) {
```
OlEzhA0, line 39: and if you return value at once, use this syntax

```
if (cond1) {
  return a;
}

if (cond2) {
  return b;
}

return c;
```

OlEzhA0 (changes requested):

src/calculateRentalCost.js
IvanFesenko: Didn't fixed. Do not use else statment

IvanFesenko (changes requested):
Please remove else statment

VitaliyBondarenko1982 (approved):

IvanFesenko (approved):
Good job

PR #2267, opened 2023-01-23
solution

silcki:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    total -= 20;
+  }
+
+  if (days >= 7) {
+    total -= 50;
+  }
+
+  return total;
```
OlEzhA0, line 42: what are the numbers `20`, `40` and `50`? add some meaningful variables

OlEzhA0 (changes requested):

PR #2269, opened 2023-01-23
Solution

OlhaKissa:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  let totalCost = costPerDay * days;
+    
+    if (days >= 7) {
+      totalCost -= 50;
+    }
+    else if (days >= 3) {
+      totalCost -= 20;
+    }
+    
+  return totalCost;
```
IvanFesenko, line 41: Fix indents.
Move discount values to the variables.
Also do not use else statement
2023-01-24_20h16_24

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let costPerDay = 40;
```
IvanFesenko, line 31: It must be const - because you dont change value

IvanFesenko (changes requested):
Some fix required. Please check comment

nazarmatsevych (approved):

PR #2270, opened 2023-01-24
Solution

kirill-vypirovskyi:


yuliia-vereshko (approved):

Andr1yK (changes requested):
What is 20, 50?
Avoid magic numbers

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 7) {
+    return days * rentForDay - bigDiscount;
+  }
+
+  if (days >= 3) {
+    return days * rentForDay - smallDiscount;
+  }
+
+  return days * 40;
```
Andr1yK, line 43: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

Save `days * rentForDay` in some const

```suggestion
    return days * rentForDay - bigDiscount; // days * rentForDay = const
  }

  if (days >= 3) {
    return days * rentForDay - smallDiscount; // days * rentForDay = const
  }

  return days * rentForDay; // const
```

Andr1yK (changes requested):

Andr1yK (approved):

PR #2272, opened 2023-01-25
calculated

edkido:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const total = days * 40;
+
+  if (days >= 7) {
+    return total - 50;
+  }
+
+  if (days >= 3) {
+    return total - 20;
+  }
```
OlEzhA0, line 39: what is `40`, `50` and `20`? just add meaningful variables

OlEzhA0 (changes requested):

vlad-demchuk (approved):

PR #2273, opened 2023-01-26
done

Kocha1991:


src/calculateRentalCost.js
```diff
@@ -22,13 +23,21 @@
...
-  // write code here
+  const payTotal = days * 40;
+
+  if (days < 3) {
+    return payTotal;
+  } else if (days >= 3 && days < 7) {
+    return payTotal - 20;
+  } else {
+    return payTotal - 50;
+  }
```
Bohdan-Maksymiuk, line 40: what a magic numbers 🪄

avoid such nesting `else if` conditions

check checklist one more time 😉

Bohdan-Maksymiuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -22,13 +23,22 @@
...
+  const payOneDay = 40;
+  const discountMore3Days = 20;
+  const discountMore7Days = 50;
+
+  const payTotal = days * payOneDay;
+  const rentMore3Days = payTotal - discountMore3Days;
+  const rentMore7Days = payTotal - discountMore7Days;
+
+  return (days < 3)
+    ? payTotal : (days >= 3 && days < 7) ? rentMore3Days : rentMore7Days;
```
SanyaBratashchuk, line 41: Avoid nested ternary

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -22,13 +23,29 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const payOneDay = 40;
+  const discountMore3Days = 20;
+  const discountMore7Days = 50;
+
+  const payTotal = days * payOneDay;
+  const rentMore3Days = payTotal - discountMore3Days;
+  const rentMore7Days = payTotal - discountMore7Days;
```
ViktorHavryliuk, line 38: Don't use hardcoded numbers in your variable names. If, for example, it is decided in future to give small discont after 4+ days instead, you will need not only to change condition, but also fix this variable's name everywhere

src/calculateRentalCost.js
```diff
@@ -22,13 +23,29 @@
...
+
+  if (days >= 7) {
```
ViktorHavryliuk, line 40: This is not as crucial as suggestion above, but better move amount of days into const too, like bigDiscountThreshold, for example

ViktorHavryliuk (changes requested):

tetiana-ivanchenko (approved):

PR #2274, opened 2023-01-26
Solution

bohdan-gavrilyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
Bohdan-Maksymiuk, line 37: what a magic numbers 🪄

avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):
Good job, need to make some improvements.
PS. js tasks also have checklists 😉

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const smallDiscount = 20;
+  const priceOnDay = 40;
+
+  if (days < 3) {
+    return days * priceOnDay;
+  } else if (days < 7) {
+    return days * priceOnDay - smallDiscount;
+  } else {
+    return days * priceOnDay - bigDiscount;
+  }
```
VitaliyBondarenko1982, line 41: Not fixed comment about nesting esle if

nazarmatsevych (approved):

PR #2275, opened 2023-01-26
:white_check_mark: implement solution for calculateRentalCost task

Bogdan-Mykhailov:
implement solution for calculateRentalCost task using ternary operator

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,7 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days < 3 ? days * 40 : days < 7 ? days * 40 - 20 : days * 40 - 50;
 }
 
 module.exports = calculateRentalCost;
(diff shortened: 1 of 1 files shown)


Yevhenii0536 (approved):
good solution, but i think it will be better if you separate steps at the first time of learn js advanced.
And this hard to read :)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+      return sumForAllDays;
+    }
+
+    case days < 7: {
+      return sumForAllDays - discount;
+    }
+
+    default:
+      return sumForAllDays - specialDiscount;
+  }
```
illiarerun, line 47: i'm not rejecting it this time, but it's better to use plain if's instead of `switch(true` block, it's may be hard to understand

illiarerun (approved):

src/calculateRentalCost.js
Bogdan-Mykhailov: got it

PR #2276, opened 2023-01-26
'done'

Hanna-Balabukha:


ViktorHavryliuk (changes requested):
Please move all the numbers into consts with descriptive names so that it's clear what those numbers mean.
Also, try to find a way to avoid multiple else if, like changing conditions and their order for example. Multiple else if can quickly get out of hand and reduce readability of your code, making it harder to understand, so it's better to avoid them as much as possible

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const minPersDiscount = 20;
+  const maxPersDiscount = 50;
+
+  if (days < minDiscountDays) {
+    return price * days;
+  } else if (days >= maxDiscountDays) {
+    return (price * days) - maxPersDiscount;
+  } else {
+    return (price * days) - minPersDiscount;
+  }
```
loralevitska, line 43: When you return the value from the function in the `if` block, the code will not reach the `else` block, so you can refactor the code to avoid nesting:

```suggestion
  if (days < minDiscountDays) {
    return price * days;
  }
  
  if (days >= maxDiscountDays) {
    return (price * days) - maxPersDiscount;
  }
  
  return (price * days) - minPersDiscount;
```

loralevitska (changes requested):
Almost perfect)
Let's refactor for practice

loralevitska (approved):
Good job!

volodymyr-soltys97 (approved):

PR #2278, opened 2023-01-28
● Solution

d-hubych:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const smallOff = 20;
+  const bigOff = 50;
```
VitaliyBondarenko1982, line 33: Better to make constants out of function.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  const baseAmount = days * dayPrice;
+
+  if (days >= 3 && days < 7) {
```
VitaliyBondarenko1982, line 37: Please, make also variables for these magic numbers

VitaliyBondarenko1982 (changes requested):

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
...
+
 function calculateRentalCost(days) {
-  // write code here
+  const baseAmount = days * dayPrice;
+
+  if (days >= minDaysSmallOff && days < minDaysBigOff) {
+    return baseAmount - smallOff;
+  }
+
+  if (days >= 7) {
```
polosanya, line 43:
```suggestion
  if (days >= minDaysBigOff) {
```

polosanya (approved):

PR #2279, opened 2023-01-29
Solution

marynatovtyn:


qoa11a (changes requested):
let's move numbers to variables with self-describing names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const carRentPrice = 40;
+  let totalCost = carRentPrice * days;
+
+  if (days >= 7) {
+    totalCost -= 50;
+  } else if (days >= 3) {
+    totalCost -= 20;
```
SanyaBratashchuk, line 37: move 50 and 20 to the variables. Also you can get rid of `else if` using return inside `if` statement

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const carRentPrice = 40;
+  const rentDiscountForThreeDays = 20;
+  const rentDiscountForSevenDays = 50;
```
Volodymyr-Melnyk-Fron10der, line 33: These variables will be creating every time function is calling. Move them out of function

Volodymyr-Melnyk-Fron10der (approved):
Cool! But keep in mind that constants should be on global level

PR #2280, opened 2023-01-29
Solution

8Melya8:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
polosanya, line 31: What is result? Let's use more descriptive names
You can also save `days * 40` instead of 0.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    result = days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    result = days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    result = days * 40 - 50;
+  }
```
polosanya, line 43: What are 3, 40, 20, 50 etc.?
Let's create variables for all main values to make your code cleaner and more informative.
You can also immediately return the full price instead of saving it in a variable

polosanya (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  if (days < 3) {
+    discount = 0;
+  }
+
+  if (days >= 3 && days < 7) {
+    discount = 20;
+  }
+
+  if (days >= 7) {
+    discount = 50;
```
IvanFesenko, line 43: Create separate variables for discount, calculate and return result in if statement

IvanFesenko (changes requested):
Some coments were not fixed.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const price = 40;
+  let rentalCoast = days * price;
+
+  if (days >= 3 && days < 7) {
+    rentalCoast -= 20;
+  }
+
+  if (days >= 7) {
+    rentalCoast -= 50;
+  }
```
OlEzhA0, line 40: what is `20` and `50` add please constants for this values

OlEzhA0 (changes requested):

TarasHoliuk (approved):

PR #2281, opened 2023-01-29
Solution

superzuc:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let count = 0;
```
polosanya, line 31: What is count? Let's use more descriptive variables?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < 3) {
+    count = days * 40;
```
polosanya, line 34: What are 40 and 3? Let's create variables for all the main values, so your code will be more clear and understandable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 3) {
+    count = days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    count = days * 40 - 50;
+  }
+
+  return count;
```
polosanya, line 45: Actually, you can save `days * 40` as a variable and just immediately return the full price inside `if`s

polosanya (changes requested):

src/calculateRentalCost.js
superzuc: For 40 i've made a const pricePerDay = 40; but the 3 number is the parameter of the function with a if function so i cannot make a variable for the parameter. Make sense? If not i have to write a function if days < 3 equal a variable daysLessThanThree? 

I try to commit the new resolution in the main time. Thanks for your time

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  let rentalCost = days * pricePerDay;
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
+
+  if (days >= 3 && days < 7) {
+    rentalCost = rentalCost - discountThreeDays;
```
qoa11a, line 37:
```suggestion
    rentalCost -= discountThreeDays;
```

qoa11a (approved):

PR #2282, opened 2023-01-29
solution

xnquaint:


IvanFesenko (changes requested):
Good job, please change variable names 
2023-01-30_16h14_54

IvanFesenko (approved):
Good job

PR #2284, opened 2023-01-29
Solution

GUSILLUS:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+  const carCosts = 40;
+
+  for (let i = 0; i <= days; i++) {
+    result = days * carCosts;
+  }
```
TarasHoliuk, line 36: this loop does nothing
the final `result` value will be `days * carCosts` anyway

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 7) {
+    result -= 50;
+  } else if (days >= 3) {
+    result -= 20;
+  }
```
TarasHoliuk, line 42:
```suggestion
  if (days >= 7) {
    result -= 50;
  } 
  
  if (days >= 3) {
    result -= 20;
  }
```
it's better to avoid `else if` if it's possible
In the such way :arrow_up: code looks clearer

TarasHoliuk (changes requested):

Viktor-Kost (approved):

PR #2285, opened 2023-01-30
Solution

yevhen1408:


ViktorHavryliuk (changes requested):
Please move all the numbers into consts with descriptive names so that it's clear what those numbers mean. Also, try to find solution without using multiple else-if, as else-if statement may quickly get out of hands and make you code harder to read and understand.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= maxdays) {
+    return daycost * days - maxdiscont;
+  } else {
+    if (days >= mindays && days < maxdays) {
+      return daycost * days - mindiscont;
+    } else {
+      return daycost * days;
+    }
+  }
```
tetiana-ivanchenko, line 45:
```suggestion
  if (days >= maxdays) {
    return daycost * days - maxdiscont;
  }
  
  if (days >= mindays && days < maxdays) {
    return daycost * days - mindiscont;
  } 
  
  return daycost * days;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const daycost = 40;
+  const mindiscont = 20;
+  const maxdiscont = 50;
+  const mindays = 3;
+  const maxdays = 7;
```
tetiana-ivanchenko, line 35: use camelCase

tetiana-ivanchenko (changes requested):

loralevitska (approved):

PR #2286, opened 2023-01-30
Solution

katiailchenko:


vadim-ilchenko (approved):
gj 🔥 

1. ideally there shouldn't be changes in package.lock
2. why did you update mate-scripts?

PR #2287, opened 2023-01-30
Solution

iryna-serniuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let amount = 40 * days;
+
+  if (days >= 7) {
+    amount -= 50;
+  } else if (days >= 3) {
+    amount -= 20;
+  }
```
```suggestion
  if (days >= 7) {
    amount -= 50;
  } 
  
  if (days >= 3) {
    amount -= 20;
  }
```
it's better to avoid `else if` if it's possible, in this way code looks clearer

Also, you can move numbers `20` and `50` to variables with meaningful names

TarasHoliuk (changes requested):

etojeDenys (approved):

PR #2288, opened 2023-01-30
Solution

RuslanLiapin:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
+      totalCost = pricePerDay * days;
+      break;
+
+    case days >= 3 && days < 7:
+      totalCost = pricePerDay * days - 20;
+      break;
+
+    case days >= 7:
+      totalCost = pricePerDay * days - 50;
+  }
```
OlEzhA0, line 46: what is `20` and `50`? add some variable for this values

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let totalCost;
+  const pricePerDay = 40;
+
+  switch (true) {
+    case days < 3:
+      totalCost = pricePerDay * days;
+      break;
```
OlEzhA0, line 38: you can return your value at once here
```suggestion
    case days < 3:
      return pricePerDay * days;
```

OlEzhA0 (changes requested):

OlEzhA0 (approved):

PR #2289, opened 2023-01-30
Solution

ArtemVelychko:


qoa11a (changes requested):
let's move numbers to variables with self-describing names

ArtemVelychko:
I moved all numbers to variables in the last change, let me know if anything else needs to be done. Thank you

src/calculateRentalCost.js
```diff
@@ -26,9 +26,20 @@
...
+  const basicDiscount = 20;
+  const discount = 50;
+
+  if (days >= 3 && days < 7) {
+    return totalRentalCost - basicDiscount;
+  } else if (days >= 7) {
+    return totalRentalCost - discount;
+  } else {
+    return totalRentalCost;
+  }
```
```suggestion
  if (days >= 3 && days < 7) {
    return totalRentalCost - basicDiscount;
  }
  
  if (days >= 7) {
    return totalRentalCost - discount;
  }
  
  return totalRentalCost;
```
There is no need to use `else if` as your code will not go further after return anyway)

polosanya (approved):
Good job!

PR #2290, opened 2023-01-30
Solution

Nataliia95254:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let total = 0;
+
+  if (days >= 7) {
+    total = (days * 40) - 50;
+  } else if (days >= 3 && days <= 6) {
+    total = (days * 40) - 20;
+  } else {
+    total = days * 40;
+  }
```
tetiana-ivanchenko, line 39:
```suggestion
  if (days >= 7) {
    return (days * 40) - 50;
  } 
  
  if (days >= 3 && days <= 6) {
    return (days * 40) - 20;
  } 
 
 return days * 40;
  
```

tetiana-ivanchenko (changes requested):
also move 20,40,50 in the variables

src/calculateRentalCost.js
vlad-demchuk: check the previous comment

declare variables with descriptive names and assign these values
image

vlad-demchuk (changes requested):

nazarmatsevych (approved):
👍

PR #2291, opened 2023-01-30
Rental Cost

RufatRM:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  let totalAmount;
+
+  if (days >= 7) {
+    totalAmount = (days * costPerDay) - 50;
```
polosanya, line 35: You can save `days * costPerDay` to a variable too. 
Also, would be great to create variables for all important values (7, 50, 3, 20 etc.) to make your cleaner and more understandable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  } else if (days >= 3) {
+    totalAmount = (days * costPerDay) - 20;
+  } else {
+    totalAmount = days * costPerDay;
+  }
```
polosanya, line 40: You can just immediately return the full price with discount, so you won't need to use else...if

polosanya (changes requested):

polosanya (approved):
Well done!

PR #2292, opened 2023-01-31
Transportation task v.0.1

NedobezhkinOleksandr:


loralevitska (changes requested):
Please move all the numbers into constants with descriptive names so that it's clear what those numbers mean. Also, try to find solution without using multiple else-if, as else-if statement may quickly get out of hands and make you code harder to read and understand.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const totalSum = days * 40;
+  const noDiscountDays = 3;
+  const smallDiscountDays = 7;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  const totalPrice = (days < noDiscountDays) ? totalSum
+    : (days < smallDiscountDays) ? totalSum - smallDiscount
+      : totalSum - bigDiscount;
```
tetiana-ivanchenko, line 39: you should use if statements where you will return right value, don't use nested ternary operator

tetiana-ivanchenko (approved):

tetiana-ivanchenko (changes requested):
please, fix comment

src/calculateRentalCost.js
NedobezhkinOleksandr: should i use only if statement? without else-if and ternary operator? Isn't it?

PR #2293, opened 2023-01-31
solution

Zavalishyn-Oleksandr:


loralevitska (changes requested):
Please move all the numbers into constants with descriptive names so that it's clear what those numbers mean.

Zavalishyn-Oleksandr:
Fixed

SanyaBratashchuk (approved):

PR #2296, opened 2023-02-01
Solution

stanislavcodes:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,23 @@
...
+  };
+  const majorDiscount = {
+    startDay: 7,
+    value: 50,
+  };
+  const finalDiscount = days >= majorDiscount.startDay
+    ? majorDiscount.value
+    : days >= minorDiscount.startDay
+      ? minorDiscount.value
+      : 0;
```
illiarerun, line 44: don't use nested ternaries, remake it with plain ifs where you can return values right away

illiarerun (changes requested):

src/calculateRentalCost.js
```diff
@@ -27,8 +27,26 @@
...
+
+  if (days >= majorDiscount.startDay) {
+    return days * rentalPriceForDay - majorDiscount.value;
+  }
+
+  if (days >= minorDiscount.startDay) {
+    return days * rentalPriceForDay - minorDiscount.value;
+  }
+
+  return days * rentalPriceForDay;
```
illiarerun, line 49: you repeat this calcualtion three times, create a variable for that, DRY

illiarerun (approved):

PR #2297, opened 2023-02-01
Solution

diamat1997:


ViktorHavryliuk (changes requested):
Please move all numbers to constants with descriptive names, so that it’s clear what those numbers mean

tetiana-ivanchenko (approved):

PR #2299, opened 2023-02-03
Develop

the-tique:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  switch (true) {
+    case days >= 3 && days < 7:
+
+      return cost - 20;
+
+    case days >= 7:;
+
+      return cost - 50;
+  }
```
SanyaBratashchuk, line 42: In my opinion `switch (true)` is not the best practice use if

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const singleDayRent = 40;
+  const cost = days * singleDayRent;
+
+  switch (true) {
+    case days >= 3 && days < 7:
+
+      return cost - 20;
```
SanyaBratashchuk, line 37: move all magic numbers to the variables

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #2300, opened 2023-02-03
Solution

Volodymyr-Kustikov:


ViktorHavryliuk (changes requested):
Please move all the numbers into consts with descriptive name, so that it's clear what they mean
Also try to reorganise your conditions to omit else ifs, as using multiple else if can quickly get out of hands and reduce readablity of your code

Volodymyr-Kustikov:
is it better?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+    case days >= 7:
+      result = sumLong;
+  }
+
+  return result;
 }
 
-module.exports = calculateRentalCost;
+
+  module.exports = calculateRentalCost;
```
SanyaBratashchuk, line 54:
```suggestion
  module.exports = calculateRentalCost;
  
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+      result = sum;
+      break;
+
+    case days >= 3 && days < 7:
+      result = sumContinue;
+      break;
+    
+    case days >= 7:
+      result = sumLong;
+  }
```
SanyaBratashchuk, line 48: In my opinion `switch(true)` is not the best practice so better use ifs

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const sum = days * price;
+  const sumContinue = sum - 20;
+  const sumLong = sum - 50;
```
SanyaBratashchuk, line 34: move 20 and 50 to the separated variables

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const fifty = 50;
+  const twenty = 20;
```
SanyaBratashchuk, line 33: Bad naming.
```suggestion
  const  = 50;
  const basicDiscount = 20;
```
come up with a better name for the first variable

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    return sum;
+  }
+
+  if (days >= 3 && days < 7) {
+    return sumContinue;
+  }
+
+  if (days >= 7) {
+    return sumLong;
+  }
```
SanyaBratashchuk, line 48:
```suggestion
   if (days >= 7) {
    return sumLong;
  }
  if (days >= 3) {
    return sumContinue;
  }
  
    return sum;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const sum = days * price;
```
SanyaBratashchuk, line 34:
```suggestion
  const basePrice = days * price;
```

SanyaBratashchuk (approved):

PR #2301, opened 2023-02-04
Solution

BogdanParubchenk01:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  let result = 0;
+
+  if (days < 3) {
+    result = rent * days;
```
nazarmatsevych, line 35:
```suggestion
    return rent * days;
```

and you cant remove your result variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    result = rent * days - 20;
+  }
+
+  if (days >= 7) {
+    result = rent * days - 50;
+  }
```
nazarmatsevych, line 44: Move rent * days to variable to avoid code dublication, also move '20' and '50' to variables

nazarmatsevych (changes requested):

PR #2304, opened 2023-02-05
Solution

NYasnaya:


ViktorHavryliuk (changes requested):
Please move all the numbers to consts with descriptive names so that it is clear what they mean

NYasnaya:
> Please move all the numbers to consts with descriptive names so that it is clear what they mean

If I do all variables const, tests don't pass, only if they const and let 
image

NYasnaya:
now is OK
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  if (days >= 3) {
+    totalRentalCost = days * rentPerDay - 20;
+  }
+
+  if (days >= 7) {
+    totalRentalCost = days * rentPerDay - 50;
+  }
+
+  return totalRentalCost;
 }
```
DarkMistyRoom, line 43:
```suggestion
  const rentPerDay = 40;
  let totalRentalCost = days * rentPerDay;

  if (days >= 3) {
    totalRentalCost -= 20;
  }

  if (days >= 7) {
    totalRentalCost -= 50;
  }

  return totalRentalCost;
}
```

DarkMistyRoom: be careful about the previous comment. you shouldn't make all variables constants, but you should move all numbers to constants with descriptive names

DarkMistyRoom: if you have questions about it, please write in the chat

DarkMistyRoom (changes requested):

loralevitska (approved):
Good job!

PR #2305, opened 2023-02-06
solution task

ivanm2706:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
tetiana-ivanchenko, line 32: move magic numbers in the variables

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2306, opened 2023-02-06
solution

meltourme:


qoa11a (changes requested):
let's move numbers to variables with describing names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const oneDay = 40;
+  let total = oneDay * days;
+  const discountBig = 50;
+  const discountSmall = 20;
```
ivbutsykin, line 34: You should name constants in uppercase (`COST_PER_DAY`, `BIG_DISCOUNT`, ...)

https://mate-academy.github.io/style-guides/javascript.html#2310you-may-optionally-uppercase-a-constant-only-if-it-1-is-exported-2-is-aconstit-can-not-be-reassigned-and-3-the-programmer-can-trust-it-and-its-nested-properties-to-never-change

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days >= 7) {
+    total = total - discountBig;
+  } else if (days >= 3) {
```
ivbutsykin, line 38: `7` and `3` also should be used as constants

ivbutsykin (changes requested):

DarkMistyRoom (approved):

PR #2310, opened 2023-02-07
Solution

Anatolii-Polishchuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
+
+  if (days >= 7) {
+    totalCost = totalCost - 50;
+  } else if (days >= 3) {
+    totalCost = totalCost - 20;
+  }
```
loralevitska, line 37:
```suggestion
  if (days >= 7) {
    return totalCost - 50;
  }
  
   if (days >= 3) {
    return  totalCost - 20;
  }
```

loralevitska (changes requested):
Almost done)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  if (days >= 7) {
+    return totalCost - 50;
+  }
+
+  if (days >= 3) {
+    return totalCost - 20;
+  }
+
+  return totalCost;
 }
```
DarkMistyRoom, line 42: Lore meant that you should put all numbers into variables. For example: 
`const PRICE = 40;
const DISCOUNT_BIG = 50;`
and so on.

of you have questions, feel free to ask in the chat

DarkMistyRoom (changes requested):

slavanosatskyi (approved):
good work!

ViktorHavryliuk (approved):

PR #2311, opened 2023-02-07
Task done

sergik94:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  switch (true) {
+    case days < 3:
+      return sum;
+
+    case days < 7:
+      return sum - 20;
+
+    case days >= 7:
+      return sum - 50;
+  }
```
etojeDenys, line 42: it would be better to use here `if` instead of `switch`

etojeDenys (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return sum;
+  }
+
+  if (days < 7) {
+    return sum - 20;
+  }
+
+  if (days >= 7) {
+    return sum - 50;
+  }
```
```suggestion
    return sum - 50;
```

etojeDenys (approved):
LGTM,

small nit

PR #2312, opened 2023-02-07
Solution

luka4ok:
Please check my solution for the task.

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalRent = days * 40;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const totalRent = days * 40;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return totalRent - minDiscount;
+  } else if (days >= 7) {
+    return totalRent - maxDiscount;
+  }
```
```suggestion
  if (days >= 3 && days < 7) {
    return totalRent - minDiscount;
  } 
  
  if (days >= 7) {
    return totalRent - maxDiscount;
  }
```

just FYI - better to avoid `else if`. it improves readability

TarasHoliuk (approved):
:+1:

PR #2313, opened 2023-02-07
solution

MSA1991:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
```
ivbutsykin, line 31: Please move your numbers (40, 7, 50, 3, 20) in constant, like:
const `COST_PER_DAY = 40`, `BIG_DISCOUNT = 50`, ...

ivbutsykin (changes requested):

ViktorHavryliuk (approved):

PR #2314, opened 2023-02-07
Solution

andrii-kovalskyi02:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    return costOfAllDays;
+  }
+
+  if (days < 7) {
+    return costOfAllDays - 20;
+  }
+
+  return costOfAllDays - 50;
```
DarkMistyRoom, line 41: Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

DarkMistyRoom (changes requested):

ViktorHavryliuk (approved):

PR #2315, opened 2023-02-07

JulyaPetrovskaya:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    prise = days * oneDayPay - 20;
+  }
+
+  if (days >= 7) {
+    prise = days * oneDayPay - 50;
+  }
+
```
tetiana-ivanchenko, line 45: move magic numbers in the variables, also if you will return value inside your if block, you can avoid one of them

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayPay = 40;
+  const discountForThreeDays = 20;
+  const discountForSevenDays = 50;
+
+  if (days < 3) {
```
mykhalenych, line 35: 3 and 7 still magic numbers

mykhalenych (changes requested):

src/calculateRentalCost.js
JulyaPetrovskaya: never heard about magic numbers, maybe I missed this topic

mykhalenych (approved):

PR #2316, opened 2023-02-07
solution

serhiilysenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days >= 7) {
+    cost = days * 40 - 50;
+  }
+
+  if (days < 3) {
+    cost = days * 40;
+  }
+
+  return cost;
 }
```
ivbutsykin, line 46: Have some remarks for you:
1. Create separate constants for numbers (3, 7, 40, 20), like: `COST_PER_DAY = 40`, `BIG_DISCOUNT = 50`, ...
2. Doesn't change cost variable, instead this create `total` variable for ex:
```
const total = days * 40;
```
3. You can complete this tasks only with 3 conditions (not 4):
```
// Notice: you should use constants
if (days >= 7) {
  return total - 50;
}

if (days >= 3) {
  return total - 20;
}

return total
```

ivbutsykin (changes requested):

Anastasiia-Svintsova (approved):

PR #2317, opened 2023-02-08
'Solution'

nazarkharkevych:


ViktorHavryliuk (changes requested):
Please move all the numbers into consts with descriptive names, so that it's clear what those numbers mean. Also, try to find a solution without using multiple else if, as they can quicly get out of hands and reduce readability of your code

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const costPerDay = 40;
+  const threeDaysDiscount = 20;
+  const sevenDaysDiscount = 50;
+  const totalCost = days * costPerDay;
+
+  const discount = days >= 3
+    ? days < 7
+      ? threeDaysDiscount
+      : sevenDaysDiscount
+    : 0;
```
DarkMistyRoom, line 40: nested ternary is evil. better avoid it

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const threeDaysDiscount = 20;
+  const sevenDaysDiscount = 50;
+  const totalCost = days * costPerDay;
+
+  const discount = days >= 3
+    ? days < 7
```
DarkMistyRoom, line 37: and also move 3 and 7 to constants too. because now we should gues what do they mean

DarkMistyRoom (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  switch (true) {
+    case days >= sevenDays:
+      return totalCost - sevenDaysDiscount;
+
+    case days >= threeDays:
+      return totalCost - threeDaysDiscount;
+
+    default:
+      return totalCost;
+  }
```
loralevitska, line 47: Better use switch when you have to completely compare your value with another. In your case, it is better to use if, but without nesting as indicated in previous comments
```suggestion
    if (days >= sevenDays) {
      return totalCost - sevenDaysDiscount;
   }
   
   if (days >= threeDays) {
      return totalCost - threeDaysDiscount;
   }
   
   return totalCost;
  }
```

loralevitska (approved):

PR #2318, opened 2023-02-08
solution

IamLoren:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const discount2 = 50;
+  let totalAmount = 0;
+
+  if (days > 0 && days < 3) {
+    totalAmount = days * price;
+  } else if (days >= 3 && days < 7) {
+    totalAmount = (days * price) - discount1;
+  } else if (days >= 7) {
+    totalAmount = (days * price) - discount2;
+  }
```
VitaliyBondarenko1982, line 42:
```suggestion
  if (days < 3) {
   return days * price;
  } 
  
   if (days >= 3 && days < 7) {
     return (days * price) - discount1;
  } 
  
  return (days * price) - discount2;
  
```

VitaliyBondarenko1982: Else if makes code not good readable.
You can return immediately from if statement

VitaliyBondarenko1982 (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const discount1 = 20;
+  const discount2 = 50;
```
IvanFesenko, line 33: Please change variable names to more logical
2023-02-10_15h00_11

IvanFesenko (changes requested):
Please rename variable

PR #2322, opened 2023-02-09
Solution

alina-kabanets:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const sum = days * 40;
+  let result = 0;
```
polosanya, line 32: Let's use more descriptive names than just sum and result

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days < 3) {
+    result = sum;
+  } else {
+    if (days >= 3 && days < 7) {
+      result = sum - 20;
+    } else {
+      result = sum - 50;
+    }
```
polosanya, line 41: Do you really need to store this values into result?
Try to return the full price immediately after check. It will help to avoid redundant else...if.
Also, create a variables for all main values (20, 50...) to make your code cleaner

polosanya (changes requested):
Almost done!

src/calculateRentalCost.js
alina-kabanets: Ok, I got it. Thank you!

agniya-i (approved):

PR #2323, opened 2023-02-09
totalSumRent

DmytroHorshunov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
-  // write code here
+  const price = 40;
+  let sumRent;
+
+  if (days < 3) {
+    sumRent = price * days;
+  } else if (days < 7) {
+    sumRent = price * days - 20;
+  } else {
+    sumRent = price * days - 50;
```
SanyaBratashchuk, line 39: move 20 and 50 to the variables and get rid of else and else if via return statement in your if's

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
DmytroHorshunov: I hope I understood correctly

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const price = 40;
+  const firstDiscont = 20;
+  const secondDiscont = 50;
```
OleziO, line 33: Constant variables should be named in upper case and snake case notation.
Example: PRICE, FIRS_DISCOUNT і тд.

OleziO (approved):

PR #2324, opened 2023-02-09
Solution

IvanVaverchak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  switch (true) {
+    case days < 3:
+      return days * 40;
+
+    case days >= 3 && days < 7:
+      return days * 40 - 20;
+
+    case days >= 7:
+      return days * 40 - 50;
+  }
```
Viktor-Kost, line 40: no need to use switch case here. 
Use just 
```
if({{your_condition}}) {
  return {{your_result}}
}
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  switch (true) {
+    case days < 3:
+      return days * 40;
```
Viktor-Kost, line 33: create a variable for days * 40, because this calculation repeat a few times at your code

Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  if (days < 3) {
+    return price;
+  }
+
+  if (days >= 3 && days < 7) {
+    return price - 20;
+  }
+
+  if (days >= 7) {
+    return price - 50;
```
vlad-demchuk, line 42: also, move `20, 50` to separate variables

vlad-demchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  }
```
OlEzhA0, line 43: what is `20` and `50`? just add some constants

OlEzhA0 (approved):
hey! I approve your work but add small improvements

PR #2328, opened 2023-02-10
Solution

DaveBeetle:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const minDiscount = 20;
+  const maxDiscount = 50;
```
Volodymyr-Melnyk-Fron10der, line 32: I suggest to move these constants out of `calculateRentalCost`
```suggestion
  const MIN_DISCOUNT= 20;
  const MAX_DISCOUNT = 50;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, Great work, only one suggestion for constants

SanyaBratashchuk (approved):

PR #2330, opened 2023-02-11
Solution

DenysDubas:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 2 && days < 7) {
+    return 40 * days - 20;
```
BozhenaBodnar, line 32: you can create variable for days*40)but in overall good:)

BozhenaBodnar (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  }
+
+  if (days > 6) {
+    return 40 * days - 50;
+  }
+
+  return days * 40;
```
illiarerun, line 39: what are 40, 50, 20, 2, 6, 7? create variables for main values

src/calculateRentalCost.js
illiarerun: Agree, you repeat `days*40` calculation 3 times, create a variable for that

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 2 && days < 7) {
```
illiarerun, line 31: you can place this condition after `6 days` one to omit additional check
```suggestion
  if (days > 2) {
```

illiarerun (changes requested):

illiarerun (approved):

PR #2332, opened 2023-02-11
Solution

dudar-y:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  let discount = 0;
+
```
Volodymyr-Melnyk-Fron10der, line 33: Keep in mind that constants should be uppercased and be initialized on the global level, at the top of the file
```suggestion
const DAILY_PRICE = 40;
  
function calculateRentalCost(days) {
  let discount = 0;

```

Volodymyr-Melnyk-Fron10der (changes requested):
I like it✅ But one comment about constants

src/calculateRentalCost.js
```diff
@@ -27,8 +27,20 @@
...
-  // write code here
+  let discount = 0;
+
+  if (days >= 3) {
+    discount = 20;
+  }
+
+  if (days >= 7) {
+    discount = 50;
+  }
```
DarkMistyRoom, line 41: let's move all these numbers to constants to give them descriptions and make them more understandable

DarkMistyRoom (changes requested):

src/calculateRentalCost.js
```diff
@@ -27,8 +27,20 @@
...
+
 function calculateRentalCost(days) {
-  // write code here
+  let totalPrice = DAILY_PRICE * days;
+
+  if (days >= 7) {
+    totalPrice -= MAX_DISCOUNT;
+  } else if (days >= 3) {
+    totalPrice -= MIN_DISCOUNT;
+  }
```
Viktor-Kost, line 41: You can just use early return instead of `else if `nesting. Try to avoid else/if. It makes your code clean
```suggestion
  if (days >= 7) {
   return totalPrice - MAX_DISCOUNT;
  } 
  if (days >= 3) {
    return totalPrice - MIN_DISCOUNT;
  }
```

Viktor-Kost (approved):
Nice!

PR #2334, opened 2023-02-11
Solution

yuliia-vereshko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return total - 50;
+  }
+
+  if (days >= 3) {
+    return total - 20;
+  }
+
+  if (days < 3) {
+    return total;
+  }
```
KuharchukDmytro, line 43: on my opinion, you dont need if statement here, because if previus condition are false, you anyway will return total

KuharchukDmytro (approved):
good job!

Andr1yK (changes requested):
What is 20, 50?
Avoid magic numbers

Andr1yK (changes requested):
What is 20, 50?
Avoid magic numbers

Andr1yK (approved):

PR #2339, opened 2023-02-12
Solution

yaroslav-vasilyev:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+  const smallestDiscount = 20;
+  const biggestDiscount = 50;
+
+  if (days >= 7) {
+    totalPrice -= biggestDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
+    totalPrice -= smallestDiscount;
+  }
```
timoxa2t, line 43:
```suggestion
  if (days >= 7) {
    totalPrice -= biggestDiscount;
  } else if (days >= 3) {
    totalPrice -= smallestDiscount;
  }
```
You can use "else if" to make code shorter.
```suggestion
  if (days >= 7) {
    return totalPrice - biggestDiscount;
  }

  if (days >= 3 ) {
    return totalPrice - smallestDiscount;
  }
```

or like this

timoxa2t (approved):
Good work on variables. You could make code cleaner

src/calculateRentalCost.js
Andr1yK, line 43: it is difficult to understand where we have changed totalPrice and what will returned

Andr1yK: I like second variant)

Andr1yK (changes requested):
Check by checklist

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
```
Andr1yK, line 31: remove useless comments

Andr1yK (approved):

PR #2342, opened 2023-02-12
Solution

BiletskyiVolodymyr:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRentPrice = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+
+  if (days >= 7) {
+    return days * dayRentPrice - bigDiscount;
```
Anastasiia-Svintsova, line 36: Move `days * dayRentPrice` to a variable, you use it 3 times

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days >= 3 && days < 7) {
+    return days * dayRentPrice - smallDiscount;
+  } else if (days < 3) {
+    return days * dayRentPrice;
+  }
```
Anastasiia-Svintsova, line 41: You don't need other checks here, just return a value
```suggestion
  } 
  
  return days * dayRentPrice;
```

Anastasiia-Svintsova (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const dayRentPrice = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const rentPrice = dayRentPrice * days;
+
+  if (days >= 7) {
+    return rentPrice - bigDiscount;
+  } else if (days >= 3 && days < 7) {
+    return rentPrice - smallDiscount;
+  }
```
witflash, line 40:
```suggestion
  if (days >= 7) {
    return rentPrice - bigDiscount;
  }
  
  if (days >= 3 && days < 7) {
    return rentPrice - smallDiscount;
  }
```

witflash: You don't need to use `else` if your `if` block returns smth.

witflash (approved):
Good work.
Check my comment for your future work.

PR #2343, opened 2023-02-12
Solution

Yevhenii0536:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
```
vadosik15, line 36: You can use 
if ( ) {
} else if ( ) {
}

construction

vadosik15 (approved):

src/calculateRentalCost.js
Yevhenii0536: yes, i know. but in some issues it could create some problems. And that's why i don't use this construction

Andr1yK: @vadosik15 Why do u need to use `else if` after `return`?

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
```
Andr1yK, line 31: do u need this comment?

src/calculateRentalCost.js
Andr1yK, line 36: what is 40/50/20?

Avoid magic numbers

Andr1yK (changes requested):

Andr1yK (approved):

PR #2345, opened 2023-02-13
Solution

MaksymMaliuk:


Lilu580 (commented):
gj

Andr1yK (changes requested):
What is 20, 50?
Avoid magic numbers

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 7) {
+    return days * price - discountFirst;
+  }
+
+  if (days >= 3) {
+    return days * price - discountSecond;
+  } else {
+    return days * price;
+  }
```
Andr1yK, line 43: check by checklist

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const discountFirst = 50;
+  const discountSecond = 20;
```
Andr1yK, line 33: Why first and second? where do I need to use first?

Andr1yK (changes requested):

PR #2347, opened 2023-02-13
solution

kuzmi491:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const result = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return result - 20;
+  }
+
+  if (days >= 7) {
+    return result - 50;
+  }
```
NikitaStambul, line 39: you can switch if statements places and don't write long condition
```suggestion
  if (days >= 7 ) {
    return result - 50;
  }

  if (days >= 3) {
    return result - 20;
  }
```

NikitaStambul (approved):

src/calculateRentalCost.js
illiarerun: agree

illiarerun, line 39: [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 1st item, I don't know what 40, 50, 20, 7 and 3 are

illiarerun (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const rent = 40;
+  const result = days * rent;
+  const firstdiscount = 20;
+  const seconddiscount = 50;
+
+  if (days >= 7) {
+    return result - seconddiscount;
+  }
+
+  if (days >= 3) {
```
illiarerun, line 40: create variables for these values, too, 7 and 3

illiarerun (approved):

PR #2362, opened 2023-02-13
solution

chernenko2:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
...
+
+  if (days >= 3 && days < 7) {
+    return (40 * days) - 20;
+  }
+
+  if (days >= 7) {
+    return (40 * days) - 50;
+  }
+
+  return 40 * days;
```
illiarerun, line 41: [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 1st item, I don't know what 40, 50, 20, 7 and 3 are

illiarerun, line 41: you repeat this calculation 3 times, create a variable for it

src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  if (days >= 3 && days < 7) {
```
illiarerun, line 33: swap your if's blocks order so you can omit second check
```suggestion
  if (days >= 3) {
```

illiarerun (changes requested):

src/calculateRentalCost.js
chernenko2: > swap your if's blocks order so you can omit second check

Thank you for your advice!)

illiarerun (approved):

PR #2365, opened 2023-02-13
calculateRentalCost

Stepan-TS:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 0) {
+    return 0;
+  } else if (days > 0 && days < 3) {
+    return totalWithoutDiscount;
+  } else if (days >= 3 && days < 7) {
+    return totalWithoutDiscount - smallDiscount;
+  }
```
Andr1yK, line 41: do u really need else if after return?

Andr1yK (changes requested):

Andr1yK (approved):

PR #2369, opened 2023-02-14
Solution

MaksymHuriev:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let total = 0;
+
+  for (let i = 0; i < days; i++) {
+    total += 40;
+  }
```
Volodymyr-Melnyk-Fron10der, line 36: Why don't just do?))
```suggestion
  let total = days * 40;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let total = 0;
```
Volodymyr-Melnyk-Fron10der, line 32: I suggest to name it `price`

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, check comments + move 20 and 50 to the constants

ViktorHavryliuk (approved):

PR #2373, opened 2023-02-15
Calculate

OlegRak86:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
-  // write code here
+  const sum = days * 40;
+  let total = 0;
+
+  if (days >= 3 && days < 7) {
+    total = sum - 20;
+  } else if (days >= 7) {
+    total = sum - 50;
+  } else {
+    total = sum;
```

Bohdan-Maksymiuk: and avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):
Good jo, but pls, pay attention to [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

PR #2374, opened 2023-02-16
Solution for first HW

vanya-kalyenichenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  for (let i = 0; i < days; i++) {
```
slavanosatskyi, line 33: loop is not need here - days * 40 will give you what you need

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    return result - 20;
+  }
+
+  if (days >= 7) {
+    return result - 50;
+  }
+
```
slavanosatskyi, line 44:
```suggestion
  if (days >= 7) {
    return result - 50;
  }
  
   if (days >= 3) {
    return result - 20;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  for (let i = 0; i < days; i++) {
+    result += 40;
```
slavanosatskyi, line 34: move all magic number (40, 20, 50) to constants with meaningful names

slavanosatskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+  const rent = 40;
+  const smallDiscond = 20;
+  const bigDiscond = 50;
+
+  result += rent * days;
```
SerhiiKirik, line 36:
```suggestion
  const rent = 40;
  const smallDiscond = 20;
  const bigDiscond = 50;
  const priceForAllDays = rent * days;
```

SerhiiKirik (approved):
Well done 🚀

PR #2376, opened 2023-02-17
solution

YuriyLarin:


ViktorHavryliuk (changes requested):
Please move all the numbers into variables with descriptive names so that it's clear what they mean. Also, try not to use multiple else if, it's not considered a good practice

nazarmatsevych (approved):

PR #2378, opened 2023-02-17
solution

swiftbuddy03:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    total = perDays * day;
+  }
+
+  if (day >= 3) {
+    total = (day * perDays) - 20;
+  }
+
+  if (day >= 7) {
+    total = (day * perDays) - 50;
+  }
```
AlisaVlasova, line 45: great)
let`s move all magic numbers to variables

AlisaVlasova (changes requested):

src/calculateRentalCost.js
swiftbuddy03: what is wrong?

swiftbuddy03: how like this? ;)

PR #2379, opened 2023-02-18
Solution

Pavlo-Petrashevskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  let totalPrice = price * days;
+
+  if (days >= 7) {
+    totalPrice -= 50;
+  } else if (days >= 3 && days < 7) {
+    totalPrice -= 20;
+  }
```
nazarmatsevych, line 38: You can create separate if statement and return the correct totalPrice inside every if statements instead of else if structure.

Also move this 'magic numbers' 20 and 50 to variables

nazarmatsevych (changes requested):

src/calculateRentalCost.js
Pavlo-Petrashevskyi: done

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
```
Volodymyr-Melnyk-Fron10der, line 33: 1. Constants should be uppercased 
2. Constants shouldn't be placed inside function, you have to move them on the global level
```suggestion
  const PRICE = 40;
  const SMALL_DISCOUNT = 20;
  const BIG_DISCOUNT = 50;
```

Volodymyr-Melnyk-Fron10der (changes requested):

src/calculateRentalCost.js
Pavlo-Petrashevskyi: done

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
  * @return {number}
  */
 
+const PRICE = 40;
+const SMALL_DISCOUNT = 20;
+const BIG_DISCOUNT = 50;
+
 function calculateRentalCost(days) {
-  // write code here
+  const TOTAL_PRICE = PRICE * days;
```
ivbutsykin, line 35: Hey, this value should in camelCase 🙂
```suggestion
  const totalPrice = PRICE * days;
```

ivbutsykin (approved):
I guess you have some confusion with constants naming (someone should be in uppercase and other one in camelCase). Please see this thread: https://mate-parttime.slack.com/archives/C03UQS59NRY/p1675851423307729. If you have any questions else, please ask them in the chat.

I approved this PR, but please, you should clearly understand how to name constants.

src/calculateRentalCost.js
Pavlo-Petrashevskyi: Thank you very much for a link!

PR #2380, opened 2023-02-18
Solution

Hamuud:


asmd564 (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  switch (true) {
```
jstmpelowycz, line 33: you don't need switch-case-statement here, let's use simple if-statements
when you use switch-case-statement alongside boolean — it means something goes wrong

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    case days < 3:
+      result = days * 40;
+
+      return result;
+    case days >= 3 && days < 7:
+      result = (days * 40) - 20;
+
+      return result;
+    case days >= 7:
+      result = (days * 40) - 50;
```
jstmpelowycz, line 43: you seem to repeat yourself with `days * 40`

jstmpelowycz (changes requested):
see the comments, also — make your code descriptive, use variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  let result = days * price;
```
jstmpelowycz, line 32: see the checklist, do not use `let`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+  if (days < 3) {
+    return result;
+  }
+
+  if (days >= 3 && days < 7) {
```
jstmpelowycz, line 38: here, `days >= 3` is always true because the previous condition is already omitted

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    result = result - 20;
+
+    return result;
+  }
+
+  if (days >= 7) {
```
jstmpelowycz, line 44: you don't need this statement

jstmpelowycz (changes requested):
let's introduce variables for 3, 7, 20, ...

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const littleDiscount = 20;
+  const dayThree = 3;
+  const daySeven = 7;
```
jstmpelowycz, line 34: the checklist, p.5

jstmpelowycz (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const littleDiscount = 20;
+  const basicRent = 3;
+  const bigRent = 7;
```
jstmpelowycz, line 34:
```suggestion
  const basicDiscountPeriod = 3;
  const greatDiscountPeriod = 7;
```

jstmpelowycz (approved):

PR #2381, opened 2023-02-19
Solution for Calculate Rental Cost

ctk12:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
```
Volodymyr-Melnyk-Fron10der, line 31: don't commit any comments

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
+  if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
```
Volodymyr-Melnyk-Fron10der, line 33: move 40, 20, 50 into constant out of function, keep in mind that constants should be uppercased

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
+  }
+
+  if (days >= 7) {
+    return (days * 40) - 50;
+  }
+
+  return days * 40;
```
Volodymyr-Melnyk-Fron10der, line 40: days * 40 is repeating 3 times, move it into a variable and reuse it throughout your code

Volodymyr-Melnyk-Fron10der (changes requested):
Almost done

qoa11a (approved):

PR #2383, opened 2023-02-19
Solution

metja:


ViktorHavryliuk (changes requested):
Please move all the numbers into consts with descriptive names so that it's clear what those numbers mean.
Also, never use constant values in conditional statements or loops, like `if(true)`, so implement this with ifs instead of switch

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const discountBig = 50;
+  const totalCost = days * cost;
+
+  if (days >= 7) {
+    return totalCost - discountBig;
+  } else if (days < 7 && days >= 3) {
+    return totalCost - discountSmall;
+  } else {
+    return totalCost;
+  }
```
loralevitska, line 42: When you use return you don't need `else` statement. 
```suggestion
  if (days >= 7) {
    return totalCost - discountBig;
  } 
  
  if (days < 7 && days >= 3) {
    return totalCost - discountSmall;
  }
  
  return totalCost;
```

loralevitska (approved):

nazarmatsevych (approved):

PR #2384, opened 2023-02-20
Develop

RatuhWD:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let solution = days * 40;
```
sTorba24, line 31: use constants, to define numerical values

sTorba24, line 31:
```suggestion
  const carRentPrice = 40
  let solution = days * carRentPrice ;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days > 6) {
+    solution -= 50;
```
sTorba24, line 34: the same here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days > 2) {
+    solution -= 20;
```
sTorba24, line 36: the same

sTorba24 (changes requested):
Good work! Some comments to make code cleaner

PR #2385, opened 2023-02-20
Solution

StepuraSerhii:
Solution with switch


ViktorHavryliuk (changes requested):
Move all the numbers into consts with descriptive names so that it's clear what those numbers mean
Also, never use constant values as condition for control statements or loops (like `if(true)`), so inthis case it's better to use some if statements instead of switch

ViktorHavryliuk (approved):

PR #2387, opened 2023-02-20
sum calculation

JasterHim:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,11 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 2) {
+    return days > 6 ? days * 40 - 50 : days * 40 - 20;
+  }
+
```
etojeDenys, line 34:
```suggestion
  const totalCost = days * 40;
  
  if (days >= 7) {
    return totalCost -  bigDiscount;
  }
  
  if (days >= 3) {
    return totalCost - smallDiscount;
   }

```

etojeDenys (changes requested):
let`s improve readability(look my comment) and move this numbers into variables

tetiana-ivanchenko (approved):

PR #2388, opened 2023-02-20
solution 

katerynashylina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  if (days >= 1 && days < 3) {
+    totalCost = fullCost;
+  }
+
+  if (days >= 3 && days < 7) {
+    totalCost = fullCost - 20;
+  }
+
+  if (days >= 7) {
+    totalCost = fullCost - 50;
```
loralevitska, line 44: Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40;
+  const fullCost = rentalCost * days;
+  let totalCost = 0;
+
+  if (days >= 1 && days < 3) {
+    totalCost = fullCost;
```
loralevitska, line 36: Try to simplify you code with return. So move from small to higher rent and return value in each `if` or from function in the end
```suggestion
    return fullCost;
```

loralevitska (changes requested):
Almost perfect! Please pay attention to my comments

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return fullCost;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (fullCost - firstDiscount);
+  }
+
+  if (days >= 7) {
+    return (fullCost - secondDiscount);
+  }
```
etojeDenys, line 46:
```suggestion
  if (days >= 7) {
    return fullCost - secondDiscount;
  }

  if (days >= 3) {
    return fullCost - firstDiscount;
  }

  return fullCost
```

etojeDenys (changes requested):
small nit:

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40;
+  const firstDiscount = 20;
```
DarkMistyRoom, line 32:
```suggestion
  const smallDiscount = 20;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const secondDiscount = 50;
```
DarkMistyRoom, line 33:
```suggestion
  const bigDiscount = 50;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const fullCost = rentalCost * days;
+
+  if (days >= 7) {
+    return fullCost - secondDiscount;
+  }
+
+  if (days >= 3) {
```
DarkMistyRoom, line 40: also, make sense to move the days amount to variables too

DarkMistyRoom (approved):

PR #2389, opened 2023-02-20
Solution

Ihor-Zakharov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = days * 40;
+
+  if (days >= 7) {
+    return rentalCost - 50;
+  }
+
+  if (days >= 3) {
+    return rentalCost - 20;
```
Volodymyr-Melnyk-Fron10der, line 38: move magic numbers into constants, but keep in mind that constants should be uppercased

Volodymyr-Melnyk-Fron10der (changes requested):
Almost done👍

SanyaBratashchuk (approved):

PR #2392, opened 2023-02-21
Rental cost solution

julmak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total;
+
+  if (days >= 7) {
+    total = days * 40 - 50;
```
vlad-demchuk, line 34: what are 40, 50, 20? move these values to variables

vlad-demchuk (changes requested):

Volodymyr-Melnyk-Fron10der (changes requested):

Anastasiia-Svintsova (approved):

PR #2393, opened 2023-02-21
first

Vladimir-Zadorozhnyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let amount = days * 40;
+
+  if (days >= 7) {
+    amount -= 50;
+  } else if (days >= 3) {
+    amount -= 20;
```
vlad-demchuk, line 36: what are 20, 50? move them to separate variables with appropriate names

vlad-demchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let amount = days * 40;
```
polosanya, line 31: Let's specify what exactly this amount is. Also, what is 40?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const firstDiscount = 20;
+  const secondDiscount = 50;
```
```suggestion
  const smallDiscount = 20;
  const bigDiscount = 50;
```
I think this would be more logical

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    amount -= secondDiscount;
+  } else if (days >= 3) {
+    amount -= firstDiscount;
+  }
```
```suggestion
  if (days >= 7) {
    return amount - secondDiscount;
  }
  
  if (days >= 3) {
    return amount - firstDiscount;
  }
```
You can just immediately return the final value to avoid redundant else...ifs

polosanya (changes requested):
Let's create variables for all important values to make your code cleaner and more understandable

agniya-i (approved):

PR #2397, opened 2023-02-22
Solution

NadiiaR:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let cost = 0;
+
+  if (days >= 7) {
+    cost = (days * 40) - 50;
```
OleziO, line 35: Move `50` to a const variable.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+
+    return cost;
+  } else if (days >= 3) {
+    cost = (days * 40) - 20;
```
OleziO, line 39: Move `20` to a const variable.

src/calculateRentalCost.js
OleziO, line 35: Move `40` to a const variable.

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
-  // write code here
+  const priceD = 40;
+  const bigSale = 50;
+  const smallSale = 20;
+  let cost = 0;
+
+  (days >= 7) ? cost = (days * priceD) - bigSale
+    : (days >= 3) ? cost = (days * priceD) - smallSale
+      : cost = days * priceD;
+
```
OleziO, line 39: Don't use nested ternary operators, it's bad practice. Rewrite with `if / else` construction.

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const bigSale = 50;
+  const smallSale = 20;
+  let cost = 0;
+
+  if (days >= 7) {
+    cost = (days * price) - bigSale;
+
```
OleziO, line 38: Delete unnecessary empty lines. Check other similar cases.

OleziO (approved):

PR #2398, opened 2023-02-23
Solution

Istenriar:
Hi! Istenriar is my nick. The name on the platform is Mariia Klymchuk.

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+
+  if (days < 3) {
+    total = 40 * days;
+  } else if (days < 7) {
+    total = 40 * days - 20;
+  } else {
+    total = 40 * days - 50;
+  }
+
+  return (total);
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,18 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  const BASIC_DAY_CHARGE = 40;
+  const BASIC_DAY_CHARGE = 40 * days;
   const SEVEN_DAY_DISCOUNT = 50;
   const THREE_DAY_DISCOUNT = 20;
 
   let total = 0;
 
   if (days < 3) {
-    total = BASIC_DAY_CHARGE * days;
+    total = BASIC_DAY_CHARGE;
   } else if (days < 7) {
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let total = 0;
+
+  if (days < 3) {
+    total = 40 * days;
+  } else if (days < 7) {
+    total = 40 * days - 20;
+  } else {
+    total = 40 * days - 50;
+  }
```
DarkMistyRoom, line 39: Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

DarkMistyRoom (changes requested):

Istenriar:
I have changed the numbers for variables

Istenriar (commented):
Hi! I wonder why you didn't accept my changes? I have replaced the numbers with meaningful variables. Although, I didn't do that for days - as considered this redundant. There's no sense in writing that THREE_DAYS = 3 as this should be understandable from the context. 

Hope to hear from you soon
Thank you

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const BASIC_DAY_CHARGE = 40;
+  const SEVEN_DAY_DISCOUNT = 50;
+  const THREE_DAY_DISCOUNT = 20;
+
+  let total = 0;
+
+  if (days < 3) {
+    total = BASIC_DAY_CHARGE * days;
```
OleziO, line 38: Don't repeat this code - `BASIC_DAY_CHARGE * days`, just move it to a variable.

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < 3) {
+    total = BASIC_DAY_CHARGE;
+  } else if (days < 7) {
+    total = BASIC_DAY_CHARGE - THREE_DAY_DISCOUNT;
+  } else {
+    total = BASIC_DAY_CHARGE - SEVEN_DAY_DISCOUNT;
+  }
+
+  return (total);
```
tetiana-ivanchenko, line 45:
```suggestion
  if (days < 3) {
    return BASIC_DAY_CHARGE;
  } else if (days < 7) {
    return BASIC_DAY_CHARGE - THREE_DAY_DISCOUNT;
  } 
  
  return BASIC_DAY_CHARGE - SEVEN_DAY_DISCOUNT;
```

tetiana-ivanchenko (changes requested):

PR #2403, opened 2023-02-26
Solution

92BS:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let price;
+  const rent = 40;
+
+  if (days >= 1) {
+    price = days * rent;
```
TarasHoliuk, line 35: set this as the initial price value - that will allow you to get rid of this if statement (and just subtract discount in other if blocks)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  }
+
+  if (days >= 3) {
+    price = (days * rent) - 20;
```
TarasHoliuk, line 39: not critical, but you can move 20 and 50 to variables with meaningful names to make it clearer

TarasHoliuk (changes requested):

agniya-i (approved):

PR #2405, opened 2023-02-27
complete task

VladGraund:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = days * 40;
```
mykhalenych, line 31: don't use magic numbers create constants for 40, 20, 50 ...etc and use them istead

mykhalenych (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  const midSale = 20;
+  const bigSale = 50;
+
+  if (days >= 7) {
+    return days * cost - bigSale;
```
DarkMistyRoom, line 36: calculate the total before the loop to avoid duplicating

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days > 2 && days <= 6) {
```
DarkMistyRoom, line 37: better use 2 if's instead of else if here

DarkMistyRoom, line 37:
```suggestion
  } else if (days > 2) {
```

DarkMistyRoom (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const cost = 40;
+  const midSale = 20;
+  const bigSale = 50;
+  const lowDayPrice = days * cost;
+  const midDayPrice = lowDayPrice - midSale;
+  const bigDayPrice = lowDayPrice - bigSale;
+
+  if (days >= 7) {
+    return bigDayPrice;
+  }
```
etojeDenys, line 40: make calculation only if you need it
```suggestion
  if (days >= 7) {
    return lowDayPrice - bigSale;
  }
```

etojeDenys (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  const midSale = 20;
+  const bigSale = 50;
+  const Price = days * cost;
+
```
IvanFesenko, line 35: Why price word is capitalized?

IvanFesenko (approved):

PR #2406, opened 2023-02-27
solution

YuriiHlushenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = 40 * days;
+
+  if (days < 3) {
```
mykhalenych, line 33: don't use magic numbers, use constants instead

mykhalenych (changes requested):

mykhalenych (changes requested):
Screenshot 2023-02-27 at 17 13 32

you created contsnts but don't use them.
also 3 and 7 still are magic numbers

mykhalenych (approved):

PR #2408, opened 2023-02-27
Task was done.

anton-chornobai:


src/calculateRentalCost.js
```diff
@@ -7,7 +7,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = days * 40;
```
mykhalenych, line 10: don't use magic numbers, use constants instead

mykhalenych (changes requested):

src/calculateRentalCost.js
anton-chornobai: I cant use const because sum might change, depends on if if statement is true. (or maybe i didnt get what you wanted me to do..)

slavanosatskyi: The idea is to move number 40 to constants - for example const pricePerDay = 40

slavanosatskyi (changes requested):
Check my comment please

src/calculateRentalCost.js
```diff
@@ -7,7 +7,16 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  let sum = days * pricePerDay;
+
+  if (days >= 7) {
+    sum -= 50;
+  } else if (days >= 3) {
+    sum -= 20;
+  }
```
etojeDenys, line 17:
```suggestion
  if (days >= 7) {
    sum -= 50;
  } 
  
    if (days >= 3) {
    sum -= 20;
  }
```

etojeDenys (changes requested):
do the same with 50 and 20 (move it into the variables (e.g. bigDiscount\smallDiscount))

etojeDenys (approved):
LGTM

VitaliyBondarenko1982 (approved):

PR #2409, opened 2023-02-27
solution

vlad-fil:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 3) {
+    discount = 20;
+  };
+
+  if (days >= 7) {
+    discount = 50;
+  };
+
+  const total = days * 40 - discount;
```
OlEzhA0, line 41: what is `20`, `50` and `40`? please add meaningful constants

OlEzhA0 (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let discount = 0;
```
OleziO, line 31: Make 2 discount constants from this variable.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const rent = 40;
+
+  if (days >= 3) {
+    discount = 20;
+  };
+
+  if (days >= 7) {
```
OleziO, line 38: Move it above the 3 days discount.

OleziO: Combine all `if` to one `if else` construction

OleziO (changes requested):

tetiana-ivanchenko (approved):

PR #2412, opened 2023-02-28
Solution

yurii-shkrobut-m:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
+  let result = 0;
+
+  if (days >= 1 && days < 3) {
+    result = total;
+  } else if (days >= 3 && days < 7) {
+    result = total - 20;
```
OleziO, line 37: Move 20 and 50 to discount variables.

OleziO: Better to move 7 days if above the 3 days if. 
If you do this you can replace `(days >= 3 && days < 7) ` to `(days >= 3)`

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return total - bigDiscount;
+  }
+
+  if (days >= 3) {
+    return total - smallDiscount;
+  }
+
+  if (days < 3) {
+    return total;
+  }
```
anstsot, line 46: it would be better to return the value without additional `if`
```suggestion
  return total;
```

anstsot (approved):
Great job, approved!

PR #2413, opened 2023-02-28
Solution

yarik-bachynskyi:


src/calculateRentalCost.js
```diff
@@ -28,6 +28,21 @@
...
+  const rent = 40;
+  const sum = (rent * days);
+  const discountForSeven = 50;
+  const discountForThree = 20;
+
+  switch (true) {
+    case days >= 7:
+      return sum - discountForSeven;
+
+    case days < 3:
```
DarkMistyRoom, line 40: also would be great to move 7 and 3 to variables too)

DarkMistyRoom (approved):

PR #2414, opened 2023-03-01
Develop

domkab:
solution


src/calculateRentalCost.js
```diff
@@ -27,8 +27,25 @@
...
+calculateRentalCost(7);
+
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40;
+  let price = 0;
+
+  for (let i = 0; i < days; i++) {
+    price = price + rentalCost;
+  }
```
polosanya, line 38:
```suggestion
  const rentalCost = 40 * days;
```
Do you really need this loop? 
Can't you just multiply days with 1 day price?
By the way, would be great to create a variable for `40` to make your code cleaner

src/calculateRentalCost.js
```diff
@@ -27,8 +27,25 @@
...
+
+  if (days >= 3 && days < 7) {
+    price = price - 20;
+  }
```
```suggestion
  if (days >= 3 && days < 7) {
    return rentalPrice - 20;
  }
```
You can just return the full price, no need to store it

polosanya (changes requested):
Please create variables for all important values (40, 20, 50 etc)

src/calculateRentalCost.js
```diff
@@ -27,8 +27,21 @@
...
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return rentalCost - threeDayDiscount;
+  } else if (days >= 7) {
+    return rentalCost - sevenDayDiscount;
+  } else {
+    return rentalCost;
+  }
```
Volodymyr-Melnyk-Fron10der, line 44: apply early return pattern here
```suggestion
  if (days >= 3 && days < 7) {
    return rentalCost - threeDayDiscount;
  } 
  
  if (days >= 7) {
    return rentalCost - sevenDayDiscount;
  } 
  
  return rentalCost;
```

src/calculateRentalCost.js
```diff
@@ -27,8 +27,21 @@
...
  */
 
+calculateRentalCost(7);
+
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const rentalCost = days * dayPrice;
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
```
Volodymyr-Melnyk-Fron10der, line 36: keep constants around constants
```suggestion
  const dayPrice = 40;
  const threeDayDiscount = 20;
  const sevenDayDiscount = 50;
  
  const rentalCost = days * dayPrice;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, almost there!

qoa11a (approved):

PR #2417, opened 2023-03-01
Solution

yao157:


src/calculateRentalCost.js
```diff
@@ -1,34 +1,44 @@
...
+
+function calculateRentalCost(days) {
+  const value = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return value - 20;
+  }
+
+  if (days >= 7) {
+    return value - 50;
```
vlad-demchuk, line 38: move 40, 20, 50 to variables with according names

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #2418, opened 2023-03-02
solution

Vaskovics:


anstsot (changes requested):
Please take a look at different cases in your taks
image

Vaskovics:
If can give explain me which other? There is =>7, =>3 and <3, any other?

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
```
Volodymyr-Melnyk-Fron10der, line 31: always remove comments

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const total = days * 40;
```
Volodymyr-Melnyk-Fron10der, line 34: I would move 40 to the constant as well

Volodymyr-Melnyk-Fron10der (approved):
Good job👍

PR #2419, opened 2023-03-02
add solution

yandapanda19:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
```
slavanosatskyi, line 31: please, move all magic numbers (40, 20, 50) to constants

slavanosatskyi (changes requested):
nice, though small fix is needed

PR #2420, opened 2023-03-02
Solution

vpanch2:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentalCost = 40 * days;
```
polosanya, line 31: What is 40? Let's create a variable for that value to make your code more descriptive

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days >= 7) {
+    rentalCost -= 50;
+  } else if (days >= 3) {
+    rentalCost -= 20;
+  }
```
```suggestion
  if (days >= 7) {
    return rentalCost - 50;
  }
  
  if (days >= 3) {
    return rentalCost - 20;
  }
```
You can just immediately return the final price to avoid "else if" and make your code a bit cleaner

polosanya (changes requested):
Let's create variable for all the improtant values (40, 20, 50 etc.)
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40;
+  const smallDiscount = 20;
+  const BigDiscount = 50;
```
etojeDenys, line 33: rename it in camelCase (bigDiscount)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  let totalCost = 0;
+
+  if (days < 3) {
+    totalCost = rentalCost * days;
+  } else if (days >= 7) {
+    totalCost = (rentalCost * days) - BigDiscount;
+  } else if (days >= 3) {
+    totalCost = (rentalCost * days) - smallDiscount;
+  }
+
```
```suggestion
  let totalCost = days * rentalCost;

  if (days >= 7) {
    return totalCost * bigDiscount;
  }
  
  if (days >= 3) {
    return totalCost * smallDiscount;
  } 
  
  return totalCost
  

```

etojeDenys (changes requested):
always try to optimize your code

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discount = [20, 50];
```
agniya-i, line 31:
```suggestion
  const discounts = [20, 50];
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const dayCost = 40;
+  let totalCost = 0;
+
+  if (days < 3) {
+    totalCost = dayCost * days;
+  } else if (days >= 3 && days < 7) {
+    totalCost = (dayCost * days) - discount[0];
+  } else {
+    totalCost = (dayCost * days) - discount[1];
+  }
```
agniya-i, line 41: please pay attention  to this previous comment 🙂
Screenshot 2023-03-05 at 12 37 37

you can return value right in `if` block and get rid of extra `else` block

agniya-i (changes requested):

Volodymyr-Melnyk-Fron10der (changes requested):
previous comments were not fixed (

PR #2422, opened 2023-03-03
Solution

HannaVasylieva:


loralevitska (changes requested):
Good job! There are things that should be fixed:

Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content. (20,50)

VitaliyBondarenko1982 (approved):

PR #2425, opened 2023-03-04
Rental Cost

oleh-dehtiarenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  result = days * 40;
+
+  if (days >= 3) {
+    result = days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    result = days * 40 - 50;
+  }
```
Volodymyr-Melnyk-Fron10der, line 41: a lot of useless actions
```suggestion
  let result = days * 40;

  if (days >= 3) {
    result -= 20;
  }

  if (days >= 7) {
    result -= 50;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  result = days * 40;
```
Volodymyr-Melnyk-Fron10der, line 33: move 40, 20, 50 into constants

Volodymyr-Melnyk-Fron10der (changes requested):
almost there

SanyaBratashchuk (approved):

PR #2429, opened 2023-03-06
Solution

LBondarchuk:


witflash (approved):
Good job.
But for the future, any static numbers (20, 40, 50) are better to move to separate constants and use these constants instead of direct values.

PR #2430, opened 2023-03-07
add task solution

Daria13k:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const carRent = 40;
+  const basicCost = carRent * days;
+  let totalCost = basicCost;
+
+  if (days >= 3 && days < 7) {
```
mykhalenych, line 37: move 3 and 7 to variable as well

mykhalenych (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days >= smallWeekend && days <= bigWeekend) {
+    totalCost = basicCost - smallDiscount;
+  };
+
+  if (days >= bigWeekend) {
+    totalCost = basicCost - bigDiscount;
+  };
+
+  return totalCost;
```
IvanFesenko, line 47: Actually you dont need totalCost  - you can return value as well
```suggestion

  if (days >= smallWeekend && days <= bigWeekend) {
    return basicCost - smallDiscount;
  };

  if (days >= bigWeekend) {
    return basicCost - bigDiscount;
  };

  return basicCost;
```

IvanFesenko (approved):

PR #2432, opened 2023-03-07
Solution

pindel-kateryna:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+
+  if (days >= 7) {
+    sum = days * price - 50;
+  } else if (days >= 3) {
+    sum = days * price - 20;
+  } else if (days < 3) {
+    sum = days * price;
+  }
+
+  return sum;
```
```suggestion
  if (days >= 7) {
    return  totalPrice - 50;
  } 
   
  if (days >= 3) {
    return totalPrice - 20;
  }

  return totalPrice;
```

etojeDenys (changes requested):
create `totalCost` variable  (days*price) and move 50/20 to some variables as well

src/calculateRentalCost.js
VitaliyBondarenko1982: not fixed comment about avoiding else/if else .
please refactor it.

VitaliyBondarenko1982 (changes requested):
Comment about avoiding `if/else` if  not  fixed.
Please, refactor this point

etojeDenys (approved):
looks awesome

PR #2437, opened 2023-03-09
create function for calculating rent cost

Esteban4769:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let discount = 0;
+
+  if (days >= 3) {
```
roshirin, line 33: Don't hardcode values, use a variable here and in other similar cases (checklist #1)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let discount = 0;
```
roshirin, line 31: Avoid using 'let' and reassigning variables values (checklist #2)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3) {
+    discount = 20;
+  }
```
roshirin, line 35: You should return the value immediately in `if` statements (checklist #3)

roshirin (changes requested):
Take a look at the recommendations in the checklist

VinnieJ-2k20 (changes requested):
Solution looks good, but may be improved.

```
To make code more descriptive, you can add variables to 3, 7, 20, 50 values.
You calculate `days * costPerDay` several times. Do it once and add variable.
```

VinnieJ-2k20 (approved):
Good job 👍

PR #2438, opened 2023-03-09
Solution

Sergey8793:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days < 3) {
+    total = days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    total = (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    total = (days * 40) - 50;
```
vlad-demchuk, line 42: move `40, 20, 50` to variables as well

vlad-demchuk (changes requested):
.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+  if (days >= 3 && days < 7) {
+    total = (days * priceDay) - discountThreeDays;
+  }
+
+  if (days >= 7) {
+    total = (days * priceDay) - discountFiveDays;
+  }
+
+  return total;
```
etojeDenys, line 48:
```suggestion

if (days >= 7) {
    return totalCost - discountFiveDays;
  }

  if (days >= 3) {
   return totalCost - discountThreeDays;
  }

  return totalCost;

  

  return total;
```

etojeDenys (changes requested):

etojeDenys (approved):
awesome 🤩

PR #2439, opened 2023-03-09
Solution

asmd564:


Esteban4769 (approved):

VinnieJ-2k20 (changes requested):
Solution looks good, but may be improved.

1. To make code more descriptive, you can add variables to 3, 7, 20, 50 values.
2. You calculate `cost * days` several times. Do it once and add variable.

VinnieJ-2k20 (changes requested):
Item 2 from the comment not fixed.

VinnieJ-2k20 (approved):

PR #2440, opened 2023-03-10
calculateRentalCost task solution

PavloReutskiy:


vinnykalisa (approved):
Good job! You can also simplify your code by getting rid of else-if constructions and using only if(s)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return (basicDiscount * days - 50);
+  }
+
+  if (days >= 3) {
+    return (basicDiscount * days - 20);
+  }
+
+  return (basicDiscount * days);
```
jstmpelowycz, line 41: you repeat yourself, let's use one more variable to calculate total cost per days

jstmpelowycz (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days >= 7) {
+    return (totalAmount - 50);
+  }
+
+  if (days >= 3) {
+    return (totalAmount - 20);
+  }
+
+  return (totalAmount);
```
jstmpelowycz, line 42: 2 points:
- you don't need parentheses in return
- let's create variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDiscount = 40;
```
jstmpelowycz, line 31: the value does not refer to any discount, let's rename it

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #2442, opened 2023-03-10
Solution

KosolapovAnatolii:


Anastasia2403 (approved):
Good job 👍 
You can also shorten your code using the _ternary operator_ . Or _else if & else_ to remove the writing of the last condition.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return days * 40 - 20;
+  }
 
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
```
VinnieJ-2k20, line 41: Condition is redundant. If `days < 7` is false, `days >= 7` is alsways true.

VinnieJ-2k20 (changes requested):
Check the comment.

VinnieJ-2k20 (changes requested):
1. You calculate `days * rentPrice` three times. Do it once and use variable.
2. If you return smth from `if` block - you don't need `else` statement (see checklist)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const rentPrice = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const prod = days * rentPrice;
```
VinnieJ-2k20, line 34: not the best choice for variable name

VinnieJ-2k20 (approved):

PR #2446, opened 2023-03-11
Done

Ilia991:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return 40 * days;
+  } else if (days < 7) {
+    return 40 * days - 20;
```
OleziO, line 34: Move price and discount to variables.

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  const costs = 40;
+  const discount = days < 7 ? 20 : 50;
+
+  if (days < 3) {
+    return 40 * days;
+  } else if (days < 7) {
+    return costs * days - discount;
+  } else {
+    return costs * days - discount;
+  }
```
tetiana-ivanchenko, line 40:
```suggestion
  if (days < 3) {
    return 40 * days;
  }
  
  if (days < 7) {
    return costs * days - discount;
  }
  
  return costs * days - discount;
```

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const costs = 40;
+  const discount = days < 7 ? 20 : 50;
+
+  if (days < 3) {
+    return costs * days;
+  }
+  
+  if (days < 7) {
+    return costs * days - discount;
+  }
```
OleziO, line 40: You dont need this `if`, its the same as last return.

OleziO (approved):

PR #2450, opened 2023-03-12
solution

dmytrobetsan:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
-  // write code here
+  const dailyPrice = 40;
+  let totalPrice = dailyPrice * days;
+
+  if (days >= 7) {
+    totalPrice = totalPrice - 50;
+  }
+
+  if (days >= 3 && days < 7) {
+    totalPrice = totalPrice - 20;
```
VitaliyBondarenko1982, line 39: make variables for all 'magic' numbers

VitaliyBondarenko1982 (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 7) {
+    totalPrice = totalPrice - bigDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
+    totalPrice = totalPrice - midDiscount;
+  }
+
+  return totalPrice;
```
etojeDenys, line 45:
```suggestion

  if (days >= 7) {
    return totalPrice - bigDiscount;
  }

  if (days >= 3) {
    return totalPrice - midDiscount;
  }

  return totalPrice;
```

etojeDenys (changes requested):
almost done

IvanFesenko (approved):

PR #2451, opened 2023-03-12
solution

kyserf:


vksazonov (approved):
Well done!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
```
VinnieJ-2k20, line 35: Redundant condition. If `days < 3` is false, `days >= 3` is always true.

VinnieJ-2k20 (changes requested):
Check the comments.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const carPrice = 40;
+  const maxDiscount = 50;
+  const discount = 20;
+  const lessDays = 3;
+  const moreDays = 7;
+
+  if (days >= moreDays) {
+    return days * carPrice - maxDiscount;
```
VinnieJ-2k20, line 38: `days * carPrice` is calculated three times. Do it once, and use variable for this.

VinnieJ-2k20 (changes requested):
Last edit.

VinnieJ-2k20 (approved):

PR #2453, opened 2023-03-12
Solution

Pozdnya:


Vlad-Lischenovskiy (approved):
Good idea to do this task

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days >= 3 && days <= 6) {
```
jstmpelowycz, line 35: here, days >= 3 is always `true`, because the previous condition is already omitted
also, what's 6? let's use given values only

jstmpelowycz (changes requested):
see the checklist, let's create variables

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #2454, opened 2023-03-13
solution

glots-a:


igorkruz (approved):

igorkruz (approved):
good job 👍

VinnieJ-2k20 (changes requested):
See checklist item 2.
You don't need to reassign variable to just return it in the next line.
Also, add variables to 40, 3, 7, 20, 50 values.

VinnieJ-2k20 (approved):
Good job 👍

PR #2456, opened 2023-03-13
Solution

AnyaLyaLya:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalAmount = days * 40;
+  const basicDiscount = 20;
+  const superDiscount = 50;
+
+  if (days >= 3 && days < 7) {
```
BondarenkoDmitriy, line 35: why are there conditions days < 7, this will work on the next else if

BondarenkoDmitriy (changes requested):

BondarenkoDmitriy (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const totalAmount = days * 40;
+  const basicDiscount = 20;
+  const superDiscount = 50;
+
+  switch (true) {
+    case (days >= 7):
+      return (totalAmount - superDiscount);
+    case (days >= 3):
+      return (totalAmount - basicDiscount);
+  }
```
jstmpelowycz, line 40: switch-case-statement is not for this purpose, let's use simple if-statements

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const totalAmount = days * 40;
+  const basicDiscount = 20;
+  const superDiscount = 50;
+
+  switch (true) {
+    case (days >= 7):
+      return (totalAmount - superDiscount);
+    case (days >= 3):
```
jstmpelowycz, line 38: you can use variables for 3, 7 as well

jstmpelowycz (changes requested):

jstmpelowycz (approved):
great! also, I reflected upon using switch-case — agree, it may make code more readable, so you can use it, but if only it really applies

PR #2458, opened 2023-03-13
calculate rent

Vlad-Lischenovskiy:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let amount = days * 40;
+
+  if (days >= 7) {
+    amount -= 50;
```
yuliia-anisimova, line 34: According to the checklist, you should return the value immediately in if statements.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days >= 3) {
```
yuliia-anisimova, line 35: It's better not to use the 'else if' in your code.

yuliia-anisimova (changes requested):
Great job! Please, check my comments and change your code according to the checklist.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const amount = days * 40;
+
+  if (days >= 7) {
+    return amount - 50;
+  }
+
+  if (days >= 3) {
+    return amount - 20;
+  }
```
jstmpelowycz, line 39: what are 3, 7, 20, ...? let's use variables

jstmpelowycz (changes requested):
awesome! almost done

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const amount = days * price;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const bigDay = 7;
+  const smallDay = 3;
```
jstmpelowycz, line 36: not the best way to name these variables

jstmpelowycz (approved):
good job!

PR #2459, opened 2023-03-13
Solution

igorkruz:


danylo-dobrovolskyi (approved):
Everything is fine, but it is better to put 'days * 40' in a new variable

igorkruz:
> Everything is fine, but it is better to put 'days * 40' in a new variable

fixed )

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const total = days * 40;
+
+  if (days >= 3 && days < 7) {
+    return total - 20;
+  }
+
+  if (days >= 7) {
+    return total - 50;
+  }
```
VinnieJ-2k20, line 39: Swap these to remove redundant condition.

VinnieJ-2k20 (changes requested):
Check the comment.

VinnieJ-2k20 (approved):
Good job 👍

PR #2464, opened 2023-03-13
create func calculateRentalCost

VenomMB:


AP-Solution (approved):
Good job, it works!

VinnieJ-2k20 (approved):
Good job 👍 
Could also add variables for 3 and 7

PR #2467, opened 2023-03-13
solution

Olekk17:


ymys1 (approved):

VinnieJ-2k20 (changes requested):
Check the comment.
Also, you calculate `days * basicDayPrice` three times. Do it once and introduce variable.

VinnieJ-2k20 (approved):
Good job 👍 
Can avoid adding (...) to return values.

PR #2468, opened 2023-03-13
Solution

Zad0rnik:


PashaMalyshkin (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const lowerDiscount = 20;
+  const basicDiscount = 50;
```
jstmpelowycz, line 33:
```suggestion
  const basicDiscount = 20;
  const greatDiscount = 50;
```

jstmpelowycz (approved):
good job! you could've created variables for 3, 7 as well

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const sum = days * dayPrice;
+
+  if (days >= 3 && days < 7) {
+    return sum - lowerDiscount;
+  }
+
+  if (days >= 7) {
+    return sum - basicDiscount;
+  }
```
jstmpelowycz, line 42: if you swap these conditions, you won't need `&&`

PR #2469, opened 2023-03-13
solution

AP-Solution:


Anna-Kuzmenko (approved):
you can change the last "case" to "default"

src/calculateRentalCost.js
```diff
@@ -28,7 +28,11 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  switch (true) {
+    case days < 3: return days * 40;
+    case days < 7: return days * 40 - 20;
+    case days >= 7: return days * 40 - 50;
```
jstmpelowycz, line 34:
```suggestion
    case days < 3: 
      return days * 40;
    case days < 7: 
      return days * 40 - 20;
    case days >= 7: 
      return days * 40 - 50;
```

jstmpelowycz (changes requested):
what are 3, 7, 20, ...? let's use variables
also, you repeat yourself many times, so let's use variables as well

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const rentCost = 40;
+
+  switch (true) {
```
jstmpelowycz, line 35: we agreed to avoid using switch-case in such cases

jstmpelowycz (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const bigDiscount = 50;
+  const rentCost = 40;
+
+  if (days < 3) {
+    return days * rentCost;
+  } else if (days >= 3 && days < 7) {
+    return days * rentCost - smallDiscount;
+  } else if (days >= 7) {
+    return days * rentCost - bigDiscount;
+  }
```
jstmpelowycz, line 41: the checklist, p.3

jstmpelowycz (changes requested):

liliia-livandovskaya (approved):

PR #2470, opened 2023-03-13
Solution

Goto-Jesus:


Evgenb000 (approved):
Good job

jstmpelowycz (changes requested):
see the checklists, those items about variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRent = 40;
+  const basicDiscount = 20;
+  const weekDiscount = 50;
+  const theWholeAmount = days * dailyRent;
+
+  switch (true) {
```
jstmpelowycz, line 36: we've agreed to avoid using switch-case in such cases
https://mateacademy.slack.com/archives/C04N8TU8PN2/p1678794145933229

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #2471, opened 2023-03-13
Solution

Kavon-Skela:


VladislavaT (approved):
Good job! 
u can also create vars for discount but I think it`s not necessary

Kavon-Skela:
Thanks for reviewing) Variables added

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPayment = 40;
+  const threeOrMoreDaysDisc = 20;
+  const sevenOrMoreDaysDisc = 50;
+  const threeDays = 3;
+  const sevenDays = 7;
```
jstmpelowycz, line 35: see the checklist, p.5

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  switch (true) {
+    case days < threeDays:
+      return days * dayPayment;
+
+    case days < sevenDays:
+      return days * dayPayment - threeOrMoreDaysDisc;
+
+    case days >= sevenDays:
+      return days * dayPayment - sevenOrMoreDaysDisc;
```
jstmpelowycz, line 45: DRY, let's calculated total cost per given days beforehand

jstmpelowycz (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPayment = 40;
+  const basicDiscount = 50;
+  const minDiscount = 20;
+  const totalCost = days * dayPayment;
+
+  switch (true) {
```
jstmpelowycz, line 36: we agreed to avoid using switch-case in such cases

jstmpelowycz (changes requested):

src/calculateRentalCost.js
Kavon-Skela: Replaced

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return totalCost;
+  }
+
+  if (days < 7) {
+    return totalCost - minDiscount;
+  }
+
+  if (days >= 7) {
+    return totalCost - basicDiscount;
+  }
```
jstmpelowycz, line 46: this condition is always true, you don't need it

jstmpelowycz (changes requested):

jstmpelowycz (approved):
good job!

PR #2474, opened 2023-03-13
decision

Nazarii98:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let result = 40 * days;
+
+  if (days > 6) {
+    result -= 50;
+  } else if (days > 2) {
+    result -= 20;
+  }
```
Goto-Jesus, line 38: You should use "switch" operator, because if you want to support your code by adding more conditions. For example next step you need to add a condition discount if you have 10 days your discount is -100$. Your code will look like this: if else {} if else {}if else {}if else {}if else {}... and your first part of the code always needs to change if => else if...

Goto-Jesus (changes requested):

VinnieJ-2k20 (changes requested):
Previous option was closer to the truth.
Use `if` with returns.

VinnieJ-2k20 (changes requested):
1. See the checklist regarding `if/else` and `return` statements
2. Add variables for meaningful values.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
 function calculateRentalCost(days) {
   // write code here
+  let cost = 40 * days;
+  const saleSmall = 20;
+  const saleBig = 50;
+
+  if (days > 6) {
+    cost -= saleBig;
+
+    return cost;
```
VinnieJ-2k20, line 39: You can just return the value.
```suggestion
    return cost - saleBig;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
+  }
+
+  if (days > 2) {
+    cost -= saleSmall;
+
+    return cost;
```
VinnieJ-2k20, line 45: Same here.

VinnieJ-2k20 (approved):
God job 👍 
Check the comments.

PR #2476, opened 2023-03-13
Solution

Anna-Kuzmenko:


Rage1502 (approved):
Good job)!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  switch (true) {
+    case (days < firstDiscountPeriod):
+      return days * costPerDay;
+
+    case (days < secondDiscountPeriod):
+      return days * costPerDay - basicDiscount;
+
+    default:
+      return days * costPerDay - proDiscount;
```
jstmpelowycz, line 45: DRY, let's calculate total cost per days beforehand

jstmpelowycz (changes requested):
good job! almost done

jstmpelowycz (changes requested):
we agreed not to use switch-case for such cases
https://mateacademy.slack.com/archives/C04N8TU8PN2/p1678794145933229

Anna-Kuzmenko:
I think it is the hardest js task )))

jstmpelowycz (approved):

PR #2478, opened 2023-03-13
task solution

VladislavaT:


TanyaS133 (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDays = 3;
+  const weekDays = 7;
+  const dayRent = 40;
```
jstmpelowycz, line 33:
```suggestion
  const dailyRent = 40;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDays = 3;
+  const weekDays = 7;
```
jstmpelowycz, line 32: these names do not fully refer to what the variables contain, take care of it next times

jstmpelowycz (approved):
good job!

PR #2479, opened 2023-03-13
rental cost calculator

ofenchuk:


didyk-ma (changes requested):
Good solution!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,11 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = days * 40;
+  const discountBasic = price - 20;
+  const discountPlus = price - 50;
+
+  return days < 3 ? price : days < 7 ? discountBasic : discountPlus;
```
VinnieJ-2k20, line 35: Don't write nested ternaries, they are extremely hard to read.
Use `if` here

VinnieJ-2k20 (changes requested):
One edit needed.

liliia-livandovskaya (approved):

PR #2482, opened 2023-03-13
Solution

TanyaS133:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    sum = days * 40;
+  }
+
+  if (days >= 3) {
+    sum = days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    sum = days * 40 - 50;
+  }
```
MylaCrystal, line 43: In this case, when the function is running and the number is greater than = 7, the variable will first be counted in condition (days >= 3) and then overwritten in condition (days >= 7). Perhaps this process could be more optimized?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < 3) {
+    sum = days * 40;
+  }
+
+  if (days >= 3) {
+    sum = days * 40 - 20;
+  }
+
+  if (days >= 7) {
```
MylaCrystal, line 41: It will be nice if instead 40 you will use a variable (in the case if this price is changed, it will be easier to change it only in 1 place)

voronevska (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,21 +28,22 @@
...
-  let sum = 0;
+  const sum = days * 40;
+  let total = 0;
 
   if (days < 3) {
-    sum = days * 40;
+    total = sum;
   }
 
   if (days >= 3) {
```
MylaCrystal, line 38: I think here you could add a restriction in the condition so that it is greater than 3, but less than 7

voronevska (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const sum = days * 40;
+  let total = 0;
```
jstmpelowycz, line 32: the checklist, p.2

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days < 3) {
+    total = sum;
+  }
```
jstmpelowycz, line 36: you don't need it

jstmpelowycz (changes requested):
also, the checklist, p.1

G-YaSha (approved):

PR #2483, opened 2023-03-13
solve task calculateRentalCost

kukla1989:


jstmpelowycz (changes requested):
the checklist, p.1-3

kukla1989:
> the checklist, p.1-3

can you please describe what exactly i did not do properly according p.2(. [CODE STYLE] - Avoid reassigning variables values. It's better to use `const` instead of `let`.) I used only const in this task.

jstmpelowycz (changes requested):
the checklist, p.1
let's use variables, for 20, 40, 50 at least

AlexRog01 (approved):
Good

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discount = 20;
+  const maxDiscount = 50;
+  const rentalCostWithoutDiscount = days * 40;
+  const minDayForDiscount = 3;
+  const week = 7;
```
jstmpelowycz, line 35:
```suggestion
  const minDayForDiscount = 3;
  const maxDayForDiscount = 7;
```

jstmpelowycz (approved):

PR #2488, opened 2023-03-14
Solution

KseniaSkr:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  // const weeklyDiscount = 50;
+  // const discount = 20;
```
IvanFesenko, line 32: Use this variable instead magic numbers

IvanFesenko (changes requested):
Use variables

slavanosatskyi (approved):

PR #2493, opened 2023-03-15
Solution

maksym-kostetskyi:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
 
 function calculateRentalCost(days) {
   // write code here
-}
+  const sum = days * 40;
+
+  switch (true) {
+    case days > 0 && days < 3:
+      return sum;
```
loralevitska, line 36: Add empty line after `return` in case

src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
 
 function calculateRentalCost(days) {
   // write code here
-}
+  const sum = days * 40;
+
+  switch (true) {
+    case days > 0 && days < 3:
```
loralevitska, line 35: Let's move all magic numbers to constants with meaningful variable names

loralevitska (changes requested):
Almost done) Let's improve your code

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
 function calculateRentalCost(days) {
   // write code here
-}
+  const dayPrice = 40;
+  const sum = days * dayPrice;
+  const countDay0 = 0;
+  const countDay3 = 3;
+  const countDay7 = 7;
+
+  switch (true) {
```
OleziO, line 38: Better to do it with `if else` construction (`switch(true)` - bad code)

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+  const dayPrice = 40;
+  const sum = days * dayPrice;
+  const countDay0 = 0;
+  const countDay3 = 3;
+  const countDay7 = 7;
+
+  if (days > countDay0 && days < countDay3) {
+    return sum;
+  } else if (days >= countDay3 && days < countDay7) {
+    return sum - 20;
```
slavanosatskyi, line 41: this is considered as magic numbers we well. Try to come up with meaningful name for it. It can be shortStayDiscount or something like this

slavanosatskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+  const shortRentDiscount = 20;
+  const longRentDiscount = 50;
+
+  if (days > countDay0 && days < countDay3) {
+    return sum;
+  } else if (days >= countDay3 && days < countDay7) {
+    return sum - shortRentDiscount;
+  } else if (days >= countDay7) {
+    return sum - longRentDiscount;
+  }
```
loralevitska, line 46: Avoid `if / else` here. After return below code is unreachable, so it can be simplify. Also for future to improve your knowledge about [if-else hell](https://www.linkedin.com/pulse/avoid-if-else-hell-javascript-afsal-m-h/)
```suggestion
  if (days > countDay0 && days < countDay3) {
    return sum;
  } 
  
  if (days >= countDay3 && days < countDay7) {
    return sum - shortRentDiscount;
  }
  
 return sum - longRentDiscount;
```

loralevitska (approved):

PR #2494, opened 2023-03-15
first try

Nikolaymelni:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const discountSmall = 20;
+  const discountBig = 50;
+  let costRent = 40 * days;
+
+  if (days < 7 && days >= 3) {
+    costRent -= discountSmall;
+  } else if (days >= 7) {
+    costRent -= discountBig;
+  }
```
```suggestion
  if (days < 7 && days >= 3) {
    return costRent - discountSmall;
  }
  
  if (days >= 7) {
    return costRent - discountBig;
  }
```
You can just immediately return the full price to make your code a bit cleaner 
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discountSmall = 20;
+  const discountBig = 50;
+  let costRent = 40 * days;
```
polosanya, line 33: It would be great to create a variable for 40 too.
Also, you can make it a const now

polosanya (changes requested):
Almost perfect!

qoa11a (approved):

PR #2496, opened 2023-03-15
new

NataliiaRodoman:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
```
OleziO, line 31: Remove comment

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+  let totalDay = 0;
+  const priseDay = 40;
```
OleziO, line 33: You don't need this variable, just return the calculation.

OleziO, line 33: Rename to `price`

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+
+  if (days < 3) {
```
OleziO, line 35: You do not need this `if`

OleziO: Just move the return calculation to the end of the function.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+    totalDay = priseDay * days;
+  };
+
+  if ((days >= 3) && (days < 7)) {
+    totalDay = priseDay * days - 20;
+  };
+
+  if (days >= 7) {
```
OleziO, line 43: Swap with the previous condition:

```
if (days >= 7) {
  return ...;
}
else if (days >= 3) {
 return ...;
}
return ...;
```

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+
+  if (days >= 7) {
+    return price * days - 50;
```
OleziO, line 34: Move discount to variables.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else {
```
OleziO, line 35: Replace `else` with `else if` to avoid unnecessary nesting.

[More info.](https://www.w3schools.com/js/js_if_else.asp)

OleziO (changes requested):

AlisaVlasova (approved):

PR #2497, opened 2023-03-15
Solution

theFallcon:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const sum = days * 40;
+
+  if (days >= 7) {
+    return sum - 50;
+  } else if (days >= 3) {
+    return sum - 20;
+  } else {
+    return sum;
+  }
```
tetiana-ivanchenko, line 39:
```suggestion
    return sum - 50;
  }
  
  if (days >= 3) {
    return sum - 20;
  }
  
  return sum;
```

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCoast = 40;
+  const sum = days * rentCoast;
+
+  if (days >= 7) {
+    return sum - 50;
+  } else if (days >= 3) {
+    return sum - 20;
```
sTorba24, line 37: 30 and 50 can be moved to the variable too

sTorba24 (approved):
GJ :) I left you a little comment

PR #2499, opened 2023-03-17
Solution

tkalindima:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days > 1 && days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  }
```
loralevitska, line 37: Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

```suggestion
  if (days > 1 && days < 3) {
    return days * 40;
  } 
  
  if (days >= 3 && days < 7) {
    return days * 40 - 20;
  }
  
  return days * 40 - 50;
```

loralevitska (changes requested):
Almost done) Let's improve

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days > 1 && days < 3) {
+    return days * rentalCost;
+  }
+
+  if (days >= 3 && days < 7) {
+    return days * rentalCost - basicDiscount;
+  }
+
+  return days * rentalCost - additionalDiscount;
```
```suggestion
  if (days >= 7) {
    return totalCost - additionalDiscount;
  }

  if (days >= 3) {
    return totalCost - basicDiscount;
  }

  return totalCost;
```

etojeDenys (changes requested):

loralevitska (approved):
🔥

PR #2502, opened 2023-03-18
solution

Fedos17:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const amount = days * 40;
```
YegorKorenkov, line 31: Also please try to add variables for all values
image

YegorKorenkov (changes requested):

vlad-demchuk (approved):

PR #2510, opened 2023-03-22
CalcRentalCost

Soi4An:


nazarmatsevych (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const amount = days * 40;
+  const amountForTheWeek = amount - 50;
```
YegorKorenkov, line 32: in the previous review's comment it was meant that it is better to put these number(20, 50)s into variables, form example discountWeek, discountThreeDays

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const amountIn3Days = amount - 20;
```
YegorKorenkov, line 33: Also try do not use number in variable name, it's bad practice

YegorKorenkov (changes requested):

OleziO (approved):

PR #2513, opened 2023-03-22
first commit

IllyaMyhalyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
+  if (days < 3) {
+    return rent * days;
+  } else
+  if (days >= 3 && days < 7) {
+    return (rent * days) - 20;
+  } else
+  if (days >= 7) {
+    return (rent * days) - 50;
+  }
```
Volodymyr-Melnyk-Fron10der, line 41: use early return pattern
```suggestion
  if (days < 3) {
    return rent * days;
  } 
  
  if (days >= 3 && days < 7) {
    return (rent * days) - 20;
  } 
  
  return (rent * days) - 50;

```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const rent = 40;
 
+  if (days < 3) {
+    return rent * days;
```
Volodymyr-Melnyk-Fron10der, line 34: why do not move this operation into variable? you are repeating it 3 times

Volodymyr-Melnyk-Fron10der (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  } else
+  if (days >= 3 && days < 7) {
+    return (rent * days) - 20;
```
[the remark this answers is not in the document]
IllyaMyhalyk: thanks for the feedback!

TarasHoliuk (approved):
:+1:

PR #2516, opened 2023-03-25
Solution

Dka888:


MikiOOOO (approved):

michalcywka (changes requested):
Please don't commit `package-lock.json` or `package.json` files

Dka888:
I restored package-lock.json and package.json files from original repository.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = 40 * days;
```
michalcywka, line 31: It would be good to have `40` as a variable with descriptive name as well, so that we know what it means

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const baseDiscount = 20;
+  const wideDiscount = 50;
+
+  if (days >= 7) {
+    return (totalCost - wideDiscount);
+  }
+
+  if (days > 2) {
+    return (totalCost - baseDiscount);
+  }
```
d-wasilewski, line 42: Use either > or >= in both cases to be concise

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCosts = 40;
+  const totalCost = dayCosts * days;
+  const baseDiscount = 20;
+  const wideDiscount = 50;
```
d-wasilewski, line 34: In general, the term "wide discount" is not a commonly used term in the context of discounts or promotion, so it might not be clear, change it for "large" "major" or something like that

d-wasilewski (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCosts = 40;
+  const totalCost = dayCosts * days;
+  const baseDiscount = 20;
```
ivadimko, line 33: Consider using more descriptive variable names than `baseDiscount` and `largeDiscount`. For example, `discountForThreeDays` and `discountForSevenDays`. 🤔

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCosts = 40;
```
ivadimko, line 31: Consider using a more descriptive variable name than `dayCosts`. For example, `dailyRentalCost`. 🤔

d-wasilewski (approved):

PR #2518, opened 2023-03-26
solution

bjeziorek:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
 function calculateRentalCost(days) {
   // write code here
+  let discount = 0;
+
+  switch (true) {
+    case days >= 3 && days < 7:
+      discount = 20;
+      break;
+    case days >= 7:
+      discount = 50 * (Math.floor(days / 7));
```
MikiOOOO, line 39: If you rent the car for 7 or more days, you get $50 off your total.

MikiOOOO (changes requested):

MikiOOOO (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let discount = 0;
+  const discountLevel1 = 20;
+  const discountLevel2 = 50;
+  const daysAmount = days;
+
+  switch (true) {
```
michalcywka, line 37: `switch` is usually used when there is a limited set of specific values that we want to compare to. Here, we have range of days, it's better to use `if`.

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let discount = 0;
+  const discountLevel1 = 20;
+  const discountLevel2 = 50;
+  const daysAmount = days;
+
+  if (daysAmount >= 3) {
+    discount = discountLevel1;
```
michalcywka, line 38: Checklist point 3, try to use return in `if` statements

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
   // write code here
+  const discountLevel1 = 20;
+  const discountLevel2 = 50;
+  const daysAmount = days;
+  const cost = daysAmount * 40;
+
+  if (daysAmount >= 3) {
+    if (daysAmount >= 7) {
+      return cost - discountLevel2;
+    }
```
michalcywka, line 40: Nesting `if`s like this usually is not a good practice. Moving this condition before `if (daysAmount >= 3)` will make the code cleaner

michalcywka (changes requested):

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2519, opened 2023-03-26
Solution

tuzegor:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 40 * days;
+
+  if (days >= 7) {
+    result = result - 50;
+  } else if (days >= 3 && days < 7) {
+    result = result - 20;
+  }
```
```suggestion
  if (days >= 7) {
    return result - 50;
  } 
  
  if (days >= 3 && days < 7) {
    return result - 20;
  }
```

tetiana-ivanchenko: also move magic numbers in the variables

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const result = 40 * days;
+
+  if (days >= 7) {
+    return result - 50;
+  }
+
+  if (days >= 3 && days < 7) {
+    return result - 20;
```
tetiana-ivanchenko, line 38: move magic numbers in the variables

tetiana-ivanchenko (changes requested):

OlEzhA0 (approved):

PR #2521, opened 2023-03-27
solution

MikiOOOO:


Dka888 (approved):

michalcywka (changes requested):
See checklist, we don't want magic numbers without meaning :) also try to use return in `if`

michalcywka (changes requested):
Please don't commit `package.json`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const regularPrice = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const finalPrice = days * regularPrice;
+
+  if (days >= 7) {
+    return finalPrice - bigDiscount;
+  } else if (days >= 3) {
```
michalcywka, line 38: Is `else` necessary here?

michalcywka (changes requested):

michalcywka (approved):

PR #2522, opened 2023-03-27
solution

ghostargg:


YBM3 (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let result = days * 40;
```
michalcywka, line 32: See checklist, we don't want magic numbers, use variables :) also try to use return in `if`

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const costPerDay = 40;
+  const daysForBigDiscout = 7;
+  const daysForSmallDiscout = 3;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days >= daysForBigDiscout) {
+    return (days * costPerDay) - bigDiscount;
```
michalcywka, line 39: We are using `days * costPerDay` multiple times, maybe we can move it to a const?

michalcywka (changes requested):

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const costPerDay = days * 40;
```
michalcywka, line 32: Ideally we would like to know what 40 is by assigning it to a variable with descriptive name :)

michalcywka (changes requested):

d-wasilewski (approved):

PR #2526, opened 2023-03-27
add task solution

sylcym:


mateusz-michalowski (approved):

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const minThreeDay = 3;
```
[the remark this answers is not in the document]
sylcym: updated

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const daysForSmallDiscount = 3;
+  const daysForBigDiscount = 7;
+  const carRentPricePerDay = 40;
+  const smallDiscountPercentage = 20;
+  const bigDiscountPercentage = 50;
+
+  if (days < daysForSmallDiscount) {
+    return carRentPricePerDay * days;
```
michalcywka, line 38: We use `carRentPricePerDay * days` multiple times, maybe it can be moved to a const?

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const daysForSmallDiscount = 3;
+  const daysForBigDiscount = 7;
+  const carRentPricePerDay = 40;
+  const smallDiscountPercentage = 20;
+  const bigDiscountPercentage = 50;
+  const totalRentCost = carRentPricePerDay * days;
+
+  if (days < daysForSmallDiscount) {
+    return totalRentCost;
+  } else if (days >= daysForSmallDiscount && days < daysForBigDiscount) {
```
michalcywka, line 40: Do we really need `else` here? Also maybe this condition in the brackets can be simplified a bit? Think in which case we will arrive at this point in the code, if we have `return` statement before :)

michalcywka (changes requested):

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2527, opened 2023-03-28
Solution

AnastasiiaVorobets:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  if (days >= 3 && days <= 6) {
+    rent = (days * 40) - 20;
+  }
 
+  if (days >= 7) {
+    rent = (days * 40) - 50;
+  }
+
+  return rent;
+}
```
DarkMistyRoom, line 46: Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  let rent = 0;
+
+  if (days < 3) {
+    rent = days * 40;
+  }
```
DarkMistyRoom, line 35:
```suggestion
  let rent = days * 40;
```

DarkMistyRoom: it gives the same result but looks shorter

DarkMistyRoom (changes requested):
Good job! Let's move all constants to a variables

nazarmatsevych (approved):
👍

PR #2531, opened 2023-03-28
task solution

Amsko6:


natalia-klonowska (approved):

michalcywka (changes requested):
Don't use magic numbers, what is the meaning of 40, 20? :) checklist point 1

d-wasilewski (changes requested):
checklist [5] and [3]

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+  const smallerOff = 20;
+  const biggerOff = 50;
+
+  if (days < 3) {
+    result += basicCost * days;
+
+    return result;
+  }
+
+  if (days >= 3 && days < 7) {
```
d-wasilewski, line 42: You can get rid of using && if you start checking from the highest condition to the lowest.
You should also look at the checkpoint 3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+  const basicCost = 40;
+  const smallerOff = 20;
+  const biggerOff = 50;
```
d-wasilewski, line 34: big, small
and name it discount, it will be cleaner

d-wasilewski (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
d-wasilewski, line 31: you don't need this variable, just return the result immediately instead of this assignment

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+    result += (basicCost * days) - smallDiscount;
+
+    return result;
+  }
+
+  if (days < 3) {
+    result += basicCost * days;
+
+    return result;
+  }
```
d-wasilewski, line 52: checklist 3, you dont need this condition

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2532, opened 2023-03-28
add task solution

marshalek37:


package.json
Ozwiena, line 1: You committed package.json, which I suppose is not necessary

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return (days * 40);
+  } else if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
```
Ozwiena, line 34: Checklist point 5 - the discount amount may change later, it would be better to create a variable here (and for the $50 amount)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return (days * 40);
+  } else if (days >= 3 && days < 7) {
```
Ozwiena, line 33: checklist point 3  :)

Ozwiena (changes requested):

package.json
marshalek37: So did you.

src/calculateRentalCost.js
marshalek37: Fixed :)

package.json
Ozwiena: Great... I did not know :)

Ozwiena (approved):

michalcywka (changes requested):
Please don't commit `package-lock.json` and `package.json`

package.json
```diff
@@ -1,28 +0,0 @@
-{
```
ivadimko, line 1: The package.json file is essential for managing dependencies and scripts in a Node.js project 🚀. It seems that the entire file was removed. Please restore the package.json file to the repository.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discount = 20;
+  const bigDiscount = 50;
```
ivadimko, line 32: The variable name 'bigDiscount' does not clearly describe the purpose or the amount of the discount. Consider renaming it to 'largeDiscount' or 'discountFor7days'. 🤔

PR #2533, opened 2023-03-28
solution

Sasha-Bachynskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
+  const lowDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
```
jv-aquino, line 35: It's a better practice to store these values as constant, improving the readability and making it easier to change them

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayCost = 40;
+  const lowDiscount = 20;
+  const bigDiscount = 50;
```
jv-aquino, line 33: The consts should be uppercase (good practice)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    return days * dayCost;
+  } else if (days < 7 && days >= 3) {
+    return (days * dayCost) - lowDiscount;
+  } else {
```
jv-aquino, line 39: Remember that the last else isn't needed at all, there's a cleaner way to do it

jv-aquino (changes requested):
Reasonable solution but you can create a great one!

PR #2534, opened 2023-03-28
solution

natalia-klonowska:


marshalek37 (approved):
Według mnie wszystko jest poprawnie :)

michalcywka (changes requested):
Please don't commit `package-lock.json` and `package.json`

michalcywka (changes requested):
We don't want to completely delete package-lock.json and package.json, we just don't want them to be changed :)
Please copy the content of these files from master branch and commit them.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const rentalCost = days * 40;
```
michalcywka, line 32: Ideally we would like 40 to also be a variable :)

michalcywka (changes requested):

d-wasilewski (approved):

PR #2535, opened 2023-03-28
solution

Ozwiena:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
-  // write code here
+  const dailyCost = 40;
+  const basicCost = dailyCost * days;
+  const basicDiscount = 20;
+  const higherDiscount = 50;
+
+  let totalCost = basicCost;
+
+  if (days >= 3 && days < 7) {
+    totalCost = basicCost - basicDiscount;
```
mwgreybeard72, line 39: I don't think this is necessary. You could instead use return and place these variables in the return command instead. However, this works just as well.

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
```
mwgreybeard72, line 19: I believe You should only have changed the ./src file. I might be wrong tho. Would be safe to remove changes made in the package.json and package-lock.json files before the mentor review. :D

mwgreybeard72 (approved):
Other than the two comments, which I do not think are that relevant I see no issues with this project.

src/calculateRentalCost.js
michalcywka: This is actually in the checklist point 3 :) try to use one line, then we don't have to create variable with `let`

michalcywka (changes requested):
Please don't commit `package.json` and `package-lock.json`

michalcywka (changes requested):
We don't want to completely delete `package-lock.json` and `package.json`, we just don't want them to be changed :)
Please copy the content of these files from `master` branch and commit them.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dailyCost = 40;
+  const basicCost = dailyCost * days;
+  const basicDiscount = 20;
+  const higherDiscount = 50;
+
+  if (days >= 3 && days < 7) {
```
michalcywka, line 36: Maybe we can reorganize checking the conditions to avoid using `&&`? :)
What if we first checked for large discount and then for smaller one?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return basicCost - basicDiscount;
+  }
+
+  if (days >= 7) {
+    return basicCost - higherDiscount;
+  }
+
+  if (days < 3) {
```
michalcywka, line 44: Do we really need this if here? :)

michalcywka (changes requested):

d-wasilewski (approved):

PR #2536, opened 2023-03-28
solution

kq333:


SuaBy514 (approved):

michalcywka (changes requested):
We don't want magic numbers here :) Checklist point 1 and 4

michalcywka (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const mdDiscount = 20;
+  const lgDiscount = 50;
+  const price = 40;
+
+  if (days <= 2) {
+    return days * price;
```
michalcywka, line 36: `days * price` is repeated, can me move it to a variable? :)

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days <= 2) {
+    return priceBeforDiscount;
+  }
+
+  if (days >= 3 && days <= 6) {
+    return priceBeforDiscount - mdDiscount;
+  }
+
+  if (days >= 7) {
```
michalcywka, line 44: Do we really need this if here?

michalcywka (changes requested):

src/calculateRentalCost.js
d-wasilewski: we don't. It's checklist point 3

d-wasilewski (changes requested):

PR #2537, opened 2023-03-28
Solution first appraoch 28/03/2023 23:01

mwgreybeard72:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const cost = 40;
+  const total = (cost * days);
+  const largeDiscount = 50;
```
ivadimko, line 34: The numbers used should have a name for better readability. Consider adding constants like DAILY_COST = 40; for better code readability. 📚

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const cost = 40;
+  const total = (cost * days);
```
michalcywka, line 33: `()` not necessary here

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
...
+  const largeDiscount = 50;
+  const smallDiscount = 20;
+  const largeDiscountReq = 7;
+  const smallDiscountReq = 3;
+
+  if (days < largeDiscountReq && days >= smallDiscountReq) {
```
michalcywka, line 39: Maybe we can avoid using `&&` in if, after reorganising them. What if we checked first for condition when there is no discount, then small discount and then the large one?

michalcywka (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,27 @@
...
+
+  if (days < largeDiscountReq) {
+    return total - smallDiscount;
+  }
+
+  if (days >= smallDiscountReq) {
+    return total - largeDiscount;
+  }
+
+  return total;
```
d-wasilewski, line 52: What if you checked from the highest condition to the lowest? You would avoid 1 if statement, and you wouldn't have to write `return total` twice

src/calculateRentalCost.js
```diff
@@ -29,6 +29,27 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const cost = 40;
+  const total = cost * days;
+  const largeDiscount = 50;
+  const smallDiscount = 20;
+  const noDiscount = 2;
```
d-wasilewski, line 36: and you can then remove this variable

src/calculateRentalCost.js
```diff
@@ -29,6 +29,27 @@
...
+  const largeDiscountReq = 7;
```
d-wasilewski, line 37: what is Req? It's not necessary to add this I guess

d-wasilewski (changes requested):

src/calculateRentalCost.js
mwgreybeard72: It's a requirement.

src/calculateRentalCost.js
mwgreybeard72: Alright

src/calculateRentalCost.js
mwgreybeard72: I was thinking the very same when I wrote it but wasn't sure.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
...
+  }
+
+  if (days < largeDiscountRequirement) {
+    return total - smallDiscount;
+  }
+
+  if (days >= smallDiscountRequirement) {
+    return total - largeDiscount;
+  }
 }
```
d-wasilewski, line 50: Almost! But try to turn them around. 
Start with this one 
```
if (days >= smallDiscountRequirement) {
    return total - largeDiscount;
  }
  ```
  and use max 2 if statements

**Review by d-wasilewski — CHANGES_REQUESTED** [review]

File: src/calculateRentalCost.js [hunk 8]
```diff
@@ -29,6 +29,24 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const cost = 40;
+  const total = cost * days;
+  const largeDiscount = 50;
+  const smallDiscount = 20;
+  const largeDiscountRequirement = 7;
+  const smallDiscountRequirement = 3;
```
d-wasilewski, line 37:
```suggestion
  const largeDiscount = 7;
  const smallDiscount = 3;
```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2538, opened 2023-03-28
Solution

andrijdudar:


src/calculateRentalCost.js
```diff
@@ -28,7 +29,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayRenta = 40;
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
+  const allDaysNoDiscount = oneDayRenta * days;
```
DarkMistyRoom, line 35:
```suggestion
  const priceNoDiscount = oneDayRenta * days;
```

DarkMistyRoom: allDays make me think about amount of days more then about price

src/calculateRentalCost.js
```diff
@@ -28,7 +29,20 @@
...
+
+  if (days >= 3 && days < 7) {
+    return allDaysNoDiscount - discountThreeDays;
+  }
+
+  if (days >= 7) {
+    return allDaysNoDiscount - discountSevenDays;
+  }
```
DarkMistyRoom, line 43:
```suggestion
  if (days >= 7) {
    return allDaysNoDiscount - discountSevenDays;
  }

  if (days >= 3) {
    return allDaysNoDiscount - discountThreeDays;
  }
```

DarkMistyRoom: we can simplify the query by changing the order

DarkMistyRoom (approved):
Good job! Pay attention to comments

PR #2540, opened 2023-03-29
Solution

Ir-ra:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (40 * days) - 50;
+  } else if (days >= 3) {
+    return (40 * days) - 20;
+  } else {
+    return 40 * days;
+  }
```
qoa11a, line 37: - Move numbers to variables so it's understandable what they represent.
- When you use `return` inside a function, it immediately exits the function. So there's no need to use `else`.

qoa11a (changes requested):

witflash (approved):
Good job 👍

PR #2542, opened 2023-03-29
solution

mariuszgit:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
DorotaLeniecDev, line 32: Good solution. To make it more adjastable you could put 40 into variable. So when the car cost for a day changes you can modify it in just one place

DorotaLeniecDev (approved):
Very good! One comment

PR #2544, opened 2023-03-29
solution

DanyloPasko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 3 && days < 7) {
+    return 40 * days - 20;
+  } else if (days >= 7) {
+    return 40 * days - 50;
+  }
+
+  return 40 * days;
 }
```
Viktor-Kost, line 38: 1) try to avoid using `magic` number. Create variable for your numeric values
2) `else if` is redundant since you use `early` return. It's enough to do like this
```
if (condition1) {
 return value1
}

if (condition2) {
 return value2
}

return value3
```

Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    return dayprice * days - 20;
+  }
+
+  if (days >= 7) {
+    return dayprice * days - 50;
+  }
+
+  return dayprice * days;
```
Viktor-Kost, line 41: nice) 
 Yet it would be even better if you create a variable for repeating part of code `dayprice * days` and for `50` `20`
 ```
 const initialPrice = dayprice * days
 const WEEK_RENTAL = 50
 const DAY_ RENTAL = 20
 ```
 I wrote in uppercase since they are constants (non-changed values)

Viktor-Kost (approved):
Good job!
It's approved, yet I've left one comment for you to know how to improve more

PR #2545, opened 2023-03-29
completed the task

RomanOstrous:


src/calculateRentalCost.js
```diff
@@ -1,34 +1,42 @@
...
+function calculateRentalCost(days) {
+  let sum = days * 40;
+
+  if (days >= 7) {
+    sum -= 50;
+  } else if (days >= 3) {
+    sum -= 20;
+  }
+
+  return sum;
```
```suggestion
  if (days >= 7) {
    return sum - 50;
  }
  
  if (days >= 3) {
    return sum - 20;
  }

  return sum;
```
You can just immediately return the full price

polosanya (changes requested):
Good job! Please create variables for all main values (40, 20, 50 etc.)

sTorba24 (approved):
It's cool!

PR #2546, opened 2023-03-29
Solution

HannieA:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let totalAmount = 40;
+
+  if (days >= 7) {
+    totalAmount = totalAmount * days - 50;
+  } else if (days >= 3) {
+    totalAmount = totalAmount * days - 20;
+  } else {
+    totalAmount = totalAmount * days;
+  }
```
OlEzhA0, line 39: what is `50` and `20`? add some meaningful constants
```suggestion
  if (days >= 7) {
    return totalAmount * days - 50;
  } 
  
  if (days >= 3) {
    return totalAmount * days - 20;
  }
  
  return totalAmount * days;
```

OlEzhA0 (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const discount3Days = 20;
+  const discount7Days = 50;
+  const totalAmount = 40;
+
+  if (days >= 7) {
+    return totalAmount * days - discount7Days;
+  } else if (days >= 3) {
+    return totalAmount * days - discount3Days;
+  }
```
OlEzhA0, line 39: we don't need to do `else if` when return some value in `if`

```
if (days >= 7) {
  return ...
}

if (days >= 3) {
  return ....
}

return ...
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  return totalAmount * days; ;
```
OlEzhA0, line 41:
```suggestion
  return totalAmount * days;
```

OlEzhA0 (approved):

PR #2548, opened 2023-03-30
add solution

PufAnimalsDev:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+  const smallDiscountDays = 3;
+  const basicDiscountDays = 7;
+
+  if (days >= basicDiscountDays) {
+    return days * pricePerDay - basicDiscount;
+  }
+
+  if (days >= smallDiscountDays && days < basicDiscountDays) {
+    return days * pricePerDay - smallDiscount;
+  }
```
d-wasilewski, line 44: Since the days >= basicDiscountDays and days >= smallDiscountDays && days < basicDiscountDays conditions are mutually exclusive, you can replace the second if statement with an else if statement. This makes it clear that only one of these conditions will be met.

But it's still good this way :)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const pricePerDay = 40;
+  const smallDiscount = 20;
+  const basicDiscount = 50;
```
d-wasilewski, line 34: I would change the naming here, it's not clear that the basic discount is bigger than small

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2549, opened 2023-03-30
calculateRentalCost

Donovanoff:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  totalAmount = price * days;
+
+  if (days >= 7) {
+    totalAmount = totalAmount - discount50;
+  } else if (days >= 3) {
+    totalAmount = totalAmount - discount20;
+  };
+
+  return totalAmount;
```
IvanFesenko, line 44: No need to reasign value - you can return if some condition is true. Also do not use else statement.

```suggestion
  const basicCost = price * days;

  if (days >= 7) {
    return basicCost - discount50;
  } 
  if (days >= 3) {
    return basicCost - discount20;
  };

  return basicCost;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const discount20 = 20;
+  const discount50 = 50;
```
IvanFesenko, line 33: Fix variable naming
2023-03-30_15h48_01

IvanFesenko (changes requested):

sTorba24 (approved):
Well done!

PR #2551, opened 2023-03-30
Solution

VladyslavKucher:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = '';
```
polosanya, line 31: image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    result = days * 40 - 50;
+  } else if (days >= 3) {
```
polosanya, line 35: No need to store it in result. Just immediately return the full price

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    result = days * 40 - 20;
+  } else {
+    result = days * 40;
```
polosanya, line 38: Don't duplicate yourself. Create a variable for `days * 40` to make your code cleaner.
Same with all main values (40, 20, 50 etc.)

polosanya (changes requested):
Good job! Let's use variables to make your code descriptive for other developers. 
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basePrice = days * 40;
+  const discountThree = -20;
```
OleziO, line 32: Make the discount a positive number and subtract it, so the code will be more readable.

OleziO, line 32: Make the discount a positive number and subtract it, so the code will be more readable.

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basePrice = days * 40;
+  const discountThree = 20;
+  const discountSeven = 50;
```
polosanya, line 33: image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return basePrice - discountSeven;
+  } else if (days >= 3) {
+    return basePrice - discountThree;
+  } else {
+    return basePrice;
+  }
```
polosanya, line 41: image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basePrice = days * 40;
```
polosanya, line 31: It would be great to create a variable for `40` too

polosanya (changes requested):
Almost perfect!

Anastasiia-Svintsova (approved):

PR #2552, opened 2023-03-30
RentalCost

Tetiana-Hishchak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    sum = days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    sum = days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    sum = days * 40 - 50;
+  }
```
OlEzhA0, line 43: what is `40`, `20` and `50`? add some meaningful constants; 

and why do we need to assign all values to `sum`? we can just return it at once

```suggestion
  if (days < 3) {
    return days * 40;
  }

  if (days >= 3 && days < 7) {
    return days * 40 - 20;
  }

  if (days >= 7) {
    return days * 40 - 50;
  }
```

OlEzhA0 (changes requested):

anstsot (approved):
Great job, approved!

PR #2553, opened 2023-03-31
Solution

Viktoriia6666:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let RentalCost = 0;
+  const RentCostPerDay = 40;
+  const BigRentalCost = 50;
+  const LittleRentalCost = 20;
```
SanyaBratashchuk, line 35: camelCase 
```suggestion
  let rentalCost = 0;
  const rentCostPerDay = 40;
  const bigRentalCost = 50;
  const littleRentalCost = 20;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
...
+  if (days <= 2) {
+    RentalCost = days * RentCostPerDay;
+  } else {
+    if (days >= 3 && days < 7) {
+      RentalCost = days * RentCostPerDay - LittleRentalCost;
+    } else {
+      if (days >= 7) {
+        RentalCost = days * RentCostPerDay - BigRentalCost;
+      }
+    }
```
SanyaBratashchuk, line 46: move `rentConstPerDay * days` to the variable

SanyaBratashchuk: You can get rid of nested ifs. By using return statement in your ifs

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+  const rentCostPerDay = 40;
+  const bigRentalCost = 50;
+  const littleRentalCost = 20;
+  let everyDay = days * rentCostPerDay;
+
+  if (days >= 3 && days < 7) {
+    everyDay = everyDay - littleRentalCost;
+  } else if (days >= 7) {
+    everyDay = everyDay - bigRentalCost;
+  }
```
Volodymyr-Melnyk-Fron10der, line 41: previous comment wasn't fixed
```suggestion
  const everyDay = days * rentCostPerDay;

  if (days >= 3 && days < 7) {
    return everyDay - littleRentalCost;
  } else if (days >= 7) {
    return everyDay = everyDay - bigRentalCost;
  }
```

Volodymyr-Melnyk-Fron10der (changes requested):

Volodymyr-Melnyk-Fron10der (approved):

PR #2558, opened 2023-04-03
Solution

IvanOrinchak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  for (let i = 1; i <= days; i++) {
+    result += 40;
+  }
```
Volodymyr-Melnyk-Fron10der, line 35:
```suggestion
  let result = 40 * days;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days < 3) {
+    return result;
+  }
+
+  if (days < 7) {
+    return result - 20;
+  }
+
+  return result - 50;
 }
```
Volodymyr-Melnyk-Fron10der, line 46: move all magic numbers to constants

Volodymyr-Melnyk-Fron10der (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceOneTrip = 40;
+  const totalCost = priceOneTrip * days;
+  const discountTwenty = 20;
+  const discountFifty = 50;
```
sTorba24, line 34: what if we change discount amount ?)
you can name it using small-big etc.

sTorba24 (approved):
Well done! I left you one little comment :)

PR #2561, opened 2023-04-04
Solution

EvgenBilenkiy:


src/calculateRentalCost.js
```diff
@@ -25,10 +25,21 @@
  * @param {number} days
  *
  * @return {number}
+ *
  */
 
 function calculateRentalCost(days) {
   // write code here
+  const costPerDay = 40;
+  let totalCost = costPerDay * days;
```
```suggestion
  const totalCost = costPerDay * days;
```
image

src/calculateRentalCost.js
```diff
@@ -25,10 +25,21 @@
...
+
+  if (days >= 7) {
+    totalCost -= 50;
+  } else if (days >= 3) {
+    totalCost -= 20;
+  }
```
```suggestion
  if (days >= 7) {
    return totalCost - 50;
  }
  
  if (days >= 3) {
    return totalCost - 20;
  }
```
Better immediately return the full price to avoid using else...if

polosanya (changes requested):
Almost done! Please create variables for 20 and 50 too

agniya-i (approved):
👍

PR #2565, opened 2023-04-05
solution

Absolution26:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return rent * days;
+  }
+
+  if (days >= 3 && days < 7) {
+    return rent * days - 20;
+  }
+
+  if (days >= 7) {
+    return rent * days - 50;
+  }
```
Volodymyr-Melnyk-Fron10der, line 43: 1. move all magic numbers into constants
2. rent * days is duplicated 3 times, move it into variable` const price = days * 40;`

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there_)

polosanya (approved):
Good job!

PR #2566, opened 2023-04-05
Solution

natalia-matash:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  if (days >= 3 && days < 7) {
```
DarkMistyRoom, line 33: Consider using constants for magic numbers like 3, 7, 40, 20, and 50. This will make the code more readable and maintainable. 🤓

src/calculateRentalCost.js
```diff
@@ -29,6 +29,16 @@
...
+    return (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    return (days * 40) - 50;
+  }
```
DarkMistyRoom, line 39: you can place this query first, then won't need to check if days<7 in the next query

DarkMistyRoom (changes requested):
Good job! Let's improve

loralevitska (approved):
🔥

PR #2567, opened 2023-04-06
Solution

RatiBorRozGa:
All tests passed

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
     "eslint": "^5.16.0",
     "eslint-plugin-jest": "^22.4.1",
     "eslint-plugin-node": "^8.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 19: In the future - Do not add extra files to PR
To add the file to a commit use git add src/.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let totalAmount = 0;
+
+  if (days > 1 && days < 3) {
+    totalAmount = days * 40;
```
SerhiiKirik, line 35: what is days * 40?
You repeat it 3 times
Move this to the variables
Also, try to avoid else if

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

SerhiiKirik (approved):
Well done 🚀

PR #2571, opened 2023-04-07
Solution

AnastassiaH:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
   // write code here
+  let total;
+
+  if (days >= 7) {
+    total = days * 40 - 50;
+  } else if (days >= 3) {
+    total = days * 40 - 20;
+  } else {
+    total = days * 40;
+  }
```
slavanosatskyi, line 40: move all magic numbers (20, 40, 50) to constants

slavanosatskyi (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const price = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  let total;
```
Viktor-Kost, line 35: it's a redundant variable
```suggestion
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+
+  if (days >= 7) {
+    total = days * price - bigDiscount;
+  } else if (days >= 3) {
+    total = days * price - smallDiscount;
+  } else {
+    total = days * price;
+  }
+
+  return total;
```
Viktor-Kost, line 45: 1) try to avoid using `else if` to make your code clearer. 
You can just use `early return`
2) also to avoid duplicate you can move `days * price` to a variable like `rentalCost = days * price`
```suggestion
  if (days >= 7) {
    return rentalCost - bigDiscount;
  } 
  
  if (days >= 3) {
    return rentalCost - smallDiscount;
  } 
  
  return rentalCost;
```

Viktor-Kost (changes requested):

YegorKorenkov (approved):
Looks good

PR #2574, opened 2023-04-08
Solution

VKonoplianyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,9 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40 * days;
+
+  return rentalCost - (days >= 7 ? 50 : days >= 3 ? 20 : 0);
```
Volodymyr-Melnyk-Fron10der, line 33: 1. short - it's cool, but nested ternary conditions is a antipattern, consider using if statements
2. move all magic numbers into constants

Volodymyr-Melnyk-Fron10der (changes requested):

SanyaBratashchuk (approved):

PR #2576, opened 2023-04-09

YelyzavetaMolochynska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = 0;
```
VitaliyBondarenko1982, line 31: this variable redundant. retutn immediately what you need

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 3) {
+    cost = days * 40;
+  } else if (days >= 3 && days < 7) {
+    cost = (days * 40) - 20;
+  } else if (days >= 7) {
+    cost = (days * 40) - 50;
+  }
+
+  return cost;
```
VitaliyBondarenko1982, line 41:
```suggestion
  if (days < 3) {
   return days * 40;
  } 
  
  if (days >= 3 && days < 7) {
    return  (days * 40) - 20;
  } 
  
  return cost = (days * 40) - 50;
```
try to avoid else if where it is possible

VitaliyBondarenko1982 (changes requested):
make variables for all `magic` numbers

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    return (days * 40) - 50;
+  }
```
VitaliyBondarenko1982, line 41:
```suggestion
  return (days * 40) - 50;
```
no need last if

VitaliyBondarenko1982: comment about constants for magic numbers not fixed
for example
 ```
  const dayCost = 40;
  const threeDaysDiscount = 20;
 ```
also make for 50, 3, 7. 
It make code more readable

VitaliyBondarenko1982 (changes requested):

etojeDenys (approved):
lgtm

PR #2579, opened 2023-04-11
Solution

illiavinkov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
+
+  if (days >= 7) {
+    totalCost -= 50;
+  } else if (days >= 3) {
+    totalCost -= 20;
+  }
```
IvanFesenko, line 37: What is the magic numbers 20, 40, 50?
Create separate const for them
2023-04-11_13h30_49

Do not use else statement
Return value if condition is true

IvanFesenko (changes requested):

VitaliyBondarenko1982 (approved):

PR #2580, opened 2023-04-11
FirstTaskJSGit

Ryshon:


witflash (changes requested):
Well done 👍🏻 
Just one thing. Could you move discount values to the constant and use them in your checking.

Ryshon:
> Well done 👍🏻
> Just one thing. Could you move discount values to the constant and use them in your checking.

Do you mean - const price = 40 * days;?

witflash (changes requested):
I mean values "20", "40" and "50"

Ryshon:
> I mean values "20", "40" and "50"

I got it)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let price = 40 * days;
+  const disc1 = 50;
+  const disc2 = 20;
```
agniya-i, line 33:
```suggestion
  const basicDiscount = 20;
```

give more details in variable name)

agniya-i (approved):

PR #2581, opened 2023-04-11
Solution

Nikita211112:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
```
polosanya, line 31: Create a const variable for `days * 40`. You duplicate this value a lot
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    sum = (days * 40) - 50;
+  } else if (days >= 3 && days < 7) {
+    sum = (days * 40) - 20;
+  } else {
+    sum = days * 40;
```
polosanya, line 38: No need to use if...else. It makes your code a bit complicated. 
Just return the full price immediately if the condition is true
image

polosanya (changes requested):
Almost done! Let's improve your code a bit)
Create variables for all main values (40, 50, 20 etc.)

loralevitska (approved):
🔥

PR #2583, opened 2023-04-11
done

p0mk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = days < 3
+    ? days * 40
+    : days < 7
+      ? (days * 40) - 20
+      : days >= 7
+        ? (days * 40) - 50
+        : 0;
```
DarkMistyRoom, line 37: Nested ternary operators are EVIL. Avoid them.
And try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

DarkMistyRoom (changes requested):

src/calculateRentalCost.js
p0mk: why ternary op. is Evil ?

anstsot: one ternary operator is good but when there are a lot it's hard to read and understand what's going on

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  let discount = 0;
+
+  if (days < 3) {
+    discount = 0;
+  }
```
anstsot, line 36: you already have a default value so there is no need to set it
```suggestion
  let discount = 0;
```

anstsot (approved):
Good job, approved

PR #2587, opened 2023-04-12
TaskSolution

ZupaFly:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let price = 0;
+
+  if (days < 3) {
+    price = days * 40;
```
Anastasiia-Svintsova, line 34: What do 40, 20 and 50 mean? Don't use mystery numbers. Move them to variables. BTW you use this value `days * 40` not once. Create a variable for it as well

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    price = (days * 40) - 20;
+
+    return price;
+  } else if (days >= 7) {
+    price = (days * 40) - 50;
+
+    return price;
+  } else {
+    return 0;
+  }
```
Anastasiia-Svintsova, line 47: Redundant

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let price = 0;
+
+  if (days < 3) {
+    price = days * 40;
+
+    return price;
```
Anastasiia-Svintsova, line 36: You return the price in each case. Just move it to the end of the function

Anastasiia-Svintsova (changes requested):
Let's add some improvements

anstsot (approved):
Great job, approved!

PR #2590, opened 2023-04-13
My commit

Khalniepiesova:


Volodymyr-Melnyk-Fron10der (changes requested):
1. get rid of count variable, just return the result from each if block right away (the last if statement will be redundant, you should remove it) 

2. move all magic numbers into constants
3. days * 40 is repeating 3 times, move it into variable as well

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const pricePerDays = 40 * days;
+
+  if (days >= 3
+    && days < 7) {
+    return pricePerDays - 20;
+  }
+
+  if (days >= 7) {
+    return pricePerDays - 50;
+  }
```
Volodymyr-Melnyk-Fron10der, line 40:
```suggestion
  if (days >= 7) {
    return pricePerDays - 50;
  }
  
    if (days >= 3) {
    return pricePerDays - 20;
  }
```

Volodymyr-Melnyk-Fron10der (changes requested):
Agai, move all magic numbers into a constants (20, 50..)

SerhiiKirik (approved):
Great 🔥

PR #2594, opened 2023-04-14
Solution

IrynaBeznoshchenko:


AntonKalnohuz (approved):
good job

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  let price = 0;
```
murat-yesimbietov, line 31: It's better to use a `const` for the base rental cost (40) and discounts (20, 50) to improve code readability and maintainability. 😊

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+
+  if (days < 3) {
```
murat-yesimbietov, line 33: Since you are using `if` and `else if` statements, there is no need to check for `days >= 3`, as it's already implied by the previous condition. 😉

murat-yesimbietov (changes requested):
Changes are required. Also, please, don't include package-lock.json file (don't add all files to the index when making a commit)

src/calculateRentalCost.js
IrynaBeznoshchenko: Code is rewrited

src/calculateRentalCost.js
IrynaBeznoshchenko: Code is rewrited

OlgaAkimenko:
good

PR #2595, opened 2023-04-15
cost

NazarVasylyshyn:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40 * days;
+
+  if (days >= 7) {
+    return cost - 50;
+  } else if (days >= 3) {
```
YegorKorenkov, line 35: Please try to avoide else / if construction(use only if)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40 * days;
```
YegorKorenkov, line 31: Please move all numbers to variables

YegorKorenkov (changes requested):

Volodymyr-Melnyk-Fron10der (approved):
Amazing job!

PR #2600, opened 2023-04-16
Solution

fibrr:


stanyslav-petryk (approved):
Good Job

src/calculateRentalCost.js
```diff
@@ -21,14 +21,20 @@
...
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3 && days <= 6) {
+    return days * 40 - 20;
+  }
+
+  if (days < 3) {
```
murat-yesimbietov, line 35: You can remove this condition `days < 3` and use just `return days * 40;` as the default case. 🙌

src/calculateRentalCost.js
```diff
@@ -21,14 +21,20 @@
...
- * @return {number}
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3 && days <= 6) {
```
murat-yesimbietov, line 31: You can replace the condition `days <= 6` with `days < 7` to avoid overlapping with the first condition. 😊

src/calculateRentalCost.js
```diff
@@ -21,14 +21,20 @@
...
- *
- * @param {number} days
- *
- * @return {number}
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
```
murat-yesimbietov, line 28: Create constant variables for such magic numbers as 40 and 50 with a meaningful name. Such approach improves the readability and maintenance of your code

murat-yesimbietov (changes requested):
Changes are required. Also, remove the package-lock.json and package.json files from the PR  (don't add all files to the index when making a commit)

src/calculateRentalCost.js
fibrr: Done

src/calculateRentalCost.js
```diff
@@ -21,14 +21,22 @@
...
-  // write code here
+  const payment = 40;
+  const lowDiscount = 20;
+  const highDiscount = 50;
+
+  if (days >= 7) {
+    return days * payment - highDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
```
dashaprotsko, line 35: You can remove the condition `days < 7` since it is redundant. If the code reaches this point, it implies that `days < 7` is already true. 😊

dashaprotsko (approved):
Well done!

PR #2601, opened 2023-04-16
Solution

Anastasiia-Shahoval:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = days * 40;
```
VitaliyBondarenko1982, line 31: create variables for `magic` numbers
40, 20, 50, 3, 7

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+
+  if (days >= 3 && days < 7) {
+    sum -= 20;
+  } else
+  if (days >= 7) {
+    sum -= 50;
+  }
```
VitaliyBondarenko1982, line 38: avoid else if where it is possible
```suggestion
  }
  
  if (days >= 7) {
    sum -= 50;
  }
```

VitaliyBondarenko1982 (changes requested):
please make some refactoring.
pay attention on [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) about naming variables.
Screenshot from 2023-04-16 18-19-26

src/calculateRentalCost.js
Anastasiia-Shahoval: Thank you! corrected

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const rentPrice = 40;
+  let carRentalAmount = days * rentPrice;
+
+  if (days >= basicDays && days < subsequentDays) {
+    carRentalAmount -= basicDiscount;
+  }
+
+  if (days >= subsequentDays) {
+    carRentalAmount -= subsequentDiscount;
+  }
```
Volodymyr-Melnyk-Fron10der, line 44:
```suggestion
  if (days >= subsequentDays) {
    carRentalAmount -= subsequentDiscount;
  }
  
  if (days >= basicDays) {
    carRentalAmount -= basicDiscount;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
-  // write code here
+  const basicDiscount = 20;
+  const subsequentDiscount = 50;
+  const basicDays = 3;
+  const subsequentDays = 7;
+  const rentPrice = 40;
+  let carRentalAmount = days * rentPrice;
+
+  if (days >= basicDays && days < subsequentDays) {
+    carRentalAmount -= basicDiscount;
```
Volodymyr-Melnyk-Fron10der, line 39: you actually can just return it right away and make carRentalAmount as a const
```suggestion
    return carRentalAmount - basicDiscount;
```

Volodymyr-Melnyk-Fron10der: using const instead of let is better for performance

Volodymyr-Melnyk-Fron10der (changes requested):

src/calculateRentalCost.js
Volodymyr-Melnyk-Fron10der: get rid of extra condition

src/calculateRentalCost.js
Anastasiia-Shahoval: Thanks! Done :)

etojeDenys (approved):
lgtm

PR #2602, opened 2023-04-16
Solution

Nikita211112:


Nikita211112:
This is the second time I've completed the task. The first time you confirmed to me, nothing has changed on the site. What could be the reason?

PR #2604, opened 2023-04-17
 Transportation on vacation

IzzetYusufov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
-  // write code here
+  switch (true) {
+    case days < 3:
+      return days * 40;
+
+    case days >= 7:
+      return (days * 40) - 50;
+
+    case days >= 3:
+      return (days * 40) - 20;
```
vlad-demchuk, line 39: move 20, 40, 50  to variables

vlad-demchuk (changes requested):

VitaliyBondarenko1982 (approved):

PR #2607, opened 2023-04-17
Solution

VladVeklenko:


Radochin-Alexander (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const carRent = 40;
```
murat-yesimbietov, line 31:
```suggestion
  const dailyRent = 40;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days < 3) {
+    return carRent * days;
+  } else if (days >= 3 && days < 7) {
```
murat-yesimbietov, line 35: It's better to remove `else` because you use the `return` operator. Also remove `days >= 3`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+    return (carRent * days) - 20;
```
murat-yesimbietov, line 36: It's better to create variables for every value (such as 20) to improve readability and maintenance. 
E.g `const threeDayDiscount = 20;`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days >= 7) {
```
murat-yesimbietov, line 37: I would remove the last `else if` and put `return (carRent * days) - 50;` as default operation

murat-yesimbietov (changes requested):
Little changes are required

murat-yesimbietov (approved):

PR #2610, opened 2023-04-17
Solution

Radochin-Alexander:


yuliiapyrih (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const discountSevenDays = 50;
+  const discountThreeDays = 20;
+  let total = oneDay * days;
+
+  if (days >= 7) {
+    total -= discountSevenDays;
+  } else if (days >= 3) {
+    total -= discountThreeDays;
+  } else {
+    total = oneDay * days;
```
dashaprotsko, line 41: This line is redundant, as the value of `total` is already calculated in line 9. You can remove line 16. 👍

dashaprotsko, line 41: This line is unnecessary. You have already calculated `total` at line 9, so there's no need to do it again. You can remove line 16. 😊

PR #2611, opened 2023-04-17
Solution

s3rg4sh:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let cost = 0;
+
+  for (let i = 0; i < days; i++) {
+    cost += 40;
+  }
```
etojeDenys, line 35: we can just write it as `days * 40`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 3 && days < 7) {
+    return cost - 20;
+  } else if (days >= 7) {
+    return cost - 50;
+  } else {
```
etojeDenys, line 41: move all these magic numbers to some vars

etojeDenys (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const costSmallOff = cost - 20;
+  const costBigOff = cost - 50;
+
+  if (days >= 3 && days < 7) {
+    return costSmallOff;
+  } else if (days >= 7) {
+    return costBigOff;
+  } else {
+    return cost;
+  }
```
VitaliyBondarenko1982, line 41:
```suggestion
  } 
  
  if (days >= 7) {
    return costBigOff;
  } 
  
  return cost;
```
try to avoid else if where it possible. it make code more readable

VitaliyBondarenko1982 (approved):

PR #2613, opened 2023-04-17
Solution

yuliiapyrih:


angeliinasen (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentForDays = 40 * days;
```
murat-yesimbietov, line 31: Consider using more descriptive variable names to improve readability. For example, change `rentForDays` to `totalRent`.

murat-yesimbietov (changes requested):
Little changes are required. Also, remove the package-lock.json and package.json files from the PR  (don't add all files to the index when making a commit)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalRent = 40 * days;
+  const smallDiscount = 20;
+  const largeDiscount = 50;
+
+  if (days >= 1 && days < 3) {
```
dashaprotsko, line 35: Since the `days` parameter should always be greater than or equal to 1, there's no need to check for `days >= 1`. You can simplify this condition to `days < 3`. 🧹

dashaprotsko (approved):
Well done :)
...but take a look at comment

PR #2615, opened 2023-04-17
Solution

kosherenkov:


yaroslav-khoma (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum;
+
+  if (days >= 1 && days < 3) {
+    sum = 40 * days;
```
murat-yesimbietov, line 34: It's better to create variables for every value (such as 40) to improve readability and maintenance. 
For example `const dailyCost = 40;`

murat-yesimbietov (changes requested):
Little changes are required. Also, remove the package-lock.json and package.json files from the PR  (don't add all files to the index when making a commit)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum;
```
AnnaPalamarchukQA, line 31: It is recommended to initialize the variable `sum` with a value, for example `0`. This ensures that the variable has a default value in case none of the conditions are met.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 1 && days < 3) {
+    sum = 40 * days;
+  }
+
+  if (days >= 3 && days < 7) {
```
AnnaPalamarchukQA, line 37: You can use `else if` instead of separate `if` statements to make it clear that these conditions are mutually exclusive. This also avoids unnecessary condition checks if a previous condition has already been met.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    sum = 40 * days - 20;
+  }
+
+  if (days >= 7) {
```
AnnaPalamarchukQA, line 41: You can use `else` instead of `if` for the last condition, as it covers all the remaining cases.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum;
+
+  if (days >= 1 && days < 3) {
```
AnnaPalamarchukQA, line 33: Since `days` is a parameter, it's guaranteed to be at least `1`. So, the condition `days >= 1` is redundant and can be removed.

AnnaPalamarchukQA (changes requested):
Please pay attention to the comments from mentors

murat-yesimbietov (approved):

PR #2620, opened 2023-04-18
Solution

Dolnys:


anpiet (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  let result = days * 40;
+
+  if (days >= 7) {
+    result = result - 50;
+  } else if (days >= 3) {
+    result = result - 20;
+  }
+
+  return result;
```
JoyCoffeeAddict, line 39: Checklist point 2, use `const`s for constant values. Also checklist point 5. Try to come up with idea how to avoid writing `else if`

JoyCoffeeAddict (changes requested):

src/calculateRentalCost.js
Dolnys: Done! ☺️

JoyCoffeeAddict: You should still capitalize whose constants, because they should never change in your code, they are like math consts e.g. `Math.PI` ,  `Math.SQRT2`

Dolnys: Ok, the variables have been corrected ☺️

choeqq (approved):
Nicely done ✅

PR #2621, opened 2023-04-18
Solution

edgarlokys:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  let totalCost = days * costPerDay;
+
+  if (days >= 7) {
+    totalCost -= 50;
+  } else if (days >= 3) {
+    totalCost -= 20;
```
anstsot, line 37: please create variables for `50` and `20`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  let totalCost = days * costPerDay;
+
+  if (days >= 7) {
+    totalCost -= 50;
```
anstsot, line 35: you can immediately return needed value
```suggestion
    return totalCost - 50;
```

anstsot (changes requested):
It’s almost done! Let’s improve a little

VitaliyBondarenko1982 (approved):

PR #2624, opened 2023-04-19
Solution

katzamkova:


IrynaBeznoshchenko (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentDay = 40;
+  const total = rentDay * days;
+
+  if (days >= 7) {
+    return total - 50;
```
dashaprotsko, line 35: Use a constant variable for the discount amount (e.g. `const bigDiscount = 50;`), so it's easier to modify and understand the code in the future 🚀

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  }
+
+  if (days >= 3) {
+    return total - 20;
```
dashaprotsko, line 39: Use a constant variable for the discount amount here as well (e.g. `const smallDiscount = 20;`), to make the code more maintainable and readable 👍

dashaprotsko (approved):
Work is approved, but pay attention to comments

PR #2626, opened 2023-04-20
Solution

artpetrenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = days * 40;
+
+  if (days > 2 && days < 7) {
+    totalAmount = totalAmount - 20;
+  } else if (days > 6) {
+    totalAmount = totalAmount - 50;
+  }
```
```suggestion
  if (days > 2 && days < 7) {
    return totalAmount - 20;
  }
  
  if (days > 6) {
    return totalAmount - 50;
  }
```

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  let totalAmount = days * 40;
+
+  if (days > 2 && days < 7) {
+    totalAmount = totalAmount - 20;
+  }
+
+  if (days > 6) {
+    totalAmount = totalAmount - 50;
+  }
```
anstsot, line 39: please create variables for `50` and `20`

anstsot (changes requested):
It’s almost done! Let’s improve a little

src/calculateRentalCost.js
artpetrenko: Interesting that someone cheked solution and asks to fix one mistake, and the other mentor asks to correct the mistakes in his opinion :) want to say - who's next? ))

SanyaBratashchuk (approved):

PR #2628, opened 2023-04-20
Solution

anpiet:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * 40;
```
lOdwrotMail, line 31: I think result is not the best name for this variable - in rturn statement you are not return result directly, but modified value of it.
What is "40" - maybe it would be better to extract it to named constant and move it above function body.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days >= 7) {
+    return result - 50;
+  } else if (days >= 3) {
+    return result - 20;
```
choeqq, line 36: I think it would be useful to extract every number into separate `const` variable - I don't really know what 7 or 50 means, please provide variables like `const DAILY_RATE = 40` etc. - [checklist ](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)point 2

choeqq (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const smallDiscount = 20;
+  const basicDiscount = 50;
+  const rentalCost = days * dayPrice;
```
JoyCoffeeAddict, line 34: You should still capitalize those constants, because they should never change in your code, they are like math consts e.g. `Math.PI` , `Math.SQRT2`

JoyCoffeeAddict (changes requested):

JoyCoffeeAddict (approved):

PR #2629, opened 2023-04-20
Solution

Pa1eOrc:


loralevitska (changes requested):
Almost done) Try to avoid magic numbers. All numbers that are unclear from the context should be put into constants with a name that is understandable in terms of content

src/calculateRentalCost.js
```diff
@@ -26,9 +26,24 @@
...
+    sale = 20;
+    sum = (carRent * days) - sale;
+  }
+
+  if (days >= 7) {
+    sale = 50;
+    sum = (carRent * days) - sale;
+  }
+
+  return sum;
```
ivanchuk-yurii, line 46:
```suggestion
  return carRent * days - sale;
```

ivanchuk-yurii (changes requested):

DarkMistyRoom (approved):

PR #2637, opened 2023-04-23
Develop

Constantin-Haranskiy:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  if (days >= 7) {
+    result = days * 40 - 50;
```
TarasHoliuk, line 34:
```suggestion
    return days * 40 - 50;
```
the result variable is redundant, you may return the calculated value immediately

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+    return result;
+  } else if (days >= 3) {
```
```suggestion
  }

  if (days >= 3) {
```
when you return inside an `if` block you don't need `else` - it increases readability

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+    result = days * 40 - 20;
+
+    return result;
+  }
+  result = days * 40;
```
TarasHoliuk, line 42: return it immediately
also, would be nice to move numbers 20m 40, and 50 to variables with meaningful names

TarasHoliuk (changes requested):

ivanchuk-yurii (approved):

PR #2638, opened 2023-04-23
Solution

Oklsana57:


AndreievM (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
+  let rentalCost;
+
+  if (days >= 1 && days < 3) {
+    rentalCost = total;
+  }
+
+  if (days >= 3 && days < 7) {
```
Volodymyr-Mandryka, line 38: Instead of checking if `days >= 3` again, you can use an `else if` statement to make the code more efficient and easier to read.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    rentalCost = total - 20;
+  }
+
+  if (days >= 7) {
```
Volodymyr-Mandryka, line 42: Same as above, use an `else if` statement instead of checking if `days >= 7` again.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * 40;
+  let rentalCost;
+
+  if (days >= 1 && days < 3) {
```
Volodymyr-Mandryka, line 34: You can simplify the condition by just checking if `days < 3`. There's no need to check if `days >= 1` since `days` should always be a positive number.

PR #2640, opened 2023-04-24
Solution

6houston6:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  let totalCost = days * 40;
+
+  if (days >= 7) {
+    totalCost -= 50;
```
Anastasiia-Svintsova, line 34: It's better not to use mystery numbers and move 50 and 20 to variables.

Anastasiia-Svintsova (approved):

PR #2642, opened 2023-04-24
Update calculateRentalCost.js

vadim-ilchenko:


vadim-ilchenko (commented):
test

src/calculateRentalCost.js
```diff
@@ -28,7 +28,7 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days * 31;
```
ivadimko, line 31: The rental cost calculation is not correct. Follow the task guidelines to implement the correct calculation based on the number of rental days.

vadim-ilchenko (commented):
(((

PR #2644, opened 2023-04-24
Solution

VolodymyrPst:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days < 3) {
```
mykhalenych, line 32: don't use magic numbers, create variables and use instead

mykhalenych (changes requested):

PR #2648, opened 2023-04-25
Solution

generatedman:


33Orange (changes requested):
Well done!

But let's try to improve your code:

- Remove repeating sections of code where days are multiplied by cost.
- Remove `else if` and `else` construction. Try to use `if` only.

```
if (smth) {
    return smth
 } 
 if (smth) {
    return smth;
 } 
 return smth default;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const carCost = 40;
+
+  if (days >= 7) {
+    return carCost * days - 50;
```
Anna-Tkachenko, line 34: Don't repeat yourself: you've repeated the `cardCost * days` logic 3 times.

Also check point 1 in checklist. For now it's not clear what is `40`, `50` and `20`

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2652, opened 2023-04-26
Solution JS_Task-Transportation-On-Vacation

vintwp:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    totalDiscount = discountSevenDays;
+  }
+
+  if (days >= 3 && days < 7) {
+    totalDiscount = discountThreeDays;
+  }
+
+  rentTotal = days * rentDaily - totalDiscount;
+
+  return rentTotal;
```
Viktor-Kost, line 47:
```suggestion
  const rentalCost = rentDaily * days;

  if (days >= 7) {
    return rentalCost - discountSevenDays
  }

  if (days >= 3 && days < 7) {
    return rentalCost - discountThreeDays  
  }

  return rentalCost;
```

Viktor-Kost (changes requested):

vlad-demchuk (approved):

PR #2653, opened 2023-04-27
calculateRentalCost solution

DidenkoDenis306:


generatedman (approved):
good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  const cost = days * 40;
+
+  switch (true) {
+    case days >= 3 && days < 7:
+      return cost - 20;
+    case days >= 7:
+      return cost - 50;
+    default:
+      return cost;
+  }
```
naziamoff, line 40: Switch (true) is kind of a crutch. Let's try to rewrite it to `if` statements

naziamoff (changes requested):
not bad, but let's change the way of implementation

33Orange (approved):
Well done!

For self improve:
You can make all hardcode values ​​into constants (price multiplier, discounts) so that they can be easily changed in the future.
```
const multiplier = 40;
const cost = days * multiplier;
```

naziamoff (approved):

PR #2655, opened 2023-04-27
calculate Rental Cost function was implemented

dmytro-popovych:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40 * days;
+  let finalRentalCost = rentalCost;
+
+  if (days >= 3 && days < 7) {
+    finalRentalCost = rentalCost - 20;
+  } else if (days >= 7) {
+    finalRentalCost = rentalCost - 50;
+  }
```
tetiana-ivanchenko, line 38:
```suggestion
  if (days >= 3 && days < 7) {
    return rentalCost - 20;
  } 
  
  if (days >= 7) {
    return rentalCost - 50;
  }
```

tetiana-ivanchenko (changes requested):

PR #2659, opened 2023-04-28
calculateRentalCost function implemented

dmytro-popovych:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40 * days;
+
+  if (days >= 3 && days < 7) {
+    return rentalCost - 20;
+  } else if (days >= 7) {
+    return rentalCost - 50;
+  }
```
Volodymyr-Melnyk-Fron10der, line 37: 1. make it clearer.
2. move all magic numbers into a constants
```suggestion
  if ( && days >= 7) {
    return rentalCost - biggerDiscount;
  }
  
    if (days >= 3) {
    return rentalCost - smallerDiscount;
  }
```

Volodymyr-Melnyk-Fron10der (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const rentalCost = 40 * days;
+  const smallerDiscount = 20;
+  const biggerDiscount = 50;
+
+  if (days >= 3 && days < 7) {
+    return rentalCost - smallerDiscount;
+  } else if (days >= 7) {
+    return rentalCost - biggerDiscount;
+  }
```
```suggestion
  } 
  
  if (days >= 7) {
    return rentalCost - biggerDiscount;
  }
```
it's better to avoid `else` and `else if` when it's possible (e.g. you return smth inside if block, as you do it here) - FYI

TarasHoliuk (approved):

PR #2660, opened 2023-04-29
Solution

Daniil933:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentalCost = 0;
+
+  rentalCost = days * 40;
+
+  if (days >= 3) {
+    rentalCost = (days * 40) - 20;
```
agniya-i, line 36: it better to return value from this block, not reassign variable 
[Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) p2,3

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  }
+
+  if (days >= 7) {
+    rentalCost = (days * 40) - 50;
```
agniya-i, line 40: create variables for `20`, `40`, `50` values and give them descriptive names

agniya-i (changes requested):

PR #2664, opened 2023-04-30
Solution

Vepr984:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
   // write code here
+  const result = days * 40;
+
+  if (days < 3) {
+    return result;
+  } else if (days >= 3 && days < 7) {
+    return (result - 20);
+  } else if (days >= 7) {
+    return (result - 50);
+  }
```
etojeDenys, line 40:
```suggestion

if (days >= 7) {
    return (result - 50);
  }

 if (days >= 3) {
    return (result - 20);
  }
  
    return result;
  
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const result = days * 40;
+
+  if (days < 3) {
+    return result;
+  } else if (days >= 3 && days < 7) {
+    return (result - 20);
+  } else if (days >= 7) {
+    return (result - 50);
```
etojeDenys, line 39: also move these magic numbers in some vars

etojeDenys (changes requested):

tetiana-ivanchenko (approved):

PR #2665, opened 2023-04-30
add task solution

maksym-nemera:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  let price = 0;
+
+  for (let i = 0; i < days; i++) {
+    price += rent;
+  }
```
ruslanpashkov, line 36: let's get rid of this loop and simplify the complexity of the algorithm to linear

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  let price = 0;
```
ruslanpashkov, line 32: i think we can give clearer names to our variables

ruslanpashkov, line 32: checklist p2

> [CODE STYLE] - Avoid reassigning variables values. It's better to use const instead of let.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  for (let i = 0; i < days; i++) {
+    price += rent;
+  }
+
+  if (days >= 3 && days < 7) {
+    return price - 20;
```
ruslanpashkov, line 39: what if our discount changes? (checklist p1)

> [VARIABLES] - use variables for the main values so that you can reuse them and give them descriptive names.

ruslanpashkov (changes requested):

src/calculateRentalCost.js
NickMitchell61: @ruslanpashkov price changed accord the days, it doesn't work with const variable

NickMitchell61: @ruslanpashkov fixed the loop (recommended above) and I got what you meant :)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  const totalPrice = days * rentOneDay;
+
+  if (days >= 3 && days < 7) {
+    return totalPrice - discountThreeDays;
+  } else if (days >= 7) {
+    return totalPrice - discountSevenDays;
+  }
+
+  return totalPrice;
```
ruslanpashkov, line 43: it is not best practice to use `else if` statements here, let's use each check separately

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentOneDay = 40;
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
+
+  const totalPrice = days * rentOneDay;
+
+  if (days >= 3 && days < 7) {
```
ruslanpashkov, line 37: do we need to do such complicated checks? let's `return` value with the largest discount by default and cases with a basic discount and without it in the conditions, so we can not specify the maximum number of days, and check only on whether it's less than we need or not

also, in this case, it will be much easier for us to expand functionality if we have new rental discounts because we need to add only one simple condition after the others

ruslanpashkov (changes requested):

ruslanpashkov (approved):
🔥

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentOneDay = 40;
```
naziamoff, line 31: Consider using more descriptive variable names, such as 'dailyRate' instead of 'rentOneDay'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const discountThreeDays = 20;
+  const discountSevenDays = 50;
```
naziamoff, line 33: Consider using more descriptive variable names, such as 'sevenDaysDiscount' instead of 'discountSevenDays'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentOneDay = 40;
+  const discountThreeDays = 20;
```
naziamoff, line 32: Consider using more descriptive variable names, such as 'basicDiscount' instead of 'discountThreeDays'.

naziamoff (approved):
good job! Slight naming suggestions

PR #2666, opened 2023-04-30
Solution

nadiiachar:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+
+  if (days > 0 && days < 3) {
+    return days * 40;
```
Ivan-Shvorak, line 34: It would be nice to create variables for rent price(40), small(20) and big discounts(50) and use them instead of hardcoded values

Ivan-Shvorak (changes requested):
Please check the comment

src/calculateRentalCost.js
nadiiachar: I made changes as you requested, and now it looks more useful

Ivan-Shvorak (approved):
Good job!

PR #2670, opened 2023-05-02
solution

TommasoGiacopini:


qoa11a (changes requested):
Good job! 👍 
The code style could be better though:
- Move numbers to the variables so their names describe what they represent.
- You don't need the last `if` statement.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return pricePerDay * days;
+  }
+
+  if (days >= 7) {
+    return pricePerDay * days - 50;
+  }
+
+  if (days >= 3) {
+    return pricePerDay * days - 20;
+  }
```
Anastasiia-Svintsova, line 43: You use `pricePerDay * days` in each case. So, move it to a variable

Anastasiia-Svintsova, line 43: You need only 2 if statements here. The function will stop after the return statement anyway.
```
if (...) {
  return ...
}

if (...) {
  return ...
}

return ...

**Review by Anastasiia-Svintsova — CHANGES_REQUESTED** [review]

File: src/calculateRentalCost.js [hunk 2]
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days < 3) {
+    return totalPrice;
+  }
+
+  if (days < 7) {
+    return totalPrice - 20;
+  }
+
+  return totalPrice - 50;
```
anstsot, line 42: it would be good to create separate variables for numbers: 20 and 50

anstsot (approved):
Great job, approved!

PR #2671, opened 2023-05-02
Solution

boda4ek:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
murat-yesimbietov, line 32: Create const variables for such magic values as 40, 20, 50
For example `const dailyCost = 40;`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+  } else if (days < 7) {
```
murat-yesimbietov, line 33: I wouldn't use else if structure in this case because every if closure has return statement. Use simple if
for example:
If (condition1) { return value1 }
if (condition2) {return value2 }

murat-yesimbietov (changes requested):
Little changes are required. Also, remove the package.json and package-lock.json files from the PR

murat-yesimbietov (approved):

PR #2672, opened 2023-05-02
calculate rental cost

pos1bl:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const price = 40;
+
+  sum = days * price;
+
+  if (days >= 3) {
+    sum -= 20;
+  }
+
+  if (days >= 7) {
+    sum -= 30;
```
vlad-demchuk, line 41: move 20, 30 to variables as well

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #2676, opened 2023-05-04
task done

zisuan-od:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let primaryCarCost = 0;
+
+  for (let i = 0; i < days; i++) {
+    primaryCarCost += 40;
```
vlad-demchuk, line 35: you do not need this loop

you can reach the same using days * 40
do not increase the complexity

src/calculateRentalCost.js
```diff
@@ -29,6 +29,19 @@
...
+  }
+
+  if (days >= 7) {
+    return (primaryCarCost - 50);
+  } else if (days >= 3) {
+    return (primaryCarCost - 20);
```
vlad-demchuk, line 41: move all magic numbers(40, 20, 20) to variables as well

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #2678, opened 2023-05-04
Car rent

levchenko-dmytro:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  if (days < 3) {
+    totalRent = days * rent;
+  }
+
+  if (days >= 3) {
+    totalRent = days * rent - 20;
+  }
+
+  if (days >= 7) {
+    totalRent = days * rent - 50;
```
vlad-demchuk, line 43: move all magic numbers(20, 50) to variables

vlad-demchuk (changes requested):

OleziO (approved):

PR #2680, opened 2023-05-05
calculateRentalCost

Dorosh90:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  const basicDiscount = 20;
+  const fullDiscount = 50;
+  let total = 0;
```
```suggestion
  const basicPrice = rent * days;
```
It would be great to create a variable for `rent * days` in order not to duplicate yourself.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+  if (days < 3) {
+    total += days * rent;
+  }
+
+  if (days >= 3 && days < 7) {
+    total += days * rent - basicDiscount;
```
```suggestion
    return basicPrice - basicDiscount;
```
There is no need to store the price, just return the full price if condition is true

polosanya (changes requested):
Almost perfect!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const fullDiscount = 50;
+  const basicPrice = rent * days;
+
+  if (days < 3) {
+    return basicPrice;
+  } else if (days >= 3 && days < 7) {
+    return basicPrice - basicDiscount;
+  } else if (days >= 7) {
+    return basicPrice - fullDiscount;
+  }
```
tetiana-ivanchenko, line 42:
```suggestion
  } 
  
  if (days >= 3 && days < 7) {
    return basicPrice - basicDiscount;
  }
  
  return basicPrice - fullDiscount;
  }
```

tetiana-ivanchenko (approved):

PR #2681, opened 2023-05-05
Solution

andrusb21:


33Orange (changes requested):
Well done!

But let's try to improve your code:

- Remove `else if` and `else` construction. Try to use `if` only.

```
if (smth) {
    return smth
} 
if (smth) {
    return smth;
} 
return smth default;
```

Anna-Tkachenko (changes requested):
Check point 1 in checklist. For now it's not clear what is `40`, `20` and `50`

Anna-Tkachenko (approved):
⭐️

PR #2683, opened 2023-05-05
Solution

marygurina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentalCost = 0;
+
+  if (days < 3) {
+    rentalCost = days * 40;
```
murat-yesimbietov, line 34: To improve code maintenance and readability, create const variable for such value, as 40, 20, 50

murat-yesimbietov (changes requested):

PR #2684, opened 2023-05-05
Solution

Ivankaaaaaa:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return 'error';
+  }
+
+  if (days < 3) {
+    return days * 40;
+  } else if (days >= 3 && days < 7) {
+    return days * 40 - 20;
+  } else if (days >= 7) {
+    return days * 40 - 50;
+  }
```
Volodymyr-Melnyk-Fron10der, line 41: use early return pattern
https://gomakethings.com/the-early-return-pattern-in-javascript/

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days <= 0) {
+    return 'error';
+  }
+
+  if (days < 3) {
+    return days * 40;
```
Volodymyr-Melnyk-Fron10der, line 36: move all magic numbers like 40 .. to the constants

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days <= 0) {
+    return 'error';
```
Volodymyr-Melnyk-Fron10der, line 32: throw a real Error (return new Error('error text'))

Volodymyr-Melnyk-Fron10der (changes requested):

src/calculateRentalCost.js
Ivankaaaaaa: Is it really necessary for such a small function(not including 40)?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costs = 40;
+
+  if (days <= 0) {
```
etojeDenys, line 33: Instead of returning an error message, throw an error with 'throw new Error("Invalid days value");'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return 'error text';
+  }
+
+  if (days < 3) {
+    return days * costs;
+  } else if (days >= 3 && days < 7) {
+    return days * costs - 20;
+  } else if (days >= 7) {
+    return days * costs - 50;
```
etojeDenys, line 42: Remove the unnecessary condition 'days >= 7' since it's already checked in the previous if statement.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costs = 40;
```
etojeDenys, line 31: The variable name 'costs' should be changed to 'costPerDay' for better readability.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days <= 0) {
+    return 'error text';
+  }
+
+  if (days < 3) {
+    return days * costs;
+  } else if (days >= 3 && days < 7) {
```
etojeDenys, line 39: Remove the unnecessary condition 'days >= 3' since it's already checked in the previous if statement.

etojeDenys (changes requested):

src/calculateRentalCost.js
```diff
@@ -9,7 +9,7 @@
  * You will need a rental car in order for you to get around in your vacation.
  * The manager of the car rental makes you some good offers.
  *
- * Every day you rent the car costs $40.
+ * Every day you rent the car costPerDay $40.
  * If you rent the car for 7 or more days, you get $50 off your total.
  * Alternatively, if you rent the car for 3 or more days,
```
OleziO, line 14: Remove all comments.

src/calculateRentalCost.js
OleziO: Not resolved.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const costPerDay = 40;
+
+  if (days <= 0) {
+    return 'Invalid days value';
+  }
+
+  if (days < 3) {
+    return days * costPerDay;
+  } else if (days < 7) {
+    return days * costPerDay - 20;
```
OleziO, line 40: Move all discounts to the variables.

OleziO (changes requested):

src/calculateRentalCost.js
```diff
@@ -1,34 +1,21 @@
...
- */
-
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const discountSinceThreeDays = 20;
+  const discountSinceSevenDays = 50;
+
+  if (days <= 0) {
+    return 'Invalid days value';
```
ivanchuk-yurii, line 9:
```suggestion
    return new Error('Invalid days value');
```

src/calculateRentalCost.js
```diff
@@ -1,34 +1,21 @@
...
+  }
+
+  if (days < 3) {
+    return days * costPerDay;
+  } else if (days < 7) {
+    return days * costPerDay - discountSinceThreeDays;
+  } else {
+    return days * costPerDay - discountSinceSevenDays;
+  }
```
ivanchuk-yurii, line 18:
```suggestion
  if (days < 7) {
    return days * costPerDay - discountSinceThreeDays;
  }
  
  return days * costPerDay - discountSinceSevenDays;
```

ivanchuk-yurii (changes requested):

sTorba24 (approved):
Well done!

PR #2685, opened 2023-05-06
added solution

niaYaro:


KatyaZaremska (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,11 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days < 3
+    ? days * 40
+    : days < 7
+      ? days * 40 - 20
+      : days * 40 - 50;
```
Anna-Tkachenko, line 35: That's great that you can use ternary operator here, but it's not the case for it's usage

Let's rewrite it using if / else statement, and check checklist points about variables creation

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2690, opened 2023-05-07
Solution

reseneweb:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const daysPrice = days * price;
```
polosanya, line 32: I think it would be great to specify that the first variable is a rate for one day, and the second is the basic price without discounts

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+
+  if (days < 3) {
+    return daysPrice;
+  }
+
+  if ((days >= 3) && (days < 7)) {
+    return daysPrice - 20;
+  }
+
+  return daysPrice - 50;
```
polosanya, line 42: Let's create variables for 20 and 50 to avoid magic numbers

polosanya (changes requested):
Almost done!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const threeDayDiscount = 20;
+  const sevenDayDiscount = 50;
+
+  if (days < 3) {
+    return priceWithoutDiscounts;
+  }
+
+  if ((days >= 3) && (days < 7)) {
+    return priceWithoutDiscounts - threeDayDiscount;
+  }
```
sTorba24, line 42:
```suggestion
  if (days < 7) {
    return priceWithoutDiscounts - threeDayDiscount;
  }
```

sTorba24: you already know that value >=3(because in other case first condition will end the program)

sTorba24 (approved):
Well done!

PR #2691, opened 2023-05-07
add solution

growingananas:


julshvets (approved):
Good job!

silchyn (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
+  const basicD = 20;
+  const maxD = 50;
```
silchyn, line 33: what is `D` here?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const minDays = 3;
```
silchyn, line 34: what means `minDays`?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = days * 40;
```
silchyn, line 31: cost for what?

src/calculateRentalCost.js
growingananas: Have a look! I have updated it)

src/calculateRentalCost.js
growingananas: It stands for  minimum number of days

silchyn (approved):

PR #2693, opened 2023-05-07
Solution

vlkzmn:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costsPerDay = 40;
+  const discount = 20;
+  const longDiscount = 50;
```
polosanya, line 33: I think it's better to name it big/large discount instead of `long`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return days * costsPerDay - longDiscount;
+  } else if (days >= 3) {
+    return days * costsPerDay - discount;
```
polosanya, line 38: image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  return days * costsPerDay;
```
polosanya, line 41: It would be great to create a variable for `days * costsPerDay` as it duplicates three times

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costsPerDay = 40;
```
polosanya, line 31:
```suggestion
  const costPerDay = 40;
```

polosanya (changes requested):
Almost done!

polosanya (changes requested):
Pay attention to the checklist

OleziO (approved):

PR #2696, opened 2023-05-07
 Solution

DenisGurskiy:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
   // write code here
+  const costs = 40;
+  let amount = 0;
+
+  if (days < 3) {
+    amount = days * costs;
+  } else if (days < 7) {
+    amount = days * costs - 20;
+  } else {
+    amount = days * costs - 50;
```
vlad-demchuk, line 40: move all magic numbers to variables (20, 50)

vlad-demchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+
+  if (days < 3) {
+    amount = days * costs;
+  } else if (days < 7) {
+    amount = days * costs - discount1;
+  } else {
+    amount = days * costs - discount2;
+  }
+
+  return amount;
```
Viktor-Kost, line 45: 1) it'd better to use early return
2) instead of discount1 and discount2 we can use more meaningful names
```suggestion
  const rentalCost = days * 40
  const baseDiscount = 20;
  const proDiscount = 50;
  
  if (days >= 7) {
    return rentalCost - proDiscount;
  }
  
  if (days >= 3) {
    return rentalCost - baseDiscount;
  } 
  
  return rentalCost 
  ```

Viktor-Kost (changes requested):
Good job!
We can improve a little bit more

vlad-demchuk (approved):

PR #2700, opened 2023-05-08
task done

artem-ivashchenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicPrice = 40;
+  const totalPrice = basicPrice * days;
+
+  if (days >= 7) {
+    return totalPrice - 50;
```
OleziO, line 35: Move all discounts to the variables.

OleziO (changes requested):

OleziO (approved):

PR #2707, opened 2023-05-08
Develop

opiquem:
Solution

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -32,11 +32,7 @@ function calculateRentalCost(days) {
   const alternativeDicount = 20;
   const basicCost = 40;
 
-  let sum = 0;
-
-  for (let i = 1; i <= days; i++) {
-    sum += basicCost;
-  }
+  let sum = basicCost * days;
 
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
-  // write code here
+  const basicDiscount = 50;
+  const alternativeDicount = 20;
+  const basicCost = 40;
+
+  let sum = 0;
+
+  for (let i = 1; i <= days; i++) {
+    sum += basicCost;
+  }
```
vladyslav-lapshyn, line 39: mb you can just multiplication `days` on `basicCost` without `for`

vladyslav-lapshyn (approved):
GJ 🔥

src/calculateRentalCost.js
Anna-Tkachenko: Agree, check @vladyslav-lapshyn comment, we can avoid using loops here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2708, opened 2023-05-08
Solution

maryanmats:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,18 @@
...
+  const cost = days * 40;
+  let result = 0;
+
+  if (days >= 7) {
+    result = cost - 50;
+  } else if (days >= 3) {
+    result = cost - 20;
+  } else {
+    result = cost;
+  }
```
YeMyr, line 41: Hi! Please, take a look at the checklist.
the result should be returned immediatly in ```if``` statement.
image

YeMyr (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,26 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const costDay = 40;
+  const weekDiscount = 50;
+  const threeDayDiscount = 20;
+  let total = 0;
+
+  if (days >= 7) {
```
silchyn, line 37: what is `7` here?

silchyn (changes requested):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,28 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const costDay = 40;
+  const weekDiscount = 50;
+  const threeDayDiscount = 20;
```
```suggestion
  const bigDiscount = 50;
  const smallDiscount = 20;
```
It is recommended not to connect the variable's name with values that can change later. For example. we might need to change the requirement for small discount from three days to two days later

src/calculateRentalCost.js
```diff
@@ -29,6 +29,28 @@
...
+  const week = 7;
+  const threeDay = 3;
```
```suggestion
  const daysForBigDiscount = 7;
  const daysForSmallDiscount = 3;
```
Same here

src/calculateRentalCost.js
```diff
@@ -29,6 +29,28 @@
...
+  let total = 0;
```
```suggestion
  const total = costDay * days;
```
image

src/calculateRentalCost.js
```diff
@@ -29,6 +29,28 @@
...
+
+  if (days >= week) {
+    total = costDay * days - weekDiscount;
```
```suggestion
    return total - bigDiscount;
```
There is no need to reassign. Just return the full price

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments 🙂

silchyn (approved):

PR #2710, opened 2023-05-08
solution

thienlaoo:


vitvadym (approved):
Good job 😎 
btw check points 4,5 in check list) 👀

Anna-Tkachenko (changes requested):
Check all points from checklist 🐝

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const carCost = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return carCost * days;
+  } else if (days >= 3 && days < 7) {
+    return (carCost * days) - smallDiscount;
```
Anna-Tkachenko, line 38: Check point 3 in checklist

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
thienlaoo: I dont use here 'else' with return statement, but I need to use 'else if' with return because of task condition, please correct me if Im mistaken

naziamoff: `else/if` looks really huge and decreases the readability of your code. We don't use `else` or `else/if` unless we really need to.

So you can do it this way:
```suggestion
  if (days < 3) {
    return carCost * days;
  }  
  
  if (days >= 3 && days < 7) {
    return (carCost * days) - smallDiscount;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  return (carCost * days) - bigDiscount;
```
naziamoff, line 41: that looks strange that bigDiscount is default.
Let's start with checking for a big discount - and as a default we will just return totalPrice

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const carCost = 40;
```
naziamoff, line 31: I would say `dailyCost`

naziamoff (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  const dailyCost = days * 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days >= 7) {
+    return dailyCost - bigDiscount;
+  }
+
+  if (days >= 3 && days < 7) {
```
naziamoff, line 39: will work just fine without the second check
```suggestion
  if (days >= 3) {
```

naziamoff (approved):

PR #2714, opened 2023-05-08
Solution

dimoshen:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
+    return days * 40 - 20;
+  }
+
+  return days * 40;
```
bohdan-mykhailenko, line 39: It might be better to not calculate just certain cases with certain numbers, because values can change and all functionality needs to rewrite.

bohdan-mykhailenko (approved):
Good job

silchyn (changes requested):
What do all of these magic numbers mean?

src/calculateRentalCost.js
dimoshen: sure. i've added variables 👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentPrice = 40;
+  const smallDiscount = 20;
+  const largeDiscount = 50;
+
+  if (days >= 7) {
```
silchyn, line 35: what is `7` here?

silchyn (changes requested):

src/calculateRentalCost.js
dimoshen: created variables

dinaminahi (approved):

silchyn (approved):

PR #2716, opened 2023-05-08
Solution

Yuliia-kucheriavenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
-  // write code here
+  const dayPrice = 40;
+  const miniDiscount = 20;
+  const basicDiscount = 50;
+
+  let totalAmount = 0;
+
+  if (days < 3) {
+    totalAmount = days * dayPrice;
+  } else if (days >= 3 && days < 7) {
```
annholovko, line 39: it is not necessary  to use `days >=3` as we checked whether the number is smaller than 3 in first condition, so this condition is already checking number from 3 to 7 without `days >=3`

annholovko (approved):

src/calculateRentalCost.js
Anna-Tkachenko: check @annholovko comment

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dayPrice = 40;
+  const miniDiscount = 20;
+  const basicDiscount = 50;
+
+  let totalAmount = 0;
+
+  if (days < 3) {
+    totalAmount = days * dayPrice;
```
Anna-Tkachenko, line 38: Don't repeat yourself: you've repeated this calculation `days * dayPrice` 3 times, let's use it only once

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
annholovko: good job!🔥
left one suggestion

Anna-Tkachenko (changes requested):
You didn't fix my comment 🐝

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const dayPrice = 40;
+  const miniDiscount = 20;
+  const basicDiscount = 50;
+  const cost = days * dayPrice;
+
+  if (days >= 7) {
+    return cost - basicDiscount;
+  } else if (days >= 3) {
+    return cost - miniDiscount;
+  }
```
Anna-Tkachenko, line 40: Check point 3 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2717, opened 2023-05-08
made calculateRental task

vladyslav-lapshyn:


33Orange (approved):
Well done! 🔥 

For self improve:
You can make all hardcode values ​​into constants (price multiplier, discounts) so that they can be easily changed in the future.

```
const multiplier = 40;
const cost = days * multiplier;
```

PR #2719, opened 2023-05-08
(ym) add solution

YeMyr:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const perDay = 40;
+  const weekDiscount = 50;
+  const threeDayDiscount = 20;
+  let total = 0;
+
+  if (days < 3) {
+    total = days * perDay;
```
dinaminahi, line 37: To simplify the code and calculations, you can firstly calculate the total cost and in the `if` statements return the value with discount

` days * perDay` is repeating 3 times

dinaminahi, line 37: To simplify the code and calculations, you can firstly calculate the total cost and in the if statements return the value with discount

 `days * perDay` is repeating 3 times

dinaminahi (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const perDay = 40;
+  const weekDiscount = 50;
+  const threeDayDiscount = 20;
+  const dayCost = perDay * days;
+
+  if (days < 3) {
```
silchyn, line 36: what is `3`?

silchyn (changes requested):

src/calculateRentalCost.js
YeMyr: ok, I understood what you mean 👌
Please, check the corrections

33Orange (approved):
Well done ! 🔥

silchyn (approved):

PR #2725, opened 2023-05-08
solution

liza-strykharchuk:


Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days < 3) {
+    return result;
+  }
+
+  if (days < 7) {
+    return result - twenty;
+  }
+
+  return result - fifty;
```
Viktor-Kost, line 44: it'd be better to exchange like this
```suggestion
  if (days >= 7) {
    return rentalCost - proDiscount;
  }

  if (days >= 3) {
    return rentalCost - baseDiscount;
  }

  return rentalCost;
```

Viktor-Kost (changes requested):
Good job!
Looks much better. 
Some minor fixed needed

Viktor-Kost (approved):
Perfect!

nazarmatsevych (approved):

nazarmatsevych (changes requested):
* test

PR #2726, opened 2023-05-08
Solution

RoksolanaTylipska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicDiscount = 20;
+  const extraDiscount = 50;
+  const rentCosts = 40;
```
33Orange, line 33: You can place this variables outside of the function.
Now they are created anew with each new function call.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  let total = days * rentCosts;
+
+  if (days >= 3) {
+    total = (days * rentCosts) - basicDiscount;
+  };
+
+  if (days >= 7) {
+    total = (days * rentCosts) - extraDiscount;
+  };
```
33Orange, line 43: You should return the value immediately in `if` statements without overwriting the variable.

```
if (condition) {
    return x;
};
```

33Orange (changes requested):

src/calculateRentalCost.js
```diff
@@ -26,9 +26,22 @@
...
  */
+const basicDiscount = 20;
+const extraDiscount = 50;
+const rentCosts = 40;
 
 function calculateRentalCost(days) {
-  // write code here
+  const total = days * rentCosts;
+
+  if (days >= 3 && days < 7) {
```
silchyn, line 36: what are `3` and `7`?

silchyn (changes requested):

src/calculateRentalCost.js
```diff
@@ -26,9 +26,22 @@
  *
  * @return {number}
  */
+const basicDiscount = 20;
+const extraDiscount = 50;
+const rentCosts = 40;
```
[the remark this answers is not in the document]
RoksolanaTylipska: image
I had comments about  this

33Orange (approved):
Good job

silchyn (approved):

PR #2727, opened 2023-05-08
solution

kirilshorin:


33Orange (changes requested):
- `Use variables for the main values so that you can reuse them and give them descriptive names.` For now it's not clear what is 40, 50 and 20.
- It will be clearer and better to always return 1 default value and 2 values ​​when some conditions are met.

```
if (condition) {
  return a;
}
if (condition) {
  return b;
}
return c;
```

silchyn (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (typeof days !== 'number' || !Number.isInteger(days) || days < 1) {
+    throw new Error('The input parameter must be a positive integer');
+  }
+
+  const dailyCost = 40;
```
33Orange, line 35: Consider using a constant for the minimum number of days for the basic discount (3) and the extra discount (7).

33Orange (changes requested):
Good job!

But you don't fix problem from last comment.
You need to create variables for minimum number of days for the basic discount and for the extra discount.

naziamoff (approved):
👍

silchyn (approved):

PR #2728, opened 2023-05-08
Task calculated

olehmahera:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40 * days;
+
+  return days > 6
+    ? cost - 50
+    : days > 2
+      ? cost - 20
+      : cost;
```
33Orange, line 37: This is not the best place to use the ternary operator. It spoils the readability of the code in this case

33Orange (changes requested):
Good job.

A few moments:
- Use variables for the main values so that you can reuse them and give them descriptive names. For now it's not clear what is 40, 50 and 20
- Use an `if` statement to return values. It's very cool that you understand the ternary operator well, but it's not always appropriate

```
if (condition) {
  return x;
}
return y;
```

silchyn (changes requested):
Fix the comments above before requesting a review again

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentalCost = 40 * days;
+  let discount = 0;
+
+  if (days > 6) {
```
33Orange, line 34: Change the condition to 'days >= 7' to include the case when days are exactly 7.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    discount = 50;
+  } else if (days > 2) {
```
33Orange, line 36: Change the condition to 'days >= 3' to include the case when days are exactly 3.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentalCost = 40 * days;
+  let discount = 0;
```
33Orange, line 32: Use `const` instead `let`. You don't need to rewrite variables.
Just create variable for all values.  For discounts, amount of days, etc.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days > 6) {
+    discount = 50;
+  } else if (days > 2) {
+    discount = 20;
+  }
+  rentalCost -= discount;
+
+  return rentalCost;
```
33Orange, line 41: You should return value immediately in `if` condition.

```
if (condition) {
  return x;
}
return y;
```

33Orange (changes requested):
Please correct your code according to the checklist:
- [VARIABLES] - use variables for the main values so that you can reuse them and give them descriptive names.
- [CODE STYLE] - Avoid reassigning variables values. It's better to use const instead of let.
- [CODE STYLE] - You should return the value immediately in if statements. And don't use else after return statement. If return is performed - code after it won't be executed anyway.

dinaminahi (approved):
Well done 👍

silchyn (approved):

PR #2729, opened 2023-05-08
add task solution

PanyaPrimal:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let discount = 0;
+
+  if (days >= 3 && days < 7) {
+    discount = 20;
```
33Orange, line 34: Consider using a constant variable for the discount amount for 3 to 6 days.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days >= 7) {
+    discount = 50;
```
33Orange, line 36: Consider using a constant variable for the discount amount for 7 or more days.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  }
+
+  return (days * 40) - discount;
```
33Orange, line 39: Consider using a constant variable for the daily rental cost.

33Orange (changes requested):
Good job.

A few moments:
- Use variables for the main values so that you can reuse them and give them descriptive names. For now it's not clear what is 40, 50 and 20
- You should return the value **immediately** in if statements. And don't use else after return statement. If return is performed - code after it won't be executed anyway. 

```
if (condition) {
  return a;
}
if (condition) {
  return b;
}
return c;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discount3To6Days = 20;
+  const dailyRentalCost = 40;
+  const discount7OrMoreDays = 50;
```
naziamoff, line 33: - we never use numbers in the name of our variables
- what if we gonna change the policy and the discount 20 will be applied starting with the second day and ending on the fifth? You gonna need to change the variable name in all the usage places.

So let's think of more reusable names for those variables, more generic

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 3 && days < 7) {
+    return days * dailyRentalCost - discount3To6Days;
+  }
```
naziamoff, line 37: if you start checking for `if days >= 7` - you won't need to checks in this `if`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days >= 7) {
+    return days * dailyRentalCost - discount7OrMoreDays;
```
naziamoff, line 40: As days * dailyRentalCost is always a constant- I would create a separate variable for that

naziamoff (changes requested):
a couple of things left and we're good to go 🚀

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+  const baseDailyRentalCost = 40;
+  const discountZero = 0;
+  const discountBasic = 20;
+  const discountAdditional = 50;
+
+  let rentalCost = days * baseDailyRentalCost;
+  let discount = discountZero;
+
+  switch (true) {
+    case days >= 7:
```
silchyn, line 40: what is `7`?

silchyn (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
-  // write code here
+  const baseDailyRentalCost = 40;
+  const discountZero = 0;
+  const discountBasic = 20;
+  const discountAdditional = 50;
+
+  let rentalCost = days * baseDailyRentalCost;
+  let discount = discountZero;
+
+  if (days >= 7) {
```
33Orange, line 39: You can create variable for number 7. For example `daysForExtraDiscount`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+    discount = discountAdditional;
+  }
+
+  if (days >= 3 && days < 7) {
```
33Orange, line 43: Same here.

33Orange (approved):
Good job

PR #2731, opened 2023-05-08
fixed Solution

olena-bielozorova:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = 0;
```
AnnaShevchenkoQA, line 31: Consider declaring the 'rentPerDay' variable to store the daily rent cost (40) and use it in calculations for better maintainability

AnnaShevchenkoQA (approved):
Good job👍

PR #2733, opened 2023-05-08
added solution

annholovko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  const dailyRent = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  return days < 3
+    ? days * dailyRent
+    : days < 7
+      ? days * dailyRent - smallDiscount
+      : days * dailyRent - bigDiscount;
```
Anna-Tkachenko, line 39: It's awesome that you've found out how to use ternary operator, but nesting here just decrease code readability. Let's rewrite it using if / else statement

Anna-Tkachenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const totalPrice = days * dailyRent;
+
+  if (days < 3) {
+    return totalPrice;
+  } else if (days < 7) {
+    return totalPrice - smallDiscount;
+  } else {
+    return totalPrice - bigDiscount;
```
Anna-Tkachenko, line 41: Check point 3 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2738, opened 2023-05-09
solution

DaniilBarilotti:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
+    const money = days * 40 - 20;
+
+    return money;
+  }
+
+  if (days >= 7) {
+    const money = days * 40 - 50;
+
+    return money;
+  }
```
etojeDenys, line 48:
```suggestion
   const money = days * 40;

  if (days >= 7) {
    return money - 50;
  }

  if (days >= 3) {
    return money - 20;
  }
  
  return money
```

etojeDenys (changes requested):
looks complicated, let's also move magic numbers (50,20) to vars

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const money = days * 40;
```
mykhalenych, line 32: use constants instead of magic numbers 50, 40, 20, 7, 3

mykhalenych (changes requested):

Viktor-Kost (approved):
Hello!)
Nice work, yet take a look at the last mentor's comment to avoid magic numbers for 7, 3

PR #2739, opened 2023-05-09
Solution

yurii-krymskiy:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  for (let i = 1; i <= days; i++) {
+    daysCount++;
+  }
+
+  if (daysCount < 3) {
+    newPrice = price * daysCount;
+  } else if (daysCount >= 3 && daysCount <= 6) {
+    newPrice = price * daysCount - firstDiscount;
+  } else if (daysCount >= 7) {
+    newPrice = price * daysCount - secondDiscount;
```
vlad-demchuk, line 46: you can reorder conditions to avoid `&&`

vlad-demchuk: also you can return your function in if/else statements, no need to assign it to `newPrice `

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #2740, opened 2023-05-10
solution

shelyaa:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  if (days >= 1) {
+    sum = days * value;
+  }
+
+  if (days >= 3) {
+    sum = days * value - 20;
+  }
+
+  if (days >= 7) {
+    sum = days * value - 50;
```
vlad-demchuk, line 43: move all magic numbers(20, 50) to variables

also, you can return values immediately inside if statements

vlad-demchuk (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const value = 40;
+  let sum = 0;
```
nazarmatsevych, line 32: The variable 'sum' is unnecessary. You can return the calculated value directly in each condition.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+    return sum;
+  }
+
+  if (days >= 3) {
+    sum = days * value - smallDiscount;
+
+    return sum;
+  }
+
+  if (days >= 1) {
```
nazarmatsevych, line 48: The last condition 'if (days >= 1)' is not needed. You can remove it and return the value directly in the last line of the function.

nazarmatsevych (approved):

PR #2744, opened 2023-05-11
code

liza-strykharchuk:
try 2


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const baseDiscount = 20;
+  const proDiscount = 50;
+  const rentalCost = days * 40;
```
polosanya, line 33: It would be great to create a variable for `40` too

polosanya (approved):
Good job!

PR #2747, opened 2023-05-12
Solution

Spstv-sys:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const arend = days * 40;
```
PolinaRomanenko, line 31: Consider renaming the variable 'arend' to 'rent' or 'rentalCost' for better clarity.

PolinaRomanenko (approved):
One comment to consider for the future, or here)

PR #2748, opened 2023-05-12
add task solution

sasdaiv:


OleziO (approved):

OleziO (changes requested):
Move all discounts to the variables

VitaliyBondarenko1982 (approved):

PR #2750, opened 2023-05-13
add task solution

w315:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    return days * 40;
+  }
+
+  if (days >= 7) {
+    return (days * 40) - 50;
+  }
+
+  if (days >= 3) {
+    return (days * 40) - 20;
+  }
```
nazarmatsevych, line 41: Move `days * 40` and other magic numbers like `20`, `50 `and `40` to variables.

Also last if statement is redundant because you have return inside previous if's

nazarmatsevych (changes requested):

nazarmatsevych (approved):
🔥

PR #2754, opened 2023-05-15
new function

VitaDemian:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days >= 3) {
+    cost -= 20;
+  }
+
+  return cost;
 }
 
 module.exports = calculateRentalCost;
+
+calculateRentalCost(25);
```
dashaprotsko, line 44: Remove the function call at the end of the file as it is not necessary for the module export.

dashaprotsko (changes requested):
Little fix is required

PR #2756, opened 2023-05-16
Solution

kropiva:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
+  } else if (days >= 3) {
```
TarasHoliuk, line 33: no need to write `else` when you `return` inside your `if` block located above - the function will end once any value is returned

so just write 2 separate `if` blocks and return without `if` in the final case

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
```
TarasHoliuk, line 32: move "magic numbers" to meaningful variables

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #2758, opened 2023-05-16
task solved

Nazarht:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let price = 40 * days;
```
volodymyr-soltys97, line 32: Consider using a constant for the daily rental cost instead of a hardcoded value.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more sophisticated

vlad-demchuk (approved):

PR #2760, opened 2023-05-17
Solution

SerhiiVoitiuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 7) {
+    sumTotal = days * costs - discountTwo;
+  } else if (days >= 3) {
+    sumTotal = days * costs - discountOne;
+  } else {
+    sumTotal = days * costs;
+  }
+
+  return sumTotal;
```
anstsot, line 44: you can return the value in each `if` block
```suggestion
  if (days >= 7) {
    return days * costs - discountTwo;
  }
  if (days >= 3) {
    return days * costs - discountOne;
  }
  return days * costs;
```

anstsot (approved):
Good job, approved

PR #2762, opened 2023-05-18
Solution

andrii-rykal:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  let totalPrice = 0;
+
+  if (days < 3) {
```
mykhalenych, line 34: better to keep all magic numbers in constants like price, totalPrice

mykhalenych (approved):

PR #2763, opened 2023-05-18
Solution

hulchak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  let result = 40 * days;
+
+  if (days >= 7) {
+    result -= 50;
+
+    return result;
+  }
+
+  if (days >= 3) {
+    result -= 20;
```
tetiana-ivanchenko, line 40: move magic numbers in the variables

tetiana-ivanchenko (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let result = 40 * days;
+  const daysForFirstDiscount = 3;
+  const daysForSecondDiscount = 7;
+
+  if (days >= daysForSecondDiscount) {
+    result -= 50;
+
+    return result;
```
anstsot, line 38: can be simplified
```suggestion
    return result - 50;
```

anstsot (approved):
Good job, approved!

PR #2766, opened 2023-05-20
task-one

mukutiuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  if (days < 3) {
+    result = days * 40;
```
TarasHoliuk, line 34:
```suggestion
    return days * 40;
```
the `result` variable is redundant here - you may immediately return what you need

Also, move "magic numbers" (40, 20, 50) to variables with meaningful names

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+    result = (days * 40) - 20;
+
+    return result;
+  }
+
+  if (days <= 7) {
+    result = (days * 40) - 50;
+
+    return result;
+  }
```
TarasHoliuk, line 49:
```suggestion
  return days * 40 - 50;
 ```
 all other cases are covered, `if` condition is redundant here - just return

TarasHoliuk (changes requested):

TarasHoliuk (approved):

PR #2768, opened 2023-05-22
Solution

ArtemBehter:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let result = days * 40;
+
+  if (days >= 7) {
+    result = result - 50;
+  } else if (days >= 3) {
+    result = result - 20;
+  }
```
DorotaLeniecDev, line 38: According to the checklist:  _use variables for the main values so that you can reuse them and give them descriptive names._ you should creat proper variables to store: price and discrounts

DorotaLeniecDev (changes requested):
Good solution but you need to create proper variables

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let rentalCost = days * 40;
+  const standartDiscount = 20;
+  const increasedDiscount = 50;
+
+  if (days >= 7) {
+    rentalCost = rentalCost - increasedDiscount;
```
darokrk, line 37: Improvement tip 💡 :

Instead of directly assigning the updated `rentalCost`, you can consider use shorter approach, to use '-='  `rentalCost -= increasedDiscount;`

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
...
+  } else if (days >= 3) {
+    rentalCost = rentalCost - standartDiscount;
```
darokrk, line 39: Same here :)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let rentalCost = days * 40;
+  const standartDiscount = 20;
```
darokrk, line 33: Typo here, it should be `standardDiscount` rather than `standartDiscount`

src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let rentalCost = days * 40;
```
darokrk, line 32: I would suggest here to rename this variable to be more generic, e.g. `price`

darokrk (approved):
Looks good 👍🏽  approved, but take a look at my suggestions and you can improve them if you like :)

PR #2772, opened 2023-05-24
implemented solution

vladyslav10shevchenko:


anstsot (changes requested):
let's improve the code a little:
1. create variables for numbers (50, 20)
2. you can return the totalCost in `if` construction directly. then you don't need to use `else`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRentCost = 40;
+  let totalCost = days * dayRentCost;
```
IvanFesenko, line 32:
```suggestion
  const totalCost = days * dayRentCost;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days >= 7) {
+    totalCost -= bigDiscount;
+
+    return totalCost;
```
IvanFesenko, line 40: No need to reassign value, just return calculation result
```suggestion
    
    return totalCost - bigDiscount;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  }
+
+  if (days >= 3) {
+    totalCost -= smallDiscount;
+
+    return totalCost;
+  }
```
IvanFesenko, line 47: Same here

IvanFesenko (changes requested):
More improvements left

Viktor-Kost (approved):

PR #2774, opened 2023-05-24
Solution_transportation-on-vacation

Svitlanchick:
https://github.com/mate-academy/js_task-transportation-on-vacation/pull/2774

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  if (days < 3) {
+    result = days * 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    result = (days * 40) - 20;
+  }
+
+  if (days >= 7) {
+    result = (days * 40) - 50;
+  }
+
+  return result;
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,21 +28,21 @@
  */
 
 function calculateRentalCost(days) {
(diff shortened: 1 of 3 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+
+  if (days < 3) {
+    result = days * 40;
```
Anastasiia-Svintsova, line 34: You use `days * 40` not once. Move it to a variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    result = (days * 40) - 20;
```
Anastasiia-Svintsova, line 38: Don't use mystery numbers. Move `20` to a variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  }
+
+  if (days >= 7) {
+    result = (days * 40) - 50;
```
Anastasiia-Svintsova, line 42: Same

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
Anastasiia-Svintsova, line 31: You don't really need to create a variable for the result. Just return the calculated value 
```
if (...) {
  return ...
}

Anastasiia-Svintsova (changes requested):

qoa11a (approved):

PR #2775, opened 2023-05-25
solution

NataliMax:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total = 0;
+
+  if (days < 3) {
+    total = days * 40;
+  }
```
DarkMistyRoom, line 35: we can reduce the number of `if`s. Set this value as default and check if we need to apply discount

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 3 && days < 7) {
+    total = days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    total = days * 40 - 50;
+  }
```
DarkMistyRoom, line 43: Reverse order of these `if`s. Than won't need to check the second condition here `days >= 3 && days < 7`

DarkMistyRoom (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const rent = days * 40;
+  let totalRent = rent;
+
+  if (days > 2) {
+    totalRent = rent - 20;
+  }
+
+  if (days > 6) {
+    totalRent = rent - 50;
+  }
```
etojeDenys, line 40: image

etojeDenys (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days <= 2) {
+    return rent;
+  }
+
+  if (days < 7) {
+    return rent - 20;
+  }
+
+  return rent - 50;
```
VitaliyBondarenko1982, line 41: not fixed comment about magic numbers.
write all numbers into constants and use them
[checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)
Screenshot from 2023-05-26 12-30-16
Screenshot from 2023-05-26 12-30-03

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):
Well done

PR #2776, opened 2023-05-25
add solution

prostonadobobb:
first lessons JS Advanced

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
+
+  if (days >= 7) {
+    sum = 40 * days - 50;
+  } else if (days >= 3) {
+    sum = 40 * days - 20;
+  } else {
+    sum = 40 * days;
+  }
+
+  return sum;
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,14 +28,12 @@
  */
 
 function calculateRentalCost(days) {
-  let sum = 0;
+  let sum = 40 * days;
 
   if (days >= 7) {
-    sum = 40 * days - 50;
+    sum -= 50;
   } else if (days >= 3) {
-    sum = 40 * days - 20;
-  } else {
-    sum = 40 * days;
+    sum -= 20;
   }
 
   return sum;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 40 * days;
```
Viktor-Kost, line 31: 1) Consider renaming this variable to `rentalCost`
2) no needs to use let, use const

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days >= 7) {
+    sum -= 50;
```
Viktor-Kost, line 34: 1) Consider using a constant for the discount value 50 instead of using the hardcoded value.
2) you can just use `early return`, like `return rentalCost - proDiscount`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+  } else if (days >= 3) {
+    sum -= 20;
+  }
```
Viktor-Kost, line 37: 1. just use if, instead of else if, since you have early return
2. consider using a constant for the discount value 30 instead of using the hardcoded value.
3. you can just use early return, like return rentalCost - basicDiscount

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  return sum;
```
Viktor-Kost, line 39: return rentalCost

Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40 * days;
+  const discontMax = 50;
+  const discontMin = 20;
```
volodymyr-soltys97, line 33: Typo in variable name. Change 'discontMin' to 'discountMin'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40 * days;
+  const discontMax = 50;
```
volodymyr-soltys97, line 32: Typo in variable name. Change 'discontMax' to 'discountMax'.

volodymyr-soltys97 (approved):
Good job 👍

PR #2778, opened 2023-05-25
Solution

RodionSav:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costOfDay = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 7 && days >= 3) {
```
vlad-demchuk, line 35: you could change order of checks to avoid && operator

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #2780, opened 2023-05-26
Solution 2

denys2:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40; // Base cost per day is $40
```
polosanya, line 31: What is 40? Try to avoid magic numbers. Also, let's make totalCost a 'const' instead of 'let'

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days >= 7) {
+    totalCost -= 50; // $50 discount for renting 7 or more days
+  } else if (days >= 3) {
+    totalCost -= 20; // $20 discount for renting 3 or more days
+  }
+
+  return totalCost;
```
```suggestion
  if (days >= 7) {
    return totalCost - 50;
  }
  
  if (days >= 3) {
    return totalCost - 20; 
  }

  return totalCost;
```
Just immediately return the full price to avoid using 'else...if'.
Also, create variables for 20 and 50 to make your code more understandable

polosanya (changes requested):
Almost done! There is no need to add comments. Better create variables, so your code is understandable and doesn't need any comments

src/calculateRentalCost.js
denys2: fix

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = days * 40;
+
+  if (days < 3) {
+    return totalCost;
+  } else if (days < 7) {
+    return totalCost - 20;
+  }
```
```suggestion
  if (days < 3) {
    return totalCost;
  } 
  
  if (days < 7) {
    return totalCost - 20;
  }
```

polosanya (changes requested):
Not all comments are fixed. Create variables for 40, 20 and 50

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const carPriceDay = 40;
+  const discountSevenDays = 50;
+  const discountTwoDays = 20;
```
```suggestion
  const bigDiscount = 50;
  const basicDiscount = 20;
```
It is better not to link the variable to values that can be changed later

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return days * carPriceDay;
```
polosanya, line 36: `days * carPriceDay` should be a variable too as it's duplicated three times

polosanya (approved):
Nice! I'm approving it, but please pay attention to the comments

PR #2782, opened 2023-05-27
solution

Platon-jun:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  let totalAmount = days * 40;
+
+  if (days >= 3 && days < 7) {
+    totalAmount -= 20;
+  } else if (days >= 7) {
+    totalAmount -= 50;
+  }
+
+  return totalAmount;
```
Volodymyr-Melnyk-Fron10der, line 39: use early return pattern
```suggestion
  if (days >= 3 && days < 7) {
   return totalAmount - 20;
  } 
  
   if (days >= 7) {
    return totalAmount - 50;
  }

```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalAmount = days * 40;
+
+  if (days >= 3 && days < 7) {
+    totalAmount -= 20;
```
Volodymyr-Melnyk-Fron10der, line 34: move all magic numbers to the constants

Volodymyr-Melnyk-Fron10der (changes requested):

nazarmatsevych (approved):

PR #2783, opened 2023-05-27
Solution

ilona-borkovska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = days * 40;
+
+  if (days < 3) {
+    return totalCost;
+  } else if (days < 7) {
```
mgrinko, line 35: you don't need `else` after return

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+    return totalCost - 20;
+  }
+
+  return totalCost - 50;
```
mgrinko, line 39: It is better to use constants for discounts like `SHORT_TERM_DISCOUNT` and `LONG_TERM_DISCOUNT`

mgrinko (approved):

mgrinko (changes requested):
Please fix my comments

anstsot (approved):
Great job, approving:sparkles:

PR #2786, opened 2023-05-29
Solution

AlisaSkryp:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= 7) {
+    return (days * dayPrice) - basicDiscount;
+  }
+
+  if (days >= 3) {
+    return (days * dayPrice) - minimalDiscount;
+  }
+
+  if (days < 3) {
```
volodymyr-soltys97, line 43: Unnecessary if statement. Since all other conditions are checked, you can use 'else' or simply return the result without any condition.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

volodymyr-soltys97 (approved):
Awesome 🚀

PR #2787, opened 2023-05-29
Vacation-Solution

nenatre:


qoa11a (changes requested):
Good job!
I have a few suggestions:
1. Let's move numbers to the variables with proper names so it's easy to understand what they represent.
2. Instead of changing the result in if statements, it's better to just return the result right away in these if statements. It'd look like this: `if (days >= bigDiscountBoundary) { return days * dayRate - bigDiscount; }`

nazarmatsevych (approved):

PR #2788, opened 2023-05-29
Solution RM

RuslanMelnykFE:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costs = 40;
```
Volodymyr-Melnyk-Fron10der, line 31:
```suggestion
  const pricePerDay = 40;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const totalCosts = costs * days;
```
Volodymyr-Melnyk-Fron10der, line 32:
```suggestion
  const totalPrice = costs * days;
```

Volodymyr-Melnyk-Fron10der (changes requested):
1. Move 20, 50 into a constants

volodymyr-soltys97 (approved):
Awesome 🚀

PR #2792, opened 2023-05-30
rent calculate solution

Tvinki228:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
-  // write code here
+  let counter = 0;
+  const rentDay = 40;
+
+  if (days < 3) {
+    counter = days * rentDay;
+  } else if (days >= 3 && days <= 6) {
+    counter = (days * rentDay) - 20;
+  } else if (days >= 7) {
+    counter = (days * rentDay) - 50;
```
vlad-demchuk, line 39: move all magic numbers(20, 50) to variables

vlad-demchuk: also, could we avoid `else if` and just use if?

vlad-demchuk (changes requested):

nazarmatsevych (approved):
👍

PR #2796, opened 2023-06-01
Solution

maxyma-sanx:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_DISCOUNT = 50;
+  const SHORT_DISCOUNT = 20;
+
+  if (days >= 7) {
+    return days * PRICE_PER_DAY - LONG_DISCOUNT;
+  }
+
+  if (days >= 3 && days < 7) {
```
anstsot, line 39: no need to add additional check for 7 days because you will return the value on line 36
```suggestion
  if (days >= 3) {
```

anstsot (approved):
Good work, approving:sparkles:

PR #2797, opened 2023-06-01
add task solution

KandaskalovOleg:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  return days < SHORT_TERM
+    ? days * PRICE_PER_DAY
+    : days < LONG_TERM
+      ? (days * PRICE_PER_DAY) - SHORT_TERM_DISCOUNT
+      : (days * PRICE_PER_DAY) - LONG_TERM_DISCOUNT;
```
anstsot, line 41: please do not use a lot of ternary operators because it's hard to read. let's rewrite by using `if` and `return` constructions

anstsot (changes requested):
Almost done :+1:
Let's improve a little

YegorKorenkov (approved):
Nice work!

PR #2801, opened 2023-06-03
Solution

hrynkevych:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const RENTAL_PRICE = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  switch (true) {
```
etojeDenys, line 37: Consider using if-else statements instead of switch for better readability and simplicity.

etojeDenys (approved):
interesting solution

src/calculateRentalCost.js
hrynkevych: Got it, thanks!

PR #2802, opened 2023-06-04
task solution

Bonum-taurus:


Volodymyr-Melnyk-Fron10der (changes requested):
Amazing! Just move days * PRICE_PER_DAY into a variable as it's repeating a 3 times

volodymyr-soltys97 (approved):
Nice work 👍

PR #2809, opened 2023-06-05
Solution

IrynaMP:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
```
mykhalenych, line 31: for static constants keep one name conversation or UPPERCASE, or camelCase

mykhalenych (approved):

PR #2812, opened 2023-06-05
Added task solution Transsportation on vacation

ruslan-ihnatenko:


lerastarynets (approved):
Great job, I liked the names you chose for most of variables)

PR #2818, opened 2023-06-06
Solution

Vitalii-Fediaiev:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const longTermDiscount = 50;
+  const middleTermDiscount = 20;
+  const rentPrice = 40;
+
+  if (days >= 7) {
+    return days * rentPrice - longTermDiscount;
```
Anastasiia-Svintsova, line 36: You use `days * rentPrice` value for each case. So create a variable for it

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days >= 3) {
+    return days * rentPrice - middleTermDiscount;
+  } else {
+    return days * rentPrice;
```
Anastasiia-Svintsova, line 40: No need for else here. The function will stop anyway because of the return statement before.
```suggestion
  }
    return days * rentPrice;
```

Anastasiia-Svintsova (changes requested):

polosanya (changes requested):
One small but important fix left

sTorba24 (approved):
Well done!

PR #2819, opened 2023-06-07
add task solution

irynabolbat:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  let totalPrice = PRICE_PER_DAY * days;
+
+  if (days >= LONG_TERM) {
+    totalPrice -= LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
+    totalPrice -= SHORT_TERM_DISCOUNT;
+  }
```
meforoma, line 43: let's make totalPrice a `const`
and use `if..return` construct to avoid `else if`

meforoma (changes requested):
please update according to the comment

meforoma (approved):
👍🏼

PR #2823, opened 2023-06-07
Solution

MarianaStorchyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const totalAmount = pricePerDay * days;
+  const longTerm = 7;
+  const shortTerm = 3;
+  const longTermDiscount = 50;
+  const shortTermDiscount = 20;
+
+  if (days >= longTerm) {
+    return totalAmount - longTermDiscount;
+  };
+
```
alinatkachukkkk, line 41: Moved the variable declarations (pricePerDay, totalAmount, longTerm, shortTerm, longTermDiscount, shortTermDiscount) inside the function scope, as they are only used within the function.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const pricePerDay = 40;
+  const totalAmount = pricePerDay * days;
+  const longTerm = 7;
+  const shortTerm = 3;
+  const longTermDiscount = 50;
+  const shortTermDiscount = 20;
+
+  if (days >= longTerm) {
+    return totalAmount - longTermDiscount;
+  };
```
alinatkachukkkk, line 40: Changed the second if statement to an else if statement. Since both conditions (days >= longTerm and days >= shortTerm) can't be true at the same time, it is more efficient to use else if to avoid unnecessary checks.

ghost (approved):
GJ!

PR #2835, opened 2023-06-08
Solution task-transportation-on-vacation

NataliGru:
- for `totalCost` variable:  `const` instead of `let` was added automatically by lint


meforoma (approved):
👍🏼

PR #2836, opened 2023-06-08
Finished task

IgorOksentyuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const TOTAL__PRICE = days * 40;
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    return TOTAL__PRICE - SHORT_TERM_DISCOUNT;
+  }
+
+  if (days >= LONG_TERM) {
+    return TOTAL__PRICE - LONG_TERM_DISCOUNT;
+  }
```
meforoma, line 43: you would have less code if evaluated the longTerm first
```suggestion
  if (days >= LONG_TERM) {
    return TOTAL__PRICE - LONG_TERM_DISCOUNT;
  }

  if (days >= SHORT_TERM) {
    return TOTAL__PRICE - SHORT_TERM_DISCOUNT;
  }

```

meforoma (approved):
one non-blocking optimisation comment

PR #2854, opened 2023-06-09
add task solution

NadiaKovalchuk:


zoriana-melnychuk (approved):

perfectAcedia (approved):
Looks perfect for me :)
However, I prefer to use CamelCase over the Snake_Case variables:
longTerm instead of LONG_TERM for example.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const PRICE_PER_DAY = 40;
+
+  if (days >= LONG_TERM) {
+    return days * PRICE_PER_DAY - LONG_TERM_DISCOUNT;
```
juljaVysotska, line 38: You use `days * PRICE_PER_DAY` two times minimum(here and 42 line), plus this is total rent value for all days. We don't need to recalculate it every time.

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them.

Best regards 🌸

mgrinko (approved):
Please keep @juljaVysotska's comment in mind for your future solutions

PR #2855, opened 2023-06-09
Solution

olegpaliy:


maksymPtashnyk (changes requested):
Good job !)
It works, but you should follow the checklist 
[Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

maksymPtashnyk (approved):
Great!

FlameInBrains (changes requested):
It is better to not use 'else' after if with 'return'. you can use two 'if' blocks instead;

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  if (days >= LONG_TERM) {
+    return PRICE_PER_DAY * days - LONG_TERM_DISCOUNT;
```
tetiana-ivanchenko, line 38: move PRICE_PER_DAY * days in the variable

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2860, opened 2023-06-09
Solution

zoriana-melnychuk:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const oneDayRentCost = 40;
```
dipinchuk, line 32: Literals should be written in capital letters. (Lesson 'Transportation on Vacation Solution' , 2:17)
```suggestion
  const COST_PER_DAY = 40;
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const longTerm = 7;
+  const shortTerm = 3;
+
+  switch (true) {
```
dipinchuk, line 38: If the condition is not an iterative of many values, it is better to use `if`. (Lesson 'Switch True', 2:10)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+    case days >= longTerm:
+      return days * oneDayRentCost - bigDiscount;
```
dipinchuk, line 40: In our case, the price calculation logic is the same, so it is better to create a variable. (Lesson 'Transportation on Vacation Solution' , 5:10) 
```suggestion
      const basePrice = days * oneDayRentCost;
      return basePrice - bigDiscount;
```

dipinchuk (changes requested):

dipinchuk (approved):
Great job!

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
```
tetiana-ivanchenko, line 31: don't left comments)

tetiana-ivanchenko (approved):

PR #2861, opened 2023-06-09
transportation_on_vacation

FlameInBrains:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const longTermDiscount = 50;
```
anya-shults, line 32: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

please rename your constants like in the GOOD EXAMPLE

anya-shults (changes requested):
Good job!) let's make your code prettier)

anya-shults (approved):
👍

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+  const shortTermDiscount = 20;
+  const pricePerDay = 40;
+  const LongTerm = 7;
+  const ShortTerm = 3;
+
+  if (days >= ShortTerm) {
+    if (days >= LongTerm) {
+      return (days * pricePerDay) - longTermDiscount;
```
tetiana-ivanchenko, line 40: 1. don't use nested condition
2. also move days * pricePerDay in the variable

tetiana-ivanchenko (changes requested):

mgrinko (approved):
Please keep this info in mind for our future code

PR #2864, opened 2023-06-09
Solution

outofs:


klmyana:
Good work and clear code! I had the same desigion, but mentor adviced me do not repear (RENT_VALUE * days) 3 times and better make it variable and reuse)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const RENT_VALUE = 40;
+
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  if (days >= LONG_TERM) {
+    return RENT_VALUE * days - LONG_TERM_DISCOUNT;
```
klmyana, line 40: Good work and clear code! I had the same desigion, but mentor adviced me do not repear (RENT_VALUE * days) 3 times and better make it variable and reuse)

klmyana, line 40: Good work and clear code! I had the same desigion, but mentor adviced me do not repear (RENT_VALUE * days) 3 times and better make it variable and reuse)

outofs: Ok, tnx. Next time I'll take your advice)

juljaVysotska, line 40: You use `RENT_VALUE * days` two times minimum(here and 44 line), plus this is total rent value for all days. We don't need to recalculate it every time.

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them.

Best regards 🌸

tetiana-ivanchenko (approved):

PR #2867, opened 2023-06-10
Add task solution

kbekher:


demetrxx (changes requested):
Wrong usage of the switch case. Learn proper usage [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch)

VitaliyBondarenko1982 (approved):

PR #2869, opened 2023-06-10
Solution

OlhaSheliakina:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRentCost = 40;
```
taras-bilas, line 31: Good job!
[Checklist] It had better change the variables' names on upper case letters

image

taras-bilas (changes requested):

src/calculateRentalCost.js
taras-bilas: According to Misha's video on the platform, we should use upper-case letters for constants :)

taras-bilas (approved):

src/calculateRentalCost.js
anstsot: agreed with [comment](https://github.com/mate-academy/js_task-transportation-on-vacation/pull/2869/files#r1226322815)
better to use upper-case letters for constants. it's like a rule for the devs

anstsot (approved):
Good work, approving:sparkles:
But pay attention to the comments to avoid the same mistakes in the future

mgrinko (approved):
In actual code in such case, I'd use LONG_TERM_DISCOUNT as this value is a parameter of your logic, the value that is not calculated but defined from the beginning. 

So it is more common practice to use CAPITAL letters for such names
But it is not a mistake just a tradition

PR #2872, opened 2023-06-11
add solution

AnatoliiLykhohodenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const shortTerm = 3;
+  const shortTermDiscount = 20;
+
+  const longTerm = 7;
+  const longTermDiscount = 50;
```
kylyvnyk, line 36: Great job:)

shortTerm = SHORT_TERM

Just a minor comment. As a tradition, all constant values should be assigned in uppercase with 
underscore (like in the example above)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+
+  let res = 0;
+
+  for (let i = 0; i < days; i++) {
+    res += 40;
+  }
```
kylyvnyk: You don't need this loop. It is enough to multiply the number of days by the constant cost per day.
res = days * 40;

Additionally, you may use this variable 'res' at the beginning and after the final 'return'. 

Also, the name 'res' is not very representative. Something like 'totalCost' or 'totalRent' would be better.

kylyvnyk (changes requested):

kylyvnyk (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+
+  let totalPrice = 0;
+
+  for (let i = 0; i < days; i++) {
+    totalPrice += PRICE_PER_DAY;
+  }
```
tetiana-ivanchenko, line 42: try to avoid loop

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2881, opened 2023-06-14
work done

IntensoSergey:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DEY = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
```
Anastasiia-Svintsova, line 35: It's better to save these constants outside the function

Anastasiia-Svintsova (approved):

PR #2883, opened 2023-06-14
Calculate rental cost solution with succes tests

ValeraViachalo:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+    case days < MIN_TERM:
+      return basePrice;
+
+    case
+      days >= MIN_TERM
+      && days <= LONG_TERM - 1:
+      return basePrice - MIN_TERM_DISCOUNT;
+
+    case days >= LONG_TERM:
+      return basePrice - LONG_TERM_DISCOUNT;
```
mykhalenych, line 51: don't forget about default case

mykhalenych (approved):

PR #2884, opened 2023-06-14
calculate price for rent car

ivan-baranovskyi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const pricePerDay = 40;
+  const discountFifty = 50;
+  const discountTwenty = 20;
+
+  if (days >= 7) {
+    return days * pricePerDay - discountFifty;
+  } else if (days >= 3) {
+    return days * pricePerDay - discountTwenty;
+  }
```
```suggestion
  if (days >= 7) {
    return days * pricePerDay - discountFifty;
  } 
  
  if (days >= 3) {
    return days * pricePerDay - discountTwenty;
  }
```
it's better not use `else` if it's possible (when you `return` in previous `if` block) - it looks clearer

TarasHoliuk (approved):
GJ :+1: 
Added one comment just FYI

PR #2890, opened 2023-06-16
Solution

aRabiej:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
```
Radoslaw-Czerniawski, line 30: Here some examples how this could be refactored:

```js
  const discount = days >= 7 ? 50 : days >= 3 ? 20 : 0;

  return days * 40 - discount;
}
```

or

```js
  let discount = 0;

  if (days >= 7) {
    discount = 50
  } else if (days >= 3) {
    discount = 20
  };

  return days * 40 - discount;
}
```

or as an arrow function

```js
const calculateRentalCost = (days) => days * 40 - (days >= 7 ? 50 : days >= 3 ? 20 : 0)
```

Radoslaw-Czerniawski (approved):
Good job, left a comment with some refactor ideas :)

PR #2891, opened 2023-06-16
Develop

Luk2asz:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  let result = days * 40;
+  const longTerm = 7;
+  const shortTerm = 3;
+  const longTermDiscount = 50;
+  const shortTermDiscount = 20;
+
+  if (days >= longTerm) {
+    result -= longTermDiscount;
+
+    return result;
```
DorotaLeniecDev, line 40: You could also write this for simplicity:
```suggestion
    return result - longTermDiscount;
```

DorotaLeniecDev (approved):
Good job, one suggestion from my side

PR #2892, opened 2023-06-17
Solution

AlexKinda:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
Anastasiia-Svintsova, line 36: It's better to save these variables outside the function

Anastasiia-Svintsova (approved):

PR #2898, opened 2023-06-18
FirstTaskDone

Kyzenko:
First task Js Advansed

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -30,22 +30,23 @@
 function calculateRentalCost(days) {
   // write code here
 
-  const cost = 40;
-  let finalPrice = '';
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
 
 function calculateRentalCost(days) {
   // write code here
+
+  const PRICE_PER_DAY = 40;
```
SerhiiKirik, line 33: Just give a name with lowerCase & camelCase
use upperCase when you need a variable in a lot of cases
```suggestion
  const pricePerDay = 40;
```

SerhiiKirik (approved):
Well done 🔥 
Take a look at the comments 🧑‍💻

PR #2909, opened 2023-06-22
Develop

Patryk-Filipiak:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,16 @@
...
  */
 
-function calculateRentalCost(days) {
-  // write code here
+function calculateRentalCost(numberOfDays) {
+  const DAY_COST = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const MIDDLE_TERM = 3;
+  const MIDDLE_TERM_DISCOUNT = 20;
```
choeqq, line 35: those should be outside of function

src/calculateRentalCost.js
```diff
@@ -27,8 +27,16 @@
...
+  const total = DAY_COST * numberOfDays;
+
+  return numberOfDays >= LONG_TERM ? total - LONG_TERM_DISCOUNT
+    : numberOfDays >= MIDDLE_TERM ? total - MIDDLE_TERM_DISCOUNT : total;
```
choeqq, line 39: its smart but I think it would be way easier to read if you'd use `if else` statement

choeqq (changes requested):

src/calculateRentalCost.js
Erchu420: done in 'Fixed soultion' ; )

src/calculateRentalCost.js
Erchu420: done in 'Fixed soultion' ; )

src/calculateRentalCost.js
JoyCoffeeAddict: You could rewrite it to avoid `else if` according to checklist point 4. Then it will be perfect

JoyCoffeeAddict (changes requested):
Rewrite according to checklist point 4

choeqq (changes requested):
Please try to make it work without `else if` and you're good to go 🚀

Erchu420:
> Please try to make it work without `else if` and you're good to go 🚀

done :)

choeqq (approved):
Nicely done 🚀

PR #2913, opened 2023-06-24
Solution

Anmonnn:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_COST = 50;
+
+  const total = days * COST;
+
+  if (days >= LONG_TERM) {
+    return total - LONG_TERM_COST;
+  } else if (days >= SHORT_TERM) {
+    return total - SHORT_TERM_COST;
+  }
```
VitaliyBondarenko1982, line 44:
```suggestion
  }
  
   if (days >= SHORT_TERM) {
    return total - SHORT_TERM_COST;
  }
```
do not use else whert it is not necessary

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #2920, opened 2023-06-27
add solution task

pti4ka121:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_ONE_DAY = 40;
+  const LONG_TERM = 7;
```
SerhiiKirik, line 32: Just in lowerCase or camelCase
Do it for constant wich you reuse in a lot of files
```suggestion
  const priceOneDay = 40;
  const long_term = 7;
```

SerhiiKirik (approved):
Nice 🔥

PR #2921, opened 2023-06-27
add solution

NikitaOrliak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
SerhiiKirik, line 33: just give the name with camelCase
use name in Upper case for the constants, which you use in a lot of files or when it's a global constants

SerhiiKirik (changes requested):
Well done 🔥 
Let's rename your variables 🧑‍💻

demetrxx (approved):
Well done 👍 Go ahead!

PR #2923, opened 2023-06-29
task solution

AnVBondar:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days >= MAX_DAYS_DISCOUNT) {
+    return PRICE - BIG_DISCOUNT;
+  }
+
+  if (days >= MIN_DAYS_DISCOUNT) {
+    return PRICE - SMALL_DISCOUNT;
+  }
+
+  return PRICE;
```
IvanFesenko, line 47: With CAPS should be named only constants that are static
```suggestion
  const price = days * AMOUNT_PER_DAY;

  if (days >= MAX_DAYS_DISCOUNT) {
    return price - BIG_DISCOUNT;
  }

  if (days >= MIN_DAYS_DISCOUNT) {
    return price - SMALL_DISCOUNT;
  }

  return price;
```

IvanFesenko (changes requested):
good job. Fix small isuue with naming

anastasiiavorobiova (approved):
Awesome!

PR #2927, opened 2023-06-30
done

volodymyr-tytarenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  let totalCost = costPerDay * days;
+
+  if (days >= 7) {
+    totalCost -= 50;
+  } else if (days >= 3) {
+    totalCost -= 20;
```
Anastasiia-Svintsova, line 37: Don't use mystery numbers. Would be great to create variables for 50 and 30 as well

Anastasiia-Svintsova (approved):

PR #2932, opened 2023-07-02
Solution_calculateRentalCost

Marinakyrychynska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHOT_TERM = 3;
```
volodymyr-soltys97, line 34: There is a typo in the constant declaration, it should be 'SHORT_TERM' instead of 'SHOT_TERM'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const SHOT_TERM_DISCOUNT = 20;
```
volodymyr-soltys97, line 35: There is a typo in the constant declaration, it should be 'SHORT_TERM_DISCOUNT' instead of 'SHOT_TERM_DISCOUNT'.

volodymyr-soltys97 (approved):
Well done 🔥

PR #2943, opened 2023-07-07
Solution

spojrzenie:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
```
DorotaLeniecDev, line 31: _Don't use literals in expressions (HARDCODE). Instead use constants explaining this values_ - according to the checklist, please fix this

DorotaLeniecDev (changes requested):

darokrk (approved):
Nicely done 💪🏽  approved 👏🏽

PR #2944, opened 2023-07-08
Solution

Illia-L:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  let discount = 0;
+
+  if (days >= SHORT_TERM) {
+    discount = SHORT_TERM_DISCOUNT;
+  };
+
+  if (days >= LONG_TERM) {
+    discount = LONG_TERM_DISCOUNT;
+  };
```
volodymyr-soltys97, line 44: The if statements should not have a semicolon at the end.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2945, opened 2023-07-08
Solution of task calculateRentalCost

thaiduk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let rentalCost = 0;
+
+  if (days < 3) {
+    rentalCost = days * 40;
```
BudnikOleksii, line 34: Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  if (days >= 3 && days < 7) {
+    rentalCost = (days * 40) - 20;
+
+    return rentalCost;
+  }
+
+  if (days >= 7) {
+    rentalCost = (days * 40) - 50;
+
+    return rentalCost;
```
BudnikOleksii, line 48: If you return "rentalCost" in every statement you don't need this variable, you can just return the result of the calculations.

BudnikOleksii (changes requested):
Nice start👍
Left some comments, check them.

PR #2956, opened 2023-07-12
Transportation on vacation

Kacper-Leman:
Soultion

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
```
DorotaLeniecDev, line 30: D_on't use literals in expressions (HARDCODE). Instead use constants explaining this values_ - cheklist point, create consts for discount values

DorotaLeniecDev (changes requested):
One change needed

DorotaLeniecDev (approved):
Good job 🆗

PR #2960, opened 2023-07-12
Solution

VasylKurtiak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const rentalCost = 40;
+  const total = rentalCost * days;
+
+  if (days >= 7) {
+    return total - greatDiscount;
+  } else if (days >= 3) {
+    return total - basicDiscount;
+  }
+
+  return total;
```
Viktor-Kost, line 42: Instead of using 'else if' on line 12, consider returning from the function immediately after each condition. This reduces the level of indentation and makes the code cleaner and easier to read.
```suggestion
  if (days >= 7) {
    return total - greatDiscount;
  } 
  
  if (days >= 3) {
    return total - basicDiscount;
  }

  return total;
```

Viktor-Kost (changes requested):

etojeDenys (approved):
lgtm

PR #2963, opened 2023-07-13
Solution

SichVictor:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceArendCar = 40;
+
+  if (days >= 7) {
+    return (days * priceArendCar) - 50;
```
OleziO, line 34: Move all discount and days anount to the constant variables.

OleziO (changes requested):

mykhalenych (approved):

PR #2967, opened 2023-07-14
solution added

FurArt:
Hi, 
completed the task and compared it with the check list

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const shortTerm = 3;
+  const longTerm = 7;
+  const basePrice = costPerDay * days;
+  let discount = 0;
+
+  if (days >= longTerm) {
+    discount = 50;
+
+    return basePrice - discount;
+  }
+
+  if (days >= shortTerm) {
+    discount = 20;
+
+    return basePrice - discount;
+  }
+
+  return basePrice;
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -33,17 +33,17 @@ function calculateRentalCost(days) {
   const SHORT_TERM_DISCOUNT = 20;
   const LONG_TERM = 7;
   const LONG_TERM_DISCOUNT = 50;
-  const BASE_PRICE = COST_PER_DAY * days;
+  const basePrice = COST_PER_DAY * days;
 
   if (days >= LONG_TERM) {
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
```
anastasiiavorobiova, line 31:
```suggestion
  const COST_PER_DAY = 40;
```
follow constants naming convention: 
```
const HOURS_IN_DAY = 24;
const hoursRemaining = currentHour - HOURS_IN_DAY;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  const shortTerm = 3;
+  const longTerm = 7;
+  const basePrice = costPerDay * days;
```
anastasiiavorobiova, line 34: See my previous comment

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+  let discount = 0;
+
+  if (days >= longTerm) {
+    discount = 50;
```
anastasiiavorobiova, line 38: avoid using magic numbers it's the programming anti-pattern

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+    return basePrice - discount;
+  }
+
+  if (days >= shortTerm) {
+    discount = 20;
```
anastasiiavorobiova, line 44: See my previous comment

anastasiiavorobiova (changes requested):
Great start! Keep doing

src/calculateRentalCost.js
FurArt: What differrent with upper case use with camelcase?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const COST_PER_DAY = 40;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const BASE_PRICE = COST_PER_DAY * days;
```
anastasiiavorobiova, line 36:
```suggestion
  const basePrice = COST_PER_DAY * days;
```
The difference: its value is being counted.

anastasiiavorobiova (changes requested):
Almost done! Added a tiny comment to fix

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2976, opened 2023-07-17
solution

Annbpiu:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  let payRent = PRICE_PER_DAY * days;
```
IvanFesenko, line 37: no need to use `let` - value never changed - so it should be `const`

IvanFesenko (approved):
Well done

src/calculateRentalCost.js
Annbpiu: thank you!

PR #2977, opened 2023-07-17
solution

Sancz0pansa:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
```
Matylda-Lewandowska, line 31: You should use variables in places like this. :)
2nd point in [checklist.](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

Something like that
```
const LONG_TERM = 7;

}
```

Not : 
```
if (numberOfDays >= 7) {
  return basePrice - 50;
}
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
...
+    return days * 40 - 50;
+  } else if (days >= 3) {
```
Matylda-Lewandowska, line 33: As in the 5th point of the checklist, it is written, you should change it, so it would look like this:
```
if (condition) {
  return x;
}

return y;
```

and not : 
```
if (condition) {
  return x;
} else {
  return y;
}
```

Matylda-Lewandowska (changes requested):
Almost done! Some changes need to be done, in my comments you will find points, from checklist, what is needed to be changed. :)

If you need any help with it, write. :)

choeqq (changes requested):
Please correct your solution according to comments above 🚀

src/calculateRentalCost.js
```diff
@@ -26,9 +26,21 @@
  *
  * @return {number}
  */
+const longTerm = 7;
+const longTermDiscount = 50;
+const shortTerm = 3;
+const shortTermDiscount = 20;
```
choeqq, line 32: those should be uppercased - [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) point 2

src/calculateRentalCost.js
```diff
@@ -26,9 +26,21 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const basePrice = days * 40;
+
+  if (days >= longTerm) {
+    return basePrice - longTermDiscount;
+  } else if (days >= shortTerm) {
```
choeqq, line 39: try to make it worki without `else if` - [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) point 4, 5

choeqq (changes requested):

src/calculateRentalCost.js
```diff
@@ -26,9 +26,23 @@
...
  * @return {number}
  */
+const LONG_TERM = 7;
+const LONG_TERM_DISCOUNT = 50;
+const SHORT_TERM = 3;
+const SHORT_TERM_DISCOUNT = 20;
 
 function calculateRentalCost(days) {
-  // write code here
+  const BASE_PRICE = days * 40;
```
Matylda-Lewandowska, line 35: This "40" better to move out like SHORT_TERM and all other valuables you did before. :)
"BASE_PRICE", because it is within the function, the name should be in lowercase, but I'm not sure.

Matylda-Lewandowska (changes requested):
Only little changes I think should be made. :) 
You are really close! 🚀

src/calculateRentalCost.js
choeqq: that's right

choeqq, line 35: so `BASE_PRICE` depends on `days` which is function argument, so it will change for sure - we typically use uppercased variable names for things that aren't changing

choeqq (changes requested):

src/calculateRentalCost.js
JoyCoffeeAddict: Very good student review!

JoyCoffeeAddict (approved):

mvjl000 (approved):

Matylda-Lewandowska (approved):
Great job! :D

PR #2979, opened 2023-07-17
Solution

irahos:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+
+  if (days < SHORT_TERM) {
+    return rentalCosts;
+  }
+
+  if (days < LONG_TERM && days >= SHORT_TERM) {
+    return rentalCosts - SHORT_TERM_DISCOUNT;
+  }
+
+  if (days >= LONG_TERM) {
```
FrontKid, line 47: You don't need to add the last condition, as the return in the previous "if" block will terminate the function execution if the condition is true. We don't want to do that becuz if the "day" is incorrect value or it didn't pass the 3 previous conditions - our function will return "undefined" 

[link](https://highload.today/undefined-v-javascript/#3)

so u can just remove it and leave return 
```return rentalCosts - LONG_TERM_DISCOUNT;```

**Review by FrontKid — CHANGES_REQUESTED** [review]
Good job but needs some tweaking

**Review by FrontKid — APPROVED** [review]

File: src/calculateRentalCost.js [hunk 2]
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const SHORT_TERM = 3;
```
tiserett, line 32: SHORT_TERM_DAYS

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const LONG_TERM = 7;
```
tiserett, line 33: LONG_TERM_DAYS

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  const rentalCosts = days * PRICE_PER_DAY;
```
tiserett, line 37: it's not a rentalCost, it's a basePrice

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days < SHORT_TERM) {
+    return rentalCosts;
+  }
+
+  if (days < LONG_TERM && days >= SHORT_TERM) {
+    return rentalCosts - SHORT_TERM_DISCOUNT;
+  }
+
+  return rentalCosts - LONG_TERM_DISCOUNT;
```
tiserett, line 47: think how you can change it to get rid of '&&'

tiserett (changes requested):
good work, but let's fix those minor issues

Anna-Tkachenko (approved):
⭐️

PR #2980, opened 2023-07-17
solution

Patryk-Buczkowski:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let costs = days * 40;
```
DorotaLeniecDev, line 31: From the cheklist: _Don't use literals in expressions (HARDCODE). Instead use constants explaining this values_
Create proper const variables

DorotaLeniecDev (changes requested):
Create proper const variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const moneyPerDay = 40;
```
DorotaLeniecDev, line 31: Create proper const variables as the checklist suggests:
```js
const LONG_TERM = 7;

}
```

DorotaLeniecDev (changes requested):
Changes needed

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
```
DorotaLeniecDev, line 31: It can also be named with uppercase as other consts
```suggestion
  const COST_PER_DAY = 40;
```

DorotaLeniecDev (approved):
Good job, approving, one comment from my side

PR #2981, opened 2023-07-17
task is done

vladkorobka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const TOTAL_PRICE = days * 40;
```
Viktor-Kost, line 31: usually we use uppercased naming for value that aren't gonna be changed. But here you have a dynamic value
```suggestion
  const total = days * 40;
```

Viktor-Kost (approved):
GJ!
Just look at recommendation I left for you

PR #2986, opened 2023-07-19
Solution

HubertKlamann:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const longTerm = 7;
+  const shortTerm = 3;
+  const longTermDiscount = 50;
+  const shortTermDiscount = 20;
```
mvjl000, line 34: Move these consts outside the function and capitalize them (checklist point 2)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const totalCost = days * 40;
```
mvjl000, line 35: I think this `40` here should also be a constant

mvjl000 (changes requested):
You just need some small changes but in general - good job! 🚀

mvjl000 (approved):

PR #2999, opened 2023-07-21
solved task

olegobiukh:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const BASE_PRICE = PRICE_PER_DAY * days;
+
+  if (days >= LONG_TERM) {
+    return BASE_PRICE - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
+    return BASE_PRICE - SHORT_TERM_DISCOUNT;
+  }
```
SanyaBratashchuk, line 42:
```suggestion
  } 
   
   if (days >= SHORT_TERM) {
    return BASE_PRICE - SHORT_TERM_DISCOUNT;
  }
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const BASE_PRICE = PRICE_PER_DAY * days;
```
SanyaBratashchuk, line 36:
```suggestion
  const basePrice = PRICE_PER_DAY * days;
```

SanyaBratashchuk (changes requested):

src/calculateRentalCost.js
olegobiukh: but why make them separate? If elseif statement makes function run faster

etojeDenys: how it will make our function work faster if we always check the condition that is in the if
image

etojeDenys (approved):
lgtm

PR #3004, opened 2023-07-23
solution

MaciejKujawa:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const numberOfDays = days;
```
choeqq, line 31: there is no need to reassign days to numberOfDays

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+  const DAY_COST = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
```
choeqq, line 36: those should be outside of function

choeqq (changes requested):

choeqq (approved):
Nicely done 🚀

PR #3012, opened 2023-07-24
Task Solution - transportation-on-vacation

Matylda-Lewandowska:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
  * @return {number}
  */
 
+const rentCarCost = 40;
+const shortTerm = 3;
+const longTerm = 7;
+const discountShortTerm = 20;
+const discountLongTerm = 50;
```
mvjl000, line 34: Capitalize the names of these consts - you can see an example in the checklist

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
...
+
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * rentCarCost;
+
+  if (days >= longTerm) {
+    return (result - discountLongTerm);
```
mvjl000, line 40: You don't need parentheses here, same thing with the second return ;)
```suggestion
    return result - discountLongTerm;
```

mvjl000 (changes requested):

src/calculateRentalCost.js
mvjl000: Oh and 'result' is not the best name, try to come up with something different ;)

Matylda-Lewandowska: Ah! Alright, I get it. I wasn't sure, if I can live it without. :)

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
...
 
+const RENT_CAR_COST = 40;
+const SHORT_TERM = 3;
+const LONG_TERM = 7;
+const DISCOUNT_SHORT_TERM = 20;
+const DISCOUNT_LONG_TERM = 50;
+
 function calculateRentalCost(days) {
-  // write code here
+  const RESULT = days * RENT_CAR_COST;
```
JoyCoffeeAddict, line 37: Actually, `RESULT` does not have to be in capital letters. Think of it in a way that constants like above: `SHORT_TERM` `LONG_TERM` should never change in your code, like `Math.PI` or `Math.SQRT2`. But a variable with keyword `const` does not always have to be written in capital letters, this is a regular const that will change based on `days`, so it should be `const result`

JoyCoffeeAddict (changes requested):

JoyCoffeeAddict (approved):

PR #3015, opened 2023-07-24
Solution

emil-owczarek:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
```
asakevych, line 31: it's a good practice to use capital letters for hardcoded values -> `COST_PER_DAY`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const baseRentalCost = days * costPerDay;
```
asakevych, line 32: it's better to create this const after all hardcoded values

asakevych (changes requested):
Good job! small changes are needed 😉

asakevych (approved):
Great job! 🚀 🔥 🔥

PR #3023, opened 2023-07-25
add task solution

griveous199:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days < 7) {
+    result = 40 * days - 20;
+
+    return result;
+  } else {
+    result = 40 * days - 50;
+
+    return result;
+  }
```
choeqq, line 47: please use variables for constant numbers - [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) point 2

choeqq, line 47: try to make it work without else - checklist point 4

choeqq (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+
+    return finalCost;
+  }
+
+  if (days < longTerm) {
+    finalCost = cost * days - smallDiscount;
+
+    return finalCost;
+  }
+  finalCost = cost * days - bigDiscount;
```
choeqq, line 49: you can also return that without assignment

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  const shortTerm = 3;
+  const longTerm = 7;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
```
choeqq, line 35: those should be uppercased and outside of function

choeqq (changes requested):

choeqq (approved):
Nicely done ✅

PR #3025, opened 2023-07-25
solution

Zhekich123:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  if (days >= 7) {
+    return cost * days - 50;
+  } else if (days >= 3) {
+    return cost * days - 20;
+  } else {
+    return cost * days;
+  }
 }
 
+console.log(calculateRentalCost(7))
```
Viktor-Kost, line 42: remove console.log

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const cost = 40;
+  
+  if (days >= 7) {
+    return cost * days - 50;
+  } else if (days >= 3) {
+    return cost * days - 20;
+  } else {
+    return cost * days;
+  }
```
Viktor-Kost, line 39: 1. move such values as 3, 7, 20 and 50 to variables
2. it's better to avoid using `else if` for better readability. You can use `early return`
Example:
``` 
if(condition1) {
  return result1
}

if(condition2) {
  return result2
}

return resultDefault
```

Viktor-Kost (changes requested):

src/calculateRentalCost.js
Zhekich123: already removed

src/calculateRentalCost.js
Zhekich123: Thanks for review. I making changes!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
+    return result1;
+  } else if (days >= threeDays) {
+    const result2 = cost * days - twentyOff;
+
+    return result2;
+  } else {
+    const resultDefault = cost * days;
+    
+    return resultDefault;
+  }
```
Viktor-Kost, line 49:
```suggestion
  if (days >= daysForBasicDiscount ) {
    return rentalCost - proDiscount
  } 
  
  if (days >= daysForProDiscount) {
    return rentalCost - proDiscount;
  } 
   
  return rentalCost;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  const threeDays = 3;
+  const sevenDays = 7;
+  const twentyOff = 20;
+  const fifteenOff = 50;
```
Viktor-Kost, line 35: it's better to add names that can be appropriate for different values
```suggestion
  const rentalCost = 40 * days
  const daysForBasicDiscount = 3;
  const daysForProDiscount = 7;
  const basicDiscount = 20;
  const proDiscount = 50;
```

Viktor-Kost (changes requested):

Viktor-Kost (approved):

IvanFesenko (approved):
Well done

PR #3049, opened 2023-07-28
Solution

Orpheus29:


polinavafik (approved):
great job!!

Orpheus29:
> great job!!

Thank you!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const NUMBER_OF_DAYS = days;
```
roman-mirzoian, line 31: `days` already comes as a variable from the function arguments, so there is no need to create a new variable for it.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  const DAILY_COST = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const BASE_PRICE = NUMBER_OF_DAYS * DAILY_COST;
```
roman-mirzoian, line 38: It is customary to name only static constants with capital letters, in which there will be no calculations, so here it would be worthwhile to name the constant using camelCase (`basePrice`).

roman-mirzoian (approved):

src/calculateRentalCost.js
Orpheus29: Done.

Anna-Tkachenko (approved):

ihor-jpeg (approved):

PR #3060, opened 2023-07-30
create vacation func

rostyslav48:


ihorid (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const RENTAL_PRICE = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  if (days >= LONG_TERM) {
+    return (RENTAL_PRICE * days) - LONG_TERM_DISCOUNT;
```
Anna-Tkachenko, line 38: Don't repeat yourself: will be better to create separate variable for this calculation as you repeated it 3 times

Anna-Tkachenko (changes requested):

Moroz-Dmytro (approved):
LGTM 👍

PR #3067, opened 2023-07-30
finish solution

Duster-1:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const discountTreeDays = 20;
+  const discountSevenDays = 50;
+  const longPeriod = 7;
+  const shortPeriod = 3;
+
+  if (days < shortPeriod) {
+    return days * rentalPrice;
+  }
+
+  if (days >= shortPeriod && days < longPeriod) {
```
volodymyr-soltys97, line 41: There is no need to check 'days >= shortPeriod' because it is always true in this context due to the previous if statement. You can remove it.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalPrice = 40;
+  const discountTreeDays = 20;
```
volodymyr-soltys97, line 32: Typo in variable name. It should be 'discountThreeDays' instead of 'discountTreeDays'.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #3068, opened 2023-07-30
Add task solution

davinnchii:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,17 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let result = 0;
+
+  if (days >= 3 && days < 7) {
+    result = days * 40 - 20;
```
AnastasiiaHombalevska, line 35: 2. [checklist] - Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

AnastasiiaHombalevska (changes requested):
God job!
Please fix your code

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const basePrice = 40;
+
+  const shortTerm = 3;
```
roman-mirzoian, line 34: Static constants (that is, those in which no calculations are made) are usually named with capital letters.
For example `SHORT_TERM`, `SHORT_TERM_DISCOUNT` etc.

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done, but do not forget to mark the column with the checklist if you checked your code with it before submitting.

PR #3069, opened 2023-07-30
Solution 1

voronine:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const mult = days * 40;
```
voievodik, line 31: Good job! 

let's add some constant variables:
**[checklist] 2. Don't use literals in expressions (HARDCODE). Instead use constants explaining this values**

src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
...
+  let result = mult;
+
+  if (days >= 7) {
+    result = mult - 50;
+  } else if (days >= 3) {
```
voievodik, line 36: change the if else construction:
**[checklist] 4. Prefer if with return over if else to simplify later conditions.**

voievodik (changes requested):

src/calculateRentalCost.js
Moroz-Dmytro: Good point

src/calculateRentalCost.js
Moroz-Dmytro: Good point

Moroz-Dmytro (changes requested):
Please, check comments

BudnikOleksii (approved):
Good job👍

PR #3071, opened 2023-07-31
add task solution

bgutai:


rostyslav48 (approved):
Good job! Code is tidy and neat)👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const SHORT_DAYS = 3;
+  const LONG_DAYS = 7;
+  const SHORT_DAYS_DISCOUNT = 20;
+  const LONG_DAYS_DISCOUNT = 50;
+  const RENTAL_PRICE = 40;
+  const TOTAL_PRICE = days * RENTAL_PRICE;
```
Moroz-Dmytro, line 36: It is customary to name static variables with a capital letter, if something is calculated in the variable, it is better to name it using the usual camelCase.

Moroz-Dmytro (changes requested):
Small comment
Also if you went through checklist set checkmark in the table
image

Anna-Tkachenko (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const SHORT_DAYS = 3;
```
Anna-Tkachenko, line 31: Will be better to name it as `term` instead of `days`

PR #3072, opened 2023-07-31
add task solution

SmartSergiZ:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const COST_PER_DAY = 40;
+  const DISCOUNT_50 = 50;
+  const DISCOUNT_20 = 20;
+
+  if (days >= 7) {
```
Denys14Samoilenko, line 35: Good job!!!
But i would make a couple of changes. 7 and 3 we can write in constants.
Screenshot_3

Denys14Samoilenko (changes requested):
Good job!!!
But i would make a couple of changes. 7 and 3 we can write in constants.
Screenshot_3

Denys14Samoilenko (approved):
Good work!!!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= LONG_TERM) {
+    return (COST_PER_DAY * days) - DISCOUNT_50;
+  }
+
+  if (days >= SHORT_TERM) {
+    return (COST_PER_DAY * days) - DISCOUNT_20;
+  }
+
+  return COST_PER_DAY * days;
```
roman-mirzoian, line 45: You now have `COST_PER_DAY * days` repeated three times - this can be transferred to a variable and reused.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const COST_PER_DAY = 40;
+  const DISCOUNT_50 = 50;
```
roman-mirzoian, line 32: No need to add what is stored there in the variable names, because the number of days can change and you will have to fix it in the entire code, not just in the variable definition.

roman-mirzoian (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const COAST_PER_DAY = 40;
+  const BIG_DISCOUNT = 50;
+  const SMALL_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const TOTAL_PER_DAYS = COAST_PER_DAY * days;
```
alexander-ignatow, line 36: Try move constants out of the function body when possible

```
  const COAST_PER_DAY = 40;
  const BIG_DISCOUNT = 50;
  const SMALL_DISCOUNT = 20;
  const LONG_TERM = 7;
  const SHORT_TERM = 3;

  const TOTAL_PER_DAYS = COAST_PER_DAY * days;
```

alexander-ignatow (approved):
LGTM 👍 

See the comment in the code

PR #3073, opened 2023-07-31
solution

olegpaunov:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * 40;
```
vserhiichuk, line 31: 1. [Checklist] Make names descriptive, so they explain the purpose of variables and functions.
and for the next lines

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+
+  if (days < 3) {
```
vserhiichuk, line 33: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
and for the next lines

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+    return result;
+  } else if (days >= 3 && days < 7) {
```
vserhiichuk, line 35: 4. [Checklist] Prefer `if` with return over `if else` to simplify later conditions.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
+    return result - 20;
```
vserhiichuk, line 36: 5. [Checklist] DON'T add `else` after `if` with return - the code after it won't be executed anyway.

vserhiichuk (changes requested):
Looks great..)
Few more things to fix in accordance to the checklist
And please look at the checklist again

src/calculateRentalCost.js
olegpaunov: are u seriosly ? -_-

vserhiichuk: I think it would be better to describe what you return as a final result (totalCost)

alexander-ignatow: @vserhiichuk @olegpaunov 

Try to be more clear in the comments. Describe the error more precisely, and when you don't agree with the comment please provide arguments why you should leave it as it is

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
alexander-ignatow, line 35: Declare the constants outside of the function body if possible
```
 const PRICE_PER_DAY = 40;
  const SHORT_TERM = 3;
  const LONG_TERM = 7;

...
 ```

alexander-ignatow (commented):
LGTM!

Good job fixing comments on the first iteration.

PR #3075, opened 2023-07-31
Fixed all

mr-olezha:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const totalySum = days * 40;
```
Wesses, line 31: Constants must be named in capital letters.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
```
Wesses, line 33: [Checklist - paragraph 2] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.
Same for other ifs.

Also you dont need this `if`, instead use `return` in the end of the function body.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return totalySum;
+  }
+
+  if (days >= 3 && days < 7) {
```
Wesses, line 37: You dont need the second condition, if you use the next `if` earlier.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return totalySum - 20;
```
Wesses, line 38: Create constants for expressions like these. [Cheklist]

Wesses (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const RENTAL_PRICE = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const BASE_PRICE = RENTAL_PRICE * days;
```
Moroz-Dmytro, line 36: It is customary to name static variables with a capital letter, if something is calculated in the variable, it is better to name it using the usual camelCase.

Moroz-Dmytro (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCONT = 50;
```
Moroz-Dmytro, line 33:
```suggestion
  const LONG_TERM_DISCOUNT = 50;
```

Moroz-Dmytro (approved):
👍

PR #3077, opened 2023-07-31
solution

vladyslav-svorin:


loralevitska (changes requested):
Almost done) Let's avoid `magic numbers` (0, 3, 7) and move them to constants with meaningful variable naming

sTorba24 (approved):
Good work!

PR #3078, opened 2023-07-31
add task solution

AngerDESTROYS:


Vitalii-Petruk (approved):
Good job

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  let price = days * PRICE_PER_DAY;
```
Moroz-Dmytro, line 37:
```suggestion
  const price = days * PRICE_PER_DAY;
```

Moroz-Dmytro: Better to use `const` here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  if (days >= LONG_TERM) {
+    return (price -= LONG_TERM_DISCOUNT);
```
Moroz-Dmytro, line 40: You don't need to change `price` variable, as you just return result here
Better to do: `return price - LONG_TERM_DISCOUNT;`

Moroz-Dmytro (changes requested):
Let's improve your solution

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
```
tiserett, line 32: LONG_TERM_DAYS

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
```
tiserett, line 34: SHORT_TERM_DAYS

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const price = days * PRICE_PER_DAY;
```
tiserett, line 37: basePrice is a better name I guess

tiserett (approved):
LGTM, but let's fix those minor issues

PR #3084, opened 2023-07-31
add task solution

NazarVotkanych:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCONT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCONT = 20;
+
+  const basePrice = days * PRICE_PER_DAY;
```
kaavka, line 37: 2. Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.
If you are using const it would be better to give it proper name (upper case and  '_' between words). 

```suggestion
  const BASE_PRICE = days * PRICE_PER_DAY;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCONT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCONT = 20;
```
kaavka, line 35: If possible you should move all constants out of the function.
```suggestion
  const PRICE_PER_DAY = 40;
  const LONG_TERM = 7;
  const LONG_TERM_DISCONT = 50;
  const SHORT_TERM = 3;
  const SHORT_TERM_DISCONT = 20;
  
   function calculateRentalCost(days) {
   const BASE_PRICE = days * PRICE_PER_DAY;
```

kaavka (changes requested):
Good job! Some minor fixes and ready to go.

src/calculateRentalCost.js
NazarVotkanych: Sure, but
It is customary to name static variables with a capital letter, if something is calculated in the variable, it is better to name it using the usual camelCase.

kaavka: Thanks, it's interesting no know. Than i am not right.

src/calculateRentalCost.js
tiserett, line 35: same here

tiserett (approved):
LGTM

But let's fix those minor variable comments

PR #3091, opened 2023-07-31
add task solution

kaavka:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
...
 
+const DAYS_FOR_MEDIUM_DISCOUNT = 3;
+const DAYS_FOR_GREAT_DISCOUNT = 7;
+const RENT_PER_DAY = 40;
+const GREAT_DISCOUNT = 50;
+const MEDIUM_DISCOUNT = 20;
+
 function calculateRentalCost(days) {
-  // write code here
+  const BASE_PRICE = days * RENT_PER_DAY;
```
bgutai, line 37: Good job, your code is clean!
But if  there something is calculated in the variable, it is better to name it using the usual camelCase.

ghost (changes requested):

Moroz-Dmytro (approved):
👍

PR #3092, opened 2023-07-31
Solution

DominikaKarmel:


DorotaLeniecDev (approved):
According to the checklist: _Don't use literals in expressions (HARDCODE). Instead use constants explaining this values_

Example:
```js
const LONG_TERM = 7;

}
```

PR #3097, opened 2023-08-01
add task solution

Galers:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+    return calculate - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= MEDIUM_TERM) {
+    return calculate - MEDIUM_TERM_DISCOUNT;
+  }
+
+  if (days < MEDIUM_TERM) {
+    return calculate;
+  }
```
VitaliyBondarenko1982, line 50:
```suggestion
  return calculate;
```

looks like redundant last if

VitaliyBondarenko1982 (changes requested):

TarasHoliuk (approved):

PR #3109, opened 2023-08-04
solution

oleg-matvienko:


BudnikOleksii (changes requested):
It works, but do the task in accordance with the [checklist ](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

alexander-ignatow (changes requested):
Much better 👍 

Still, you have one more magic number left

BudnikOleksii (approved):
Good job!

PR #3114, opened 2023-08-07
DONE

Venestella:


SerhiiTKO (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const DAY_PRICE = 40;
+  const DISCONT_1 = 20;
+  const DISCONT_2 = 50;
+  const TERM_SHORT = 3;
+  const TERN_LONG = 7;
```
djtyrf312, line 35: There is a typo in the variable name. 'TERN_LONG' should be 'TERM_LONG'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const DAY_PRICE = 40;
+  const DISCONT_1 = 20;
+  const DISCONT_2 = 50;
```
djtyrf312, line 33: The variable name 'DISCONT_2' is not clear. Consider using a more descriptive name, like 'DISCOUNT_LONG_TERM'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const TERM_SHORT = 3;
+  const TERN_LONG = 7;
+
+  if (days < TERM_SHORT) {
+    return DAY_PRICE * days;
+  }
+
+  if ((days >= TERM_SHORT) && (days < TERN_LONG)) {
```
djtyrf312, line 41: You don't need to check 'days >= TERM_SHORT' here because it's already covered in previous if statement. You can simplify this by just checking 'days < TERM_LONG'.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return DAY_PRICE * days - DISCONT_1;
+  }
+
+  if (days >= TERN_LONG) {
```
djtyrf312, line 45: You don't need to check 'days >= TERM_LONG' here because the previous conditions handle other cases. You can just use else statement here.

djtyrf312 (changes requested):
fix comments and remove package.json and package-lock.json files

ActorOfPlay (approved):

PR #3116, opened 2023-08-07
Solution

mpobihailo:


svitlanazai (approved):
good job)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
```
eule1, line 32: The constant LONG_TERM is not descriptive enough, it could be renamed to something like MIN_DAYS_FOR_LONG_TERM_DISCOUNT.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
```
eule1, line 31: You can move the PRICE_PER_DAY constant outside of the function, as it's a constant that doesn't change between function calls.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
```
eule1, line 33: You can move the LONG_TERM_DISCOUNT constant outside of the function, as it's a constant that doesn't change between function calls.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const SHORT_TERM = 3;
```
eule1, line 34: The constant SHORT_TERM is not descriptive enough, it could be renamed to something like MIN_DAYS_FOR_SHORT_TERM_DISCOUNT.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
```
eule1, line 35: You can move the SHORT_TERM_DISCOUNT constant outside of the function, as it's a constant that doesn't change between function calls.

eule1 (approved):

src/calculateRentalCost.js
mpobihailo: > You can move the PRICE_PER_DAY constant outside of the function, as it's a constant that doesn't change between function calls.

Thanks, I've already fixed it

src/calculateRentalCost.js
mpobihailo: > The constant LONG_TERM is not descriptive enough, it could be renamed to something like MIN_DAYS_FOR_LONG_TERM_DISCOUNT.

Thanks, I've already fixed it

src/calculateRentalCost.js
mpobihailo: > The constant SHORT_TERM is not descriptive enough, it could be renamed to something like MIN_DAYS_FOR_SHORT_TERM_DISCOUNT.

Thanks, I've already fixed it

src/calculateRentalCost.js
mpobihailo: > You can move the SHORT_TERM_DISCOUNT constant outside of the function, as it's a constant that doesn't change between function calls.

Thanks, I've already fixed it

PR #3117, opened 2023-08-07
Add my code

svitlanazai:


H07633R:
git commit -m should be 'Solution', you have 'Add my code'
brench from where you push should be 'develop', not 'master'

sorrow8 (changes requested):
git commit -m should be 'Solution', you have 'Add my code'
brench from where you push should be 'develop', not 'master'

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const basePrice = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return basePrice - LONG_TERM_DISCOUNT;
```
Huskov-Anton, line 40: The LONG_TERM_DISCOUNT and SHORT_TERM_DISCOUNT variables are fixed values, but they should be calculated as a percentage of the total price. This is because a discount is usually a percentage of the total amount. For example, a long term discount could be 10% off for rentals over 7 days.

Huskov-Anton (approved):

PR #3118, opened 2023-08-08
Solution

sorrow8:


olhahvozdetska (commented):
good job!

olhahvozdetska (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const midTerm = 3;
+  const longDiscount = 50;
+  const midDiscount = 20;
+  const price = 40 * days;
+
+  if (days >= longTerm) {
+    return price - longDiscount;
+  }
+
+  if (days >= midTerm) {
```
Huskov-Anton, line 41: The value of the 'midTerm' variable (3) is hardcoded. It's better to make it a parameter of the function to increase its flexibility and reusability.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const longTerm = 7;
+  const midTerm = 3;
+  const longDiscount = 50;
+  const midDiscount = 20;
+  const price = 40 * days;
```
Huskov-Anton, line 35: The calculation of the total rental cost (variable 'price') should be done after the condition checks. This is to ensure that we only perform the calculation if necessary, thus improving the performance of the function.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= longTerm) {
```
Huskov-Anton, line 37: The value of the 'longTerm' variable (7) is hardcoded. It's better to make it a parameter of the function to increase its flexibility and reusability.

Huskov-Anton (changes requested):

Huskov-Anton (approved):

PR #3121, opened 2023-08-08
Solution

olhahvozdetska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  const longRentalDiscount = 50;
+  const shortRentalDiscount = 20;
+
+  if (days >= 7) {
```
ViacheslavKhomov, line 35: Should add const for `7` and name it and the same is for `3`

ViacheslavKhomov (changes requested):
Nice job, just little fixes are required according to the checklist ^)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  const longTermOfDays = 7;
+  const longRentalDiscount = 50;
+  const shortTermOfDays = 3;
+  const shortRentalDiscount = 20;
+
+  if (days >= longTermOfDays) {
```
Huskov-Anton, line 37: Prefer if with return over if else to simplify later conditions.
DON'T add else after if with return - the code after it won't be executed anyway

Huskov-Anton (changes requested):

Huskov-Anton (approved):

PR #3129, opened 2023-08-09
Solution

YuriiRudyi:


MyroslavaLe (approved):
good job:)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const pricePerDay = 40;
+  const shortTerm = 3;
+  const shortTermDiscount = 20;
+  const longTerm = 7;
+  const longTermDiscount = 50;
+  const allDaysPrice = days * pricePerDay;
+
+  if (days >= longTerm) {
+    return allDaysPrice - longTermDiscount;
+  } else if (days >= shortTerm) {
```
Huskov-Anton, line 40: Prefer if with return over if else to simplify later conditions.
DON'T add else after if with return - the code after it won't be executed anyway

Huskov-Anton (changes requested):
Good work, little changes are required

src/calculateRentalCost.js
YuriiRudyi: fixed

Huskov-Anton (approved):

PR #3133, opened 2023-08-09
Solution

lkvashchuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,16 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const dayRent = 40;
+  let totalPrice = days * dayRent;
+
+  if (days >= 7) {
+    totalPrice -= 50;
```
Oleksii-Siviolov, line 35: It is better to make variables for the discounts.

Oleksii-Siviolov (approved):
GJ, but next time make variables for the values, that can be reused or changed frequently.

PR #3141, opened 2023-08-11
Solution

patrykyo20:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    return (days * cost) - 50;
+  }
+
+  if (days >= 3) {
+    return (days * cost) - 20;
+  }
+
+  return days * cost;
```
VitaliyBodnar, line 41: - move all numbers to the variables as it will make a code more readable and cleaner 
- `(days * cost)` is duplicated, so make sense to move it to the variable

VitaliyBodnar (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
```
mvjl000, line 33: Capitalize the names of this constant, they should also be in camel case (same as it is in the checklist).

By the way, such a constants can be outside of the function ;)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  const price = days * cost;
+
+  if (days >= 7) {
+    return price - bigDiscount;
+  } else if (days >= 3) {
+    return price - smallDiscount;
+  }
```
mvjl000, line 41: You could provide constants for those days values
```suggestion
  if (days >= LONG_TERM) {
    return price - bigDiscount;
  } else if (days >= SHORT_TERM) {
    return price - smallDiscount;
  }
```

mvjl000 (changes requested):

src/calculateRentalCost.js
```diff
@@ -26,18 +26,19 @@
  *
  * @return {number}
  */
+const cost = 40;
```
aureliagorzna, line 29: Capitalize the name of this constant - other than that everything looks good

aureliagorzna (changes requested):

mbruhler (approved):
Approved

PR #3152, opened 2023-08-14
Solution

maximmorenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,7 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  return days < 3 ? days * 40 : days < 7 ? days * 40 - 20 : days * 40 - 50;
```
IvanFesenko, line 31: Do not use nesting ternary operators - use if statements instead (without else or else if).
Also what does those magic numbers are? Create variable with respective names
Check [ckecklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

IvanFesenko (changes requested):
Good job. Lets make ypur code readable - check my comment

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < DAYS_FOR_MIN_DISCOUNT) {
+    return days * DAILY_RENT;
+  }
+
+  if (days < DAYS_F0R_MAX_DISCOUNT) {
+    return days * DAILY_RENT - MIN_DISCOUNT;
+  }
+
+  return days * DAILY_RENT - MAX_DISCOUNT;
```
IvanFesenko, line 45: You have same calculation 3 times - try do not repeat your self
```suggestion
  const basicCost = days * DAILY_RENT;
  
  if (days >= DAYS_F0R_MAX_DISCOUNT) {
    return basicCost - MAX_DISCOUNT;
  }

  if (days >= DAYS_FOR_MIN_DISCOUNT) {
    return basicCost - MIN_DISCOUNT;
  }

  return basicCost;
```

IvanFesenko (changes requested):
One more remark

src/calculateRentalCost.js
maximmorenko: of course, chack again pls

IvanFesenko (approved):
Pay attention:
RENT_ALL_DAYS - is not constant so it should named with camelCase.
Also my suggestion has different if cases - re-check it please

PR #3154, opened 2023-08-15
ADD: function calculateRentalCost

GennadiiLazovskii:
Create function calculateRentalCost


GennadiiLazovskii:
Create function calculateRentalCost

PR #3156, opened 2023-08-15
Solution

yuliiaaaaa:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const SEVEN_DAY_DISCOUNT = 50;
```
denys-danyliuk, line 31:
```suggestion
  const LONG_TERM_DISCOUNT = 50;
```
don't hardcode exact number of days, it may change

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const THREE_DAY_DISCOUNT = 20;
```
denys-danyliuk, line 32: same here

denys-danyliuk (approved):

PR #3159, opened 2023-08-16
task solution

TenetaJ:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const DISCOUNT_20$ = 20;
```
Anastasiia-Svintsova, line 31: Variable names should not contain symbols, like '$'. And try to void numbers in names. Rename to `SHORT_TERM_DISCOUNT` maybe. Or maybe you have a better option

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const DISCOUNT_50$ = 50;
+  const RENT_DAY = 40;
+  const MORE_7_DAYS = 7;
```
Anastasiia-Svintsova, line 34:
```suggestion
  const  LONG_TERM = 7;
```

Anastasiia-Svintsova (changes requested):
Nice start! Rename variables to improve understanding and follow naming conventions, please

TenetaJ:
done

Anastasiia-Svintsova (approved):

PR #3163, opened 2023-08-16
Solution

hrysta:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const THREE_DAY_DISCOUNT = 20;
+  const SEVEN_DAY_DISCOUNT = 50;
+
+  switch (true) {
+    case days < 3 :
```
StasSokolov, line 36: Снимок экрана 2023-08-22 в 15 03 18

In this case will be better use constants explaining this values

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const THREE_DAY_DISCOUNT = 20;
+  const SEVEN_DAY_DISCOUNT = 50;
+
+  switch (true) {
```
StasSokolov, line 35: Prefer `if` instead `switch`, `switch` better use when the number of conditions is large

StasSokolov1 (changes requested):
Hi. Almost done only few comments!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+  const SHORT_DAY_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (days < SHORT_TERM) {
+    return PRICE_PER_DAY * days;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
```
polosanya, line 41: Instead of combining the conditions, you can just firstly check if it's a long term, then if it's a short term. 
Just return the default price by default

polosanya (changes requested):
Almost perfect! A few more changes left

A1eeex (approved):
looks good to me

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (days < SHORT_TERM) {
+    return TOTAL_PRICE;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    return TOTAL_PRICE - SHORT_TERM_DISCOUNT;
+  }
```
StasSokolov, line 44: Instead of combining the conditions, you can just firstly check if it's a long term, then if it's a short term.

StasSokolov1 (approved):
LGTM🔥

PR #3169, opened 2023-08-18
Updated Solution

mykhailonl:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const BASIC_DISCOUNT = 20;
+  const PRICE = 40;
+  const LONG_STAY = 7;
+  const LONG_STAY_DISCOUNT = 50;
+
+  if (days < SHORT_STAY) {
+    return days * PRICE;
+  }
+
+  if (days >= SHORT_STAY && days < LONG_STAY) {
```
volodymyr-soltys97, line 41: There is no need to check if 'days' is greater than or equal to 'SHORT_STAY' again. It's already been checked in the previous if statement.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return days * PRICE - BASIC_DISCOUNT;
+  }
+
+  if (days >= LONG_STAY) {
```
volodymyr-soltys97, line 45: There's no need to check if 'days' is greater than or equal to 'LONG_STAY'. If the code execution has reached this point, 'days' is definitely greater than or equal to 'LONG_STAY'.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #3179, opened 2023-08-20
Update calculateRentalCost.js

0xzynex:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  if (days === 6) {
+    return 220;
+  }
+
+  if (days === 3) {
+    return 100
+  }
+
+  return 80;
 }
```
VitaliyBondarenko1982, line 44: pay attention on [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)
all values from task description write inte constants and use them for return some value.

VitaliyBondarenko1982 (changes requested):
Need fix.

PR #3180, opened 2023-08-21
calculateRentalCost solution

vasyakohotyk:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
 
 function calculateRentalCost(days) {
   // write code here
```
Viktor-Kost, line 31: Unused comment '// write code here' should be removed.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
...
+  const RENTAL_DAY_VALUE = 40;
+  const DISCAUNT_SHORT_TERM = 20;
```
Viktor-Kost, line 33: There is a spelling mistake in DISCAUNT_SHORT_TERM, it should be DISCOUNT_SHORT_TERM.

Viktor-Kost (changes requested):
Nice!
Just minor fixes needed

anastasiiavorobiova (approved):
Great work!

PR #3184, opened 2023-08-22
created realisation for calculateRentalCost

MykytaRudenko:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,28 @@
...
+
 function calculateRentalCost(days) {
   // write code here
-}
+  let totalCost = days * COST_OF_RENT;
+
+  if (days >= LONG_TERM) {
+    totalCost -= LONG_TERM_DISCOUNT;
+
+    return totalCost;
```
```suggestion
    return totalCost - LONG_TERM_DISCOUNT;;
```
looks better

etojeDenys (changes requested):

PR #3189, opened 2023-08-23
Develop

indigo04:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const DAY_PRICE = 40;
+  const BASE_PRICE = days * DAY_PRICE;
```
anastasiiavorobiova, line 33:
```suggestion
  const basePrice = days * DAY_PRICE;
```

anastasiiavorobiova: It's better to name it in camelCase notation because we count it based on the 'days' variable that could change and depend on function arguments

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/calculateRentalCost.js
indigo04: done

anastasiiavorobiova (approved):
Awesome!

PR #3192, opened 2023-08-24
js one

Xiaomimaxik:


Xiaomimaxik:
finish

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days === 6) {
+    return 220;
+  }
+
+  if (days === 3) {
+    return 100;
+  }
+
+  return 80;
```
IvanFesenko, line 43: What is the magic numbers?
You should calculate cost and not return constant value.
Also check [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) - there are some hints

IvanFesenko (changes requested):
You function should calculate value. What will be const in this case?
`calculateRentalCost(17)`

PR #3196, opened 2023-08-25
Solution

RUBIC0N5545:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,33 @@
...
+  let k = 1;
+
+  cost = NORMAL_COST * days;
+
+  if (days >= LONG_TERM) {
+    for (let j = 1; j <= days; j += LONG_TERM) {
+      cost -= LONG_TERM_DISC;
+    }
+
+    return cost;
```
etojeDenys, line 45: loop looks redundant here, we can use this condition:
```
return NORMAL_COST * days -  LONG_TERM_DISC
```

etojeDenys (approved):

etojeDenys (changes requested):
let's fix that

etojeDenys (approved):
lgtm

PR #3203, opened 2023-08-28
first commit

MaxMaksym:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+  const BASIC_DISCOUNT = 0;
+
+  const price = days * PRICE_PER_DAY;
+  const discount = days >= LONG_TERM ? LONG_TERM_DISCOUNT
+    : days >= SHORT_TERM ? SHORT_TERM_DISCOUNT : BASIC_DISCOUNT;
```
polosanya, line 40: Nested ternary operators make your code harder to read. Better use `if` with immediate `return` instead 
```
if (something) {
  return price - discount;
}

return price;
 ```

polosanya (changes requested):
Nice! Let's improve your code a bit)

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comment

PR #3212, opened 2023-08-30
add task solution

Viktoriia-Vita:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const discountBig = 50;
+  const discountSmall = 20;
+  const amountCost = days * price;
+  const daysBigDiscount = 7;
```
Esceype, line 35: Better use screaming snake case naming convention for constants like this

Esceype (changes requested):
GJ! Read the comment and use it in each const name

xiimiik (approved):
Nice work!

PR #3213, opened 2023-08-30
The function 'calculateRentalCost' was created

Dmytro-Bashlai:


anastasiia-tilikina (approved):
Well done!

FYI it's better to use `camelCase` for var names.

PR #3215, opened 2023-08-31
First Solution [Svystun Bohdan]

BSvystun2707:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const smallDiscount = 20;
+  let total = rentPrice;
+  const LONG_TERM = 7;
+  const SMALL_TERM = 3;
+
+  if (days >= LONG_TERM) {
+    total = rentPrice - bigDiscount;
+  } else if (days >= SMALL_TERM) {
+    total = rentPrice - smallDiscount;
+  } else if (days < SMALL_TERM) {
```
VeronikaSynieokaia, line 43: The condition 'days < SMALL_TERM' is unnecessary. If none of the previous conditions are met, the function will automatically return total. You can safely remove this condition.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
-  // write code here
+  const priceDay = 40;
+  const rentPrice = days * priceDay;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  let total = rentPrice;
+  const LONG_TERM = 7;
+  const SMALL_TERM = 3;
+
+  if (days >= LONG_TERM) {
```
VeronikaSynieokaia, line 39: Consider adding a comment here explaining why you are checking if days is greater or equal to LONG_TERM and why you are applying a discount in this case. It will improve code readability.

VeronikaSynieokaia (approved):
well done but check the comments

PR #3220, opened 2023-09-02
solution

avakiel:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
```
artemforsoff, line 32: If we are working with time, it is better to have the variable tell us in what units the value is stored.
For example, if the variable is storing seconds: `WORKING_TIME_IN_SECONDS`.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const BASE_PRICE = PRICE_PER_DAY * days;
```
artemforsoff, line 36: The variable `BASE_PRICE` is not a constant value, because it's calculated, so use camel case for naming.

artemforsoff: and the name base price doesn't fully say what's stored there, because as a developer I'm asking questions: "Price for what? Price for someday?". The variable stores the final price without discount, maybe it would be better to call it that.

artemforsoff (approved):
On the whole, the task was done well 🔥
and it is important to be very responsible when naming variables, as other developers will have to work with the written code in the future.

PR #3227, opened 2023-09-05
solution

iamgranovskaya:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  // eslint-disable-next-line prefer-const
+  let basePrice = days * PRICE_PER_DAY;
```
loralevitska, line 38: Remove comment. Try it write obvious code that explain all we need without comments)
```suggestion
  const basePrice = days * PRICE_PER_DAY;
```

loralevitska (changes requested):
Almost perfect) Let's improve

etojeDenys (approved):
lgtm

PR #3233, opened 2023-09-06
Solution

yslyshchenko:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  const cost = days * rent;
+  const longTermDiscount = 50;
+  const shortTermDiscount = 20;
+
+  if (days >= 7) {
+    return (cost - longTermDiscount);
```
```suggestion
    return cost - longTermDiscount;
```
brackets are redundant - just FYI

TarasHoliuk (approved):

PR #3239, opened 2023-09-10
Solution

robsza1122:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
```
darokrk, line 31: The per day cost of 40 might change in the future. Please consider declaring it as a constant at the top of your function for better maintainability.

e.g. `const price = 40;`

`const totalPrice = days * 40;`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const longtermDiscount = 50;
+  const shorttermDiscount = 20;
+
+  if (days >= 7) {
```
darokrk, line 35: The number 7 here is a magic number. Please avoid using magic numbers. Instead, declare a constant at the top of your function.

e.g.

`const longTermDays = 7`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+    result = result - longtermDiscount;
+  } else if (days >= 3) {
```
darokrk, line 37: The number 3 here is a magic number. Please avoid using magic numbers. Instead, declare a constant at the top of your function.

e.g. `const shortTermDays = 3;`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+  const longtermDiscount = 50;
```
darokrk, line 32: you have typo in those two variables names, please use camel case, `longTermDiscount`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  const shorttermDiscount = 20;
```
darokrk, line 33: same here, typo

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    result = result - longtermDiscount;
```
darokrk, line 36: just use,

`return totalPrice - longTermDiscount`

don't need to reassign value to variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  } else if (days >= 3) {
+    result = result - shorttermDiscount;
```
darokrk, line 38: same as above, just return values instead of reassigning:

`return totalPrice - shortTermDiscount`

darokrk (changes requested):
Not bad, left few mentions for improvements 👨🏽‍💻 🛠️

darokrk (changes requested):
Please fix mentions before re-request PR again :)

darokrk (approved):
Good job 💪🏽

PR #3247, opened 2023-09-12
calculate retntal cost

iLavruk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days >= LONG_TERM) {
+    return days * cost - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM) {
+    return days * cost - SHORT_TERM_DISCOUNT;
+  }
+
+  return days * cost;
 }
```
etojeDenys, line 46: we can move `days * cost` in some var

etojeDenys (approved):
good job

PR #3260, opened 2023-09-14
solved task

vlevik:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const ONE_DAY_PRICE = 40;
+  const BASE_PRICE = days * ONE_DAY_PRICE;
```
tilugdmytro, line 32: It would be better to keep simple constants all together, and this one with calculation somewhere underneath

tilugdmytro (approved):
Great job i believe!

Oleksandr-kopaevich (approved):
Hey @vlevik , good job, all the tests have passed.

PR #3263, opened 2023-09-15
Develop

yuliabakun:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days < MID_TERM) {
+    return DAILY_RENT * days;
+  }
+
+  if (days >= MID_TERM && days < LONG_TERM) {
+    return DAILY_RENT * days - MID_TERM_DISC;
+  }
+
+  return DAILY_RENT * days - LONG_TERM_DISC;
```
Helen-arch, line 45: you can use variable for calculating basic price instead of duplicating `DAILY_RENT * days`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const LONG_TERM = 7;
+  const MID_TERM_DISC = 20;
+  const LONG_TERM_DISC = 50;
+
+  if (days < MID_TERM) {
+    return DAILY_RENT * days;
+  }
+
+  if (days >= MID_TERM && days < LONG_TERM) {
+    return DAILY_RENT * days - MID_TERM_DISC;
```
Helen-arch, line 42: 4. [checklist] Prefer `if` with `return` over if else to simplify later conditions.

( better place long term `if` condition first to prevent repeating second condition here )

Helen-arch (changes requested):
Very nice! just a few fixes needed

Oleksandr-kopaevich (approved):
Hey, good job, all the tests have passed,

PR #3264, opened 2023-09-15
solution for calculateRentalCost task

ivaZaiets:


BetterSol (approved):
Good job! Everything looks efficient :)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,20 @@
...
+  const LONG_PERIOD = 7;
+  const SHORT_PERIOD = 3;
+
+  const LONG_PERIOD_DISCOUNT = 50;
+  const SHORT_PERIOD_DISCOUNT = 20;
+
+  let sum = ONE_DAY_RENT_COST * days;
+
+  sum = days >= LONG_PERIOD ? sum = sum - LONG_PERIOD_DISCOUNT + 20 : sum;
+  sum = days >= SHORT_PERIOD ? sum = sum - SHORT_PERIOD_DISCOUNT : sum;
```
Oleksandr-kopaevich, line 43: You can rewrite it like this 
```suggestion
  sum = days >= LONG_PERIOD ? sum - LONG_PERIOD_DISCOUNT + 20 : sum;
  sum = days >= SHORT_PERIOD ? sum - SHORT_PERIOD_DISCOUNT : sum;
```

Oleksandr-kopaevich (changes requested):
Hey, @ivaZaiets that's good that all the tests have passed, but in the attempt of trying to write fewer lines of code, you made it hard to read. You must learn to find the balance but readability and easiness are always in the first place when you work with a team, and in the real world you always work with a team. Here are details:
1 - do not use ternary operators instead of `if` operator. It's fine to use them when you have two different scenarios of variable definition, but in your case you change variable only if conditions is true and if false you just return the same value. It's better to use traditional `if` block for this, because in your case you have too much overwriting for the variable which is hard to keep in mind
2 - Here in the [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) the second points says `Don't use literals in expressions (HARDCODE).` and you have 20 in the formula
I feel bad for rejecting it, but you must fix the comments.

src/calculateRentalCost.js
Oleksandr-kopaevich: but in the end do not use here ternary operator at all

StasSokolov1 (approved):
Awesome job🚀

PR #3265, opened 2023-09-15
add task solving

Helen-arch:


YaroslavYarynych (approved):
Good job! Checklist is okay.

Oleksandr-kopaevich (approved):
Hey, good job, all the tests have passed, but please, I don't recommend using such an approach in the future. It's not common to write like this and it can be confusing. When you have conditions most of the developers will expect `if` block, and just because you can do it in some other clever way, it doesn't mean you need to do so. Always ask yourself, what will you gain if you write code like this, is it worth making it harder for other devs who thinks in the common way?

PR #3266, opened 2023-09-15
Develop

BetterSol:
transportation on vacation task solved

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -29,6 +29,7 @@
 
 function calculateRentalCost(days) {
   const PRICE_PER_DAY = 40;
+  let DAYS_RATE = days * PRICE_PER_DAY;
 
   const SHORT_TERM = 3;
   const LONG_TERM = 7;
@@ -36,19 +37,19 @@ function calculateRentalCost(days) {
   const SHORT_TERM_DISCOUNT = 20;
   const LONG_TERM_DISCOUNT = 50;
 
-  if (days < SHORT_TERM) {
-    return days * PRICE_PER_DAY;
-  }
-
-  if (days < LONG_TERM) {
-    return days * PRICE_PER_DAY - SHORT_TERM_DISCOUNT;
-  }
-
   if (days >= LONG_TERM) {
-    return days * PRICE_PER_DAY - LONG_TERM_DISCOUNT;
+    DAYS_RATE -= LONG_TERM_DISCOUNT;
+
+    return DAYS_RATE;
   }
 
-  return 'Looks like you didn\'t rent any car:)';
+  if (days >= SHORT_TERM) {
+    DAYS_RATE -= SHORT_TERM_DISCOUNT;
+
+    return DAYS_RATE;
+  }
+
+  return DAYS_RATE;
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,20 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  const ShortTermRent = days * 40;
-  const MediumTermRent = days * 40 - 20;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  const mediumTermDiscount = 20;
+  const longTermDiscount = 50;
+  const shortTermRent = days * pricePerDay;
+  const mediumTermRent = days * pricePerDay - mediumTermDiscount;
+  const longTermRent = days * pricePerDay - longTermDiscount;
+
+  if (days < 3) {
```
o-drozzdyk, line 38: 2. [CHECKLIST] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.

Please, add changes for all code lines

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
```
o-drozzdyk, line 31: It's better to name const variables using capital letters and underscores

o-drozzdyk (changes requested):
Nice job!)

But please, add some changes.

src/calculateRentalCost.js
BetterSol: thank you! I'm definitely fixing that

src/calculateRentalCost.js
BetterSol: it really is

StasSokolov1 (changes requested):
Almost done! Check a few minor comments please

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+  const PRICE_PER_DAY = 40;
+
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (days < SHORT_TERM) {
+    return days * PRICE_PER_DAY;
```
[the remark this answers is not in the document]
BetterSol: looks much better this way, thanks)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  let DAYS_RATE = days * PRICE_PER_DAY;
```
polosanya, line 32:
```suggestion
  const basicPrice = days * PRICE_PER_DAY;
```
image
Also, it's a calculated variable, so should be written in camelCase

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (days >= LONG_TERM) {
+    DAYS_RATE -= LONG_TERM_DISCOUNT;
+
+    return DAYS_RATE;
```
polosanya, line 43:
```suggestion
     return DAYS_RATE - LONG_TERM_DISCOUNT;
```
Just return the full price without reassigning

polosanya (changes requested):
Almost perfect!

src/calculateRentalCost.js
BetterSol: Thank you! I'll take this into account:)

src/calculateRentalCost.js
BetterSol: 👍

superpooperxxx (approved):

PR #3267, opened 2023-09-15
task is done

YaroslavYarynych:


src/calculateRentalCost.js
```diff
@@ -28,19 +28,19 @@
...
+  const SMALL_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
 
-  if (days < weekendTrip) {
-    return days * price;
-  } else if (days >= weekendTrip && days < weekTrip) {
-    return days * price - weekendDiscount;
+  if (days < SHORT_TERM) {
+    return days * PRICE_PER_DAY;
+  } else if (days >= SHORT_TERM && days < LONG_TERM) {
```
Drach-Illya, line 39: 5. DON'T add else after if with return - the code after it won't be executed anyway.

else is unnecessary here, you should remove it and just have two separate if constructions.

Drach-Illya (changes requested):

Drach-Illya (commented):
Great job. Fix this issue and I believe you should be good.

src/calculateRentalCost.js
YaroslavYarynych: Okay, I understand, thank u, I`m gonna do it now

Oleksandr-kopaevich (approved):
Hey, good job, all the tests have passed

PR #3268, opened 2023-09-15
add task solution

o-drozzdyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const MEDIUM_TERM = 3;
+  const MEDIUM_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const PRICE = 40;
+
+  const TOTAL_PRICE = days * PRICE;
```
OlhaKoziuk, line 37: it's not a static variable like LONG_TERM. We calculate it. We should write it in lowercase [totalPrice]

OlhaKoziuk (approved):
Good job! Please correct the name of variable

polosanya (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const MEDIUM_TERM = 3;
+  const MEDIUM_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const PRICE = 40;
+
+  let totalPrice = days * PRICE;
```
polosanya, line 37: image
You don't reassign this variable, no need to make it let

polosanya (changes requested):
Almost perfect! One small but important fix left

StasSokolov1 (approved):
LGTM!

PR #3269, opened 2023-09-15
add JS solution

Drach-Illya:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+    totalCost += DAILY_RENT;
+  }
+
+  if (days >= MEDIUM_TERM && days < LONG_TERM) {
+    totalCost -= REG_DISCOUNT;
+  }
+
+  if (days >= LONG_TERM) {
+    totalCost -= BIG_DISCOUNT;
+  }
```
mpohorenyi, line 48: 4. [CHECKLIST]Prefer `if` with `return` over `if else` to simplify later conditions.

If you use `return` inside an `if`, you will not need to write double conditions

mpohorenyi (changes requested):

mpohorenyi (changes requested):
Good job!

Please fix the code.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+  let totalCost = 0;
+  const DAILY_RENT = 40;
+  const MEDIUM_TERM = 3;
+  const REG_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const BIG_DISCOUNT = 50;
+
+  for (let i = 0; i < days; i++) {
+    totalCost += DAILY_RENT;
+  }
```
mpohorenyi, line 40: Why are you using a loop? You can do this with a simple multiplication operation

Drach-Illya: ~~I prefer it that way~~

Proof that I did it myself and not just by copying the solution in the video.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = 0;
+  const DAILY_RENT = 40;
+  const MEDIUM_TERM = 3;
+  const REG_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const BIG_DISCOUNT = 50;
+
+  for (let i = 0; i < days; i++) {
```
tetyana-pol, line 38: I agree with Michail, no needs to use a loop

tetyana-pol (approved):
I agree with Michael, no need a loop here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  const REG_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const BIG_DISCOUNT = 50;
+
+  totalCost = DAILY_RENT * days;
+
+  if (days >= MEDIUM_TERM && days < LONG_TERM) {
+    return totalCost - REG_DISCOUNT;
+  }
+
```
StasSokolov, line 43: can be simplify
```suggestion
  if (days >= MEDIUM_TERM) {
    return totalCost - REG_DISCOUNT;
  }

```

StasSokolov1 (approved):
Great job, i'm approving it with only one minor comment!

src/calculateRentalCost.js
Drach-Illya: We would need to swap the if constructions around, as 7 is higher than MEDIUM_TERM. We would need to check for long term, then medium term. But you are still correct, it is a retention from when I did not have return inside the if constructions. So yes, it can be simplified further.

PR #3272, opened 2023-09-15
solved the task

mifyy:


Stim1 (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicPrice = 40;
+  const THREE_DAYS_DISC = 20;
+  const SEVEN_PLUS_DAYS_DISC = 50;
+
+  if (days < 3) {
```
StasSokolov, line 35: let's create variable for `3` and reuse

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    return days * basicPrice;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * basicPrice) - THREE_DAYS_DISC;
```
StasSokolov, line 40: move to variables `days * basicPrice`

StasSokolov1 (changes requested):
Good job, let's improve your code a bit more!

mifyy:
Sure! I've made necessary corrections, please check my code once again.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const basicPrice = 40;
+  const THREE_DAYS_DISC = 20;
+  const THREE_DAYS = 3;
```
```suggestion
  const SHORT_TERM = 3;
```
Don't hardcode the numbers in variables' names. What if we need to change the term for small discount?
Fix all cases

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const SEVEN_PLUS_DAYS_DISC = 50;
+  const SEVEN_DAYS = 7;
+  const rawRentalCost = days * basicPrice;
+
+  if (days >= THREE_DAYS && days < SEVEN_DAYS) {
+    return (rawRentalCost) - THREE_DAYS_DISC;
```
polosanya, line 39: You can simplify it by checking if it's the long term first

polosanya (changes requested):
Almost done!

superpooperxxx (approved):
🔥

PR #3277, opened 2023-09-16
add task solution

MVoutch:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const SHORT_TIME = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const LONG_TERM = 7;
```
```suggestion
  const SHORT_TERM = 3;
  const SHORT_TERM_DISCOUNT = 20;

  const LONG_TERM = 7;
```
I think it's better to follow the same naming)

polosanya (approved):
Well done!

PR #3281, opened 2023-09-16
transportation on vacation task

serhii-mylyi:


src/calculateRentalCost.js
```diff
@@ -27,8 +27,35 @@
...
+const SECOND_OFF = 50;
+const LONG_TERM_3 = 3;
+const LONG_TERM_6 = 6;
+const LONG_TERM_7 = 7;
+
+
+
 function calculateRentalCost(days) {
-  // write code here
+  let finalCost;
```
TarasHoliuk, line 40:
```suggestion
  let finalCost = days * RENT_COST;
```
you may calculate the base cost initially instead of doing it in every block

src/calculateRentalCost.js
```diff
@@ -27,8 +27,35 @@
...
+
+  if (days < LONG_TERM_3) {
```
TarasHoliuk, line 42: you may check the long term first and immediately return calculated final value:

`return finalCost - SECOND_OFF;`

src/calculateRentalCost.js
```diff
@@ -27,8 +27,35 @@
...
+    finalCost = days * RENT_COST;
+
+    return finalCost;
+  }
+
+  if (days >= LONG_TERM_3 && days <= LONG_TERM_6) {
```
TarasHoliuk, line 48: if you check biggest value first of all it will allow you to get rid of `&&` here

src/calculateRentalCost.js
```diff
@@ -27,8 +27,35 @@
...
+    finalCost = (days * RENT_COST) - FIRST_OFF;
+
+    return finalCost;
+  }
+
+  if (days >= LONG_TERM_7) {
```
TarasHoliuk, line 54: here just return the final cost w/o discount (if redundant)

TarasHoliuk (changes requested):

VitaliyBodnar (approved):

PR #3287, opened 2023-09-17
solution

Clover247:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days < 3) {
```
dolphinnq, line 32: 2. [CHECKLIST] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.

Create some variables to make this code more understandable to others.

dolphinnq (changes requested):
Good job!
Add some changes to your code

src/calculateRentalCost.js
Clover247: Thank you

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const price = days * PRICE_PER_DAY;
```
```suggestion
  const basicPrice = days * PRICE_PER_DAY;
```
Not critical, but I think it's more accurate

polosanya (approved):
Well done!

PR #3291, opened 2023-09-18
Develop

HardCodWorker:


src/calculateRentalCost.js
```diff
@@ -28,19 +28,19 @@
...
 
   if (days < 3) {
-    return days * rentalPrice;
+    return days * rentalDayPrice;
   }
 
   if (days >= 3 && days < 7) {
-    return (days * rentalPrice) - mediumDiscount;
+    return (days * rentalDayPrice) - mediumDiscount;
   }
```
Ayogoharo, line 41: It would be easier to check from the top, so you don't need two conditions for the second `if`.
example 
`if ( days >= 7) {return}
if (days >= 3){return}
return`

Ayogoharo (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalDayPrice = 40;
+  const mediumDiscount = 20;
+  const bigDiscount = 50;
```
superpooperxxx, line 33: Those values are actual constants (Like `π` in math), we are defining their names in UPPERCASE.
image

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return days * rentalDayPrice;
+  }
+
+  if (days >= 3 && days < 7) {
+    return (days * rentalDayPrice) - mediumDiscount;
```
superpooperxxx, line 40: You are doing the same calculations `(days * rentalDayPrice)` in every if check. So I suggest putting it in a separate variable

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentalDayPrice = 40;
+  const mediumDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return days * rentalDayPrice;
+  }
```
superpooperxxx, line 37: image

superpooperxxx (changes requested):
Some improvements needed😢

HardCodWorker:
all changes applied

YevhenKharko (approved):
Seems OK to me, good job!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const SHORTERM = 3;
+  const LONGTERM = 7;
+
+  if (days < SHORTERM) {
+    return TOTALPRICE;
+  }
+
+  if (days >= SHORTERM && days < LONGTERM) {
+    return TOTALPRICE - MEDIUMDISCOUNT;
+  }
```
StasSokolov, line 44: can be simplify
```suggestion
  if (days >= SHORTERM) {
    return TOTALPRICE - MEDIUMDISCOUNT;
  }
```

StasSokolov1 (approved):
Great job, i'm approving it!

PR #3296, opened 2023-09-18
made function that calculate rental cost

Ayogoharo:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+
+  switch (true) {
+    case days >= LONG_TERM:
+      return (PRICE_PER_DAY * days) - LONG_TERM_DISCOUNT;
+    case days >= SHORT_TERM:
+      return (PRICE_PER_DAY * days) - SHORT_TERM_DISCOUNT;
+    default:
+      return (PRICE_PER_DAY * days);
+  }
```
superpooperxxx, line 44: 1. It is not a good practice to use `switch` case like this. Usually we pass a variable to test for some cases, but you pass a boolean. Better use `if` here.
2. You are duplicating `(PRICE_PER_DAY * days)`. Create a variable for this calculation.

superpooperxxx (changes requested):
Good job🔥 but I recommend some improvements

superpooperxxx (approved):
🔥🔥🔥

PR #3297, opened 2023-09-18
Completed the task

Maksim114488:


superpooperxxx (changes requested):
Hey, you basically ignored almost every item of [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 😢

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const shortTermDiscount = 20;
+  const shortTerm = 3;
+  const longTermDiscount = 50;
+  const longTerm = 7;
+  const carRentCost = 40;
+
+  if (days < shortTerm) {
+    return days * carRentCost;
```
polosanya, line 38: Let's create a variable for `days * carRentCost` in order not to duplicate it three times

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  }
+
+  if (days >= shortTerm && days < longTerm) {
```
polosanya, line 41: You can simplify it by checking for long term first

polosanya (approved):
Good job!

PR #3304, opened 2023-09-20
vacation car solution

donizer:


anastasiia-tilikina (changes requested):
let's put values for discount and days to variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const AMOUNT_$_PER_DAY = 40;
+  const DISCOUNT_7_DAYS = 50;
+  const DISCOUNT_3_DAYS = 20;
```
ivbutsykin, line 33: Don't use symbols like `$`, `%` in variable names. Same recommendation for numbers here (use words):
```suggestion
  const AMOUNT_PER_DAY = 40;
  const LONG_TERM_DISCOUNT = 50;
  const SHORT_TERM_DISCOUNT = 20;
```

PR #3307, opened 2023-09-21
Solution task transportation on vacation

garmoni:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
```
OleziO, line 28: Remove all comments from the code.

OleziO (changes requested):

loralevitska (approved):
🔥

PR #3328, opened 2023-09-29
Solution

TVChulkova:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const STANDART_TERM = 3;
+  const STANDART_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const PRICE = 40;
+  const BASE_PRICE = PRICE * days;
```
Viktor-Kost, line 36: usually we write variable names using uppercase if this value isn't changed (like you did with vars above)
```suggestion
  const basePrice = PRICE * days;
```

Viktor-Kost (approved):
GJ!

PR #3337, opened 2023-10-02
Solution

AlinaChukh:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days >= 7) {
+    return (days * rent) - 50;
+  }
+
+  if (days >= 3) {
+    return (days * rent) - 20;
+  }
+
+  if (days < 3) {
+    return days * rent;
```
xiimiik, line 42: make constant variables for values that are used several times in the code

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+
+  if (days >= 7) {
+    return (days * rent) - 50;
```
xiimiik, line 34: -50 and - 20 should be separate variables, because it is not clear from the context what they mean

xiimiik (changes requested):
gj, left some comments

xiimiik (approved):
nice work!

PR #3351, opened 2023-10-05
first commit

Intellium7:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+
```
Viktor-Kost, line 34:
```suggestion
const basicRent = RENT_VALUE * days
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rent = 40;
```
Viktor-Kost, line 31:
```suggestion
  const RENT_VALUE = 40;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+
+  if (days < 3) {
+    return days * rent;
+  }
+
+  if (days < 7) {
+    return days * rent - SHORT_TERM_DISCOUNT;
+  }
+
+  return days * rent - LONG_TERM_DISCOUNT;
```
Viktor-Kost, line 43: move day values to vars as well and reuse basicRent var

Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const RENT_VALUE = 40;
+  const BASIC_RENT = RENT_VALUE * days;
```
etojeDenys, line 32: this var is not the constant 'cause the value depends on the days argument

so let name it in camelCase

etojeDenys (changes requested):

maxim2310 (approved):
Good work!!!

PR #3368, opened 2023-10-09
Solution

Markianek:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
```
anastasiiavorobiova, line 31: Screenshot 2023-10-09 at 16 10 16
Consider following the checklist to write code according to the best practices

anastasiiavorobiova (changes requested):
Great work! Almost done!

IvanFesenko (approved):
Well done

PR #3370, opened 2023-10-09
add task solution

Tunik-hub:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceRent = 40;
+  let discount = 0;
+
+  if (days > 2 && days <= 6) {
```
Viktor-Kost, line 34: move day values to variables to avoid using magic numbers

Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
+  const pricePerDay = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const standardPrice = days * pricePerDay;
+
+  if (days >= LONG_TERM) {
+    return standardPrice - 50;
+  } else if (days >= SHORT_TERM) {
+    return standardPrice - 20;
+  }
```
Viktor-Kost, line 40: no needs to use `else if` when you use early return
```suggestion
  if (days >= LONG_TERM) {
    return standardPrice - 50;
  } 
  
  if (days >= SHORT_TERM) {
    return standardPrice - 20;
  }
```

Viktor-Kost (approved):
GJ!
Just make a minor fix

PR #3373, opened 2023-10-10
task solution

dkornienko31:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceOfDay = 40;
```
VitaliyBodnar, line 31: use also UPPER_CASE for priceOfDay

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):

PR #3375, opened 2023-10-11

Donovanoff:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+
+  if (days >= 3 && days < 7) {
+    result = price * days - basicDiscount;
+  } else if (days >= 7) {
+    result = price * days - maxDiscount;
+  } else {
+    result = price * days;
+  }
+
+  return result;
```
VitaliyBodnar, line 44: - can be less
```suggestion
  const basicPrice = price * days;

  if (days >= 3 && days < 7) {
    return basicPrice - basicDiscount;
  }

  if (days >= 7) {
    return basicPrice - maxDiscount;
  }
```
- also move 7 and to the constants 
- it's better to use UPPER_CASE for static constants

VitaliyBodnar (changes requested):

VitaliyBodnar (changes requested):
.

VitaliyBodnar (changes requested):
- not fixed 
Screenshot 2023-10-11 at 13 46 45

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const basicDiscount = 20;
+  const maxDiscount = 50;
+  const basicPrice = price * days;
+  const week = 7;
+
+  if (days >= 3 && days < week) {
```
lerastarynets, line 37: pls move 3 to const as well

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #3393, opened 2023-10-17
done task

dimakrugly:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const shortTermDiscount = 20;
+  const longTermDiscount = 50;
+
+  if (days >= shortTerm && days < longTerm) {
+    return resultValue - shortTermDiscount;
+  }
+
+  if (days >= longTerm) {
+    return resultValue - longTermDiscount;
+  }
```
sTorba24, line 43:
```suggestion
  if (days >= longTerm) {
      return resultValue - longTermDiscount;
   }
   
   // now you are absolutely sure that days < longTerm
   
  if (days >= shortTerm) {
    return resultValue - shortTermDiscount;
  }
```

sTorba24 (changes requested):
Well done! Lets improve a little

src/calculateRentalCost.js
dimakrugly: Done! All improvments added.
Also changed constants style to match SCREAMING_SNAKE_CASE

PR #3401, opened 2023-10-21
add task solution

mykola-kharin:


lerastarynets (approved):
Well done! Short and clear, I like it)

PR #3403, opened 2023-10-22
Solution

Miishca:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
```
anastasiiavorobiova, line 36:
```suggestion
const totalPrice = PRICE_PER_DAY * days;
```
Consider creating an additional variable to avoid code repeating

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  if (days < SHORT_TERM) {
+    return PRICE_PER_DAY * days;
```
anastasiiavorobiova, line 38:
```suggestion
    return totalPrice;
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome work!

PR #3404, opened 2023-10-23
Solution

MarekSax:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
+  if (days >= 7) {
+    result = result - 50;
+  } else if (days >= 3) {
+    result = result - 20;
```
volodymyr-soltys97, line 36: Consider will be better if you use two construction if, instead if/else if

src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
+  if (days >= 7) {
+    result = result - 50;
```
volodymyr-soltys97, line 34: Create a variables for all magic numbers

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

nazarmatsevych (approved):
Awesome 🔥

PR #3407, opened 2023-10-23
Solution

Shadi7733:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,31 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const basePrice = days * DAY_PRICE;
+
+  if (days >= LONG_TERM) {
+    totalCost = basePrice - LONG_TERM_DISCOUNT;
+
+    return totalCost;
```
etojeDenys, line 42:
```suggestion
    return  basePrice - LONG_TERM_DISCOUNT;
```
we can simplify it

etojeDenys: fix it everywhere

etojeDenys (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    return basePrice - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM) {
+    return basePrice - SHORT_TERM_DISCOUNT;
+  }
+
+  if (days < 3) {
+    return basePrice;
+  }
```
etojeDenys, line 48: actually if is redundant here, we can just return the basePrice

etojeDenys (changes requested):
nit:

etojeDenys (approved):
awesome!

PR #3408, opened 2023-10-23
Task solution

Laisens:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+
+  if (days < 3) {
+    return days * price;
+  }
+
+  if (days < 7) {
+    return days * price - 20;
+  }
+
+  return days * price - 50;
```
nazarmatsevych, line 46: Move this magic numbers '20' and '50' to variables

nazarmatsevych (changes requested):
Almost done, please check comments

src/calculateRentalCost.js
Laisens: i jak mi to teraz zrobić

nazarmatsevych (approved):
Well done!

PR #3411, opened 2023-10-24
Develop

verandetta:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  const shotTerm = 2;
+  const longTerm = 7;
```
etojeDenys, line 33: we usually name constants in upper case
image

etojeDenys (changes requested):

lerastarynets (approved):
You're doing amazing, keep it up!

PR #3416, opened 2023-10-25
Solutuin

BlackstarAngle:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  if (days > 2 && days < 7) {
+    return days * 40 - 20;
+  }
+
+  if (days > 6) {
+    return days * 40 - 50;
+  }
 }
 
 module.exports = calculateRentalCost;
```
kruladmiral, line 44: wow . what does it  do ; )?

BlackstarAngle: when there are at least 7 days it returns days * 40(because that's the price for a day) and -50 because that's a discount for 7 days or more

kruladmiral (approved):

src/calculateRentalCost.js
kruladmiral: I asked about module.exports. Never mind, sorry i didn't approve immediately as everything was fine.

anna-lymorenko (approved):
Well done!
For the future, please remember to check your code with checklist (in this case you should avoid hardcode - more info in the checklist)

PR #3431, opened 2023-10-26
Solution

mxpcsr:


maxim2310 (changes requested):
Good work, but would be nice to create constant for every magic number that you use here

mxpcsr:
i changed it and it still doesnt work

src/calculateRentalCost.js
```diff
@@ -26,9 +26,20 @@
  *
  * @return {number}
  */
+const DAILY_RATE = 40;
+const SEVEN_DAY_DISCOUNT = 50;
+const THREE_DAY_DISCOUNT = 20;
+const MAX_DAYS = 7;
+const MIN_DAYS = 3;
```
volodymyr-soltys97, line 33: You need to move these variables inside the function

src/calculateRentalCost.js
```diff
@@ -26,9 +26,20 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= MAX_DAYS) {
+    return DAILY_RATE * days - SEVEN_DAY_DISCOUNT;
+  } else if (days >= MIN_DAYS) {
+    return DAILY_RATE * days - THREE_DAY_DISCOUNT;
+  } else {
+    return DAILY_RATE * days;
+  }
```
volodymyr-soltys97, line 42:
```suggestion
  if (days >= MAX_DAYS) {
    return DAILY_RATE * days - SEVEN_DAY_DISCOUNT;
  }
  
  if (days >= MIN_DAYS) {
    return DAILY_RATE * days - THREE_DAY_DISCOUNT;
  }
  
  return DAILY_RATE * days;
```

volodymyr-soltys97 (changes requested):
Let's improve your code

PR #3434, opened 2023-10-26
Solution

ilyapgitt:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const price = days * PRICE_PER_DAY;
```
MrHunterBoi, line 36: Would add a few spaces here between blocks of `const` variables for better readability

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days < SHORT_TERM) {
+    return price;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
```
MrHunterBoi, line 42: There is no need for `days >= SHORT_TERM` condition since the previous if statement ensures that

MrHunterBoi (changes requested):
Nicely done

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
...
+
+  const LONG_TERM = 7;
+
+  const LONG_TERM_DISCOUNT = 50;
+
+  const SHORT_TERM = 3;
+
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const price = days * PRICE_PER_DAY;
```
roman-mirzoian, line 41: It is good practice to separate constants and regular variables, and even better to move immutable constants outside the function definition
```suggestion
  const PRICE_PER_DAY = 40;
  const LONG_TERM = 7;
  const LONG_TERM_DISCOUNT = 50;
  const SHORT_TERM = 3;
  const SHORT_TERM_DISCOUNT = 20;

  const price = days * PRICE_PER_DAY;
```

roman-mirzoian (approved):
Well done, but pay attention to the comment.

PR #3435, opened 2023-10-26
Solution

Petersiliana:


anna-lymorenko (approved):
Well done!
For the future, please remember to check your code with checklist (in this case you should avoid 'if else' - more info in the checklist)

PR #3436, opened 2023-10-26
made solution for calculateRentalCost function

MrHunterBoi:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,25 @@
...
-  // write code here
+  const BASE_RENT_COST = 40;
+
+  const SHORT_RENT_TERM = 3;
+  const SHORT_RENT_DISCOUNT = 20;
+
+  const LONG_RENT_TERM = 7;
+  const LONG_RENT_DISCOUNT = 50;
+
+  const rentCost = days * BASE_RENT_COST;
```
ArtemSydun, line 39: const values should be named with uppercase?)

ArtemSydun (commented):
good job

Yevhen-Kolosov (approved):
Well done! 👍 
No major mistakes were made according to the checklist. ✅

src/calculateRentalCost.js
BudnikOleksii: > const values should be named with uppercase?)

Only constant variables. In this case, the variable depends on the `days` parameter, so it's just a variable.

BudnikOleksii (approved):
Nicely done 🚀

PR #3437, opened 2023-10-26
Solution

zaura333:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const COST_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const MEDIUM_TERM = 3;
+  const MEDIUM_TERM_DISCOUNT = 20;
```
Patrykmclaren14, line 36: Good job. The only thing I would change is to move the highlighted variables in front of the functions to the global scope. They can help us later in another function.

patrykyo20 (changes requested):

mvjl000 (approved):
Nailed it!

PR #3438, opened 2023-10-26
solution

Barosz30:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let price = days * 40;
+
+  if (days >= 7) {
+    price -= 50;
+  } else if (days >= 3) {
+    price -= 20;
```
zaura333, line 37: Global variables could be implemented instead of literals (point 2 in checklist)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,15 @@
...
+  }
+
+  return price;
```
zaura333, line 40: Instead of using else if, you could use only if with return (as in point 4 of checklist)

zaura333 (changes requested):

mbruhler (approved):
Approved

mbruhler (changes requested):
Sorry, you have one change to go

constant values should be written UPPERCASE
Also, you can return just once, not in every if statement

Barosz30:
I made return where it's possible cuz that was on video, so that it wont iterate whole sentance. But thanks for uppercase tip :D

mbruhler:
both will work in regard to return statement but if you calculate to variable u can do it just once.
It depends on how you organize your code. Both will work so you can actually skip it.

mbruhler (changes requested):
variable constants are still camelCase. Should be UPPERCASE

Barosz30:
My bad, i didnt push branch xD Sorry

mbruhler (approved):
Approved

PR #3453, opened 2023-10-27
Solution

BoblykNazar:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+  const basePrice =  days * 40;
```
mykhailoyurkovskyi, line 35: create a variable for price per day and use it here!

mishawk (changes requested):

src/calculateRentalCost.js
BoblykNazar: I had it done, look at the second one commit=)

mishawk (approved):
nicely done👌👍

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+  const PRICE_PER_DAY = 40;
+  const basePrice = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return basePrice - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
```
roman-mirzoian, line 42: It is not necessary to check `days < LONG TERM`, because if this condition is not met, then the previous `if` will be executed

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+  const PRICE_PER_DAY = 40;
```
roman-mirzoian, line 35: Immutable constants are best defined outside the function definition

roman-mirzoian (approved):

PR #3454, opened 2023-10-27
add task solution

OstinM:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM_DISCOUNT = 50;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  let BASE_PRICE = days * PRICE_PER_DAY;
```
Fluebubble, line 36: `BASE_PRICE` is bad name for non-constants. `basePrice` is better.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days >= LONG_TERM) {
+    return BASE_PRICE - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM) {
+    BASE_PRICE -= SHORT_TERM_DISCOUNT;
```
Fluebubble, line 43: You can i `BASE_PRICE`
Better to return `BASE_PRICE` immediately return 11, as in previous `if`.
If you do this, you can declare `BASE_PRICE` as a `const` instead of `let`

Fluebubble (changes requested):
Nice work!😉 Add minor fix

Fluebubble (approved):
Well done!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM_DISCOUNT = 50;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const BASE_PRICE = days * PRICE_PER_DAY;
```
BudnikOleksii, line 36: Agree without previous comment, in UPPER_CASE we name only constants. In this case base price just a variable that depends on an argument `days`, so it should be in camelCase

BudnikOleksii (changes requested):
Good work 👍 
Just one comment, check it

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM_DISCOUNT = 50;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
roman-mirzoian, line 35: Immutable constants are best defined outside the function definition

roman-mirzoian (approved):
Well done 👍

src/calculateRentalCost.js
Fluebubble: @roman-mirzoian not in this case, because if you export a function it won't work anywhere else

PR #3457, opened 2023-10-27
solution

Fluebubble:


yevtok (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const PRICE_PER_DAY = 40;
+  const BASE_PRICE = days * PRICE_PER_DAY;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
```
roman-mirzoian, line 37: Immutable constants are best defined outside the function definition

src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const PRICE_PER_DAY = 40;
+  const BASE_PRICE = days * PRICE_PER_DAY;
```
roman-mirzoian, line 33: Static constants, such as strings or numbers, are called uppercase. 
If the constant is calculated in some way, then it is better to call it camelCase

roman-mirzoian (approved):

PR #3458, opened 2023-10-27
Solution

Anton-Kuchmasov:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,29 @@
 
 function calculateRentalCost(days) {
   // write code here
+  let totalPrice = 0;
```
mickhakk, line 32: 3. [checklist] Prefer const over let where possible, to avoid unintentional changes.
for example:
const totalPrice = priceOfRentPerDay * days;

mickhakk (changes requested):

src/calculateRentalCost.js
Anton-Kuchmasov: fixed, no need for this variable as well - we can return result in even one expression

mickhakk (commented):
Good job!!!

mickhakk (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const DISCOUNT_FOR_SHORT_TERM = 20;
+  const DISCOUNT_FOR_LONG_TERM = 50;
+
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
```
roman-mirzoian, line 36: Immutable constants are best defined outside the function definition

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days < SHORT_TERM) {
+    return (PRICE_PER_DAY * days);
+  }
+
+  if (days < LONG_TERM) {
+    return (PRICE_PER_DAY * days) - DISCOUNT_FOR_SHORT_TERM;
+  }
+
+  return (PRICE_PER_DAY * days) - DISCOUNT_FOR_LONG_TERM;
```
roman-mirzoian, line 46: You write the calculation `PRICE_PER_DAY * days` three times - this can be done once, saving the result in a variable and reusing this variable

roman-mirzoian (changes requested):

src/calculateRentalCost.js
```diff
@@ -27,8 +27,25 @@
...
+const PRICE_PER_DAY = 40;
+const DISCOUNT_FOR_SHORT_TERM = 20;
+const DISCOUNT_FOR_LONG_TERM = 50;
+
+const SHORT_TERM = 3;
+const LONG_TERM = 7;
+
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_ALL_TERM = PRICE_PER_DAY * days;
```
roman-mirzoian, line 38: It is customary to call immutable static constants (numbers, periods) uppercase, and if the variable is computational, it is better to call it camelCase:
```suggestion
  const finalPrice = PRICE_PER_DAY * days;
```

roman-mirzoian (approved):
Well done 👍

mickhakk (approved):

polosanya (approved):
👍

PR #3472, opened 2023-10-28
finished calculateRentalCost tack

kate-prakofyeva:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,9 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
+  return (days >= 7) ? (result -= 50) : (days >= 3) ? (result -= 20) : result;
```
bertsu02, line 33: Use constants that explain values, Checklist(2). Other than that all great !

src/calculateRentalCost.js
mvjl000, line 1: Please refer to the checklist. Your solution works however it's not ideal in terms of code quality.
Especially I would like you to bring here constants (for values like base price, long term, short term etc.). They should be outside the function so you could reuse them in other functions.

If you have to use multiple ternary operators at once it means that it is not the best use case for it. If statement would be much more readable. In the checklist you can read how to do it right.

mvjl000 (changes requested):

src/calculateRentalCost.js
kate-prakofyeva: Thank you, now I know how it's supposed to be

DanyloSemenov (approved):

mbruhler (approved):
Approved

PR #3474, opened 2023-10-28
task is done

YuriySobakar:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const price = PRICE_PER_DAY * days;
```
yefimov77, line 32: const TOTAL_PRICE = PRICE_PER_DAY * days;

1.Make names descriptive, so they explain the purpose of variables and functions. 
const = STABLE_CONST_NAME;

yefimov77 (approved):
That`s ok!

src/calculateRentalCost.js
roman-mirzoian, line 32: A previous reviewer made a good comment, but you don't have to uppercase the variable.
Uppercase is used only for static constants, that is, which are not calculated and do not change.
```js
const totalPrice = PRICE_PER_DAY * days;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
roman-mirzoian, line 36: Immutable constants are best defined outside the function definition

roman-mirzoian (approved):
Well done, but pay attention to the comments 👍

PR #3475, opened 2023-10-28
solution

arturgorniak:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let total;
+
+  if (days > 2 < 7) {
```
nazarmatsevych, line 33: Weird condition, if you want to have 2 conditions in one if statement then you need to combine them using logical operators like && or ||

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+    total = (days * 40) - 20;
```
nazarmatsevych, line 34: move `days * 40` to separate variable and reuse it to avoid code duplication

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  }
+
+  if (days <= 2) {
+    total = days * 40;
+  }
+
+  if (days >= 7) {
+    total = (days * 40) - 50;
```
nazarmatsevych, line 42: Move 'magic numbers' like '20', '40' and '50' to variables and use them

nazarmatsevych (changes requested):

nazarmatsevych (approved):
👍

PR #3485, opened 2023-10-30
solution

jacekkosiba:


mosowiecka (approved):

mbruhler (changes requested):
Please use UPPERCASE for constant values (dailyCost => DAILY_COST)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+  const DAILY_COST = 40;
+  let totalAmount = days * DAILY_COST;
+
+  if (days >= LONG_TERM) {
+    totalAmount -= LONG_TERM_DISCOUNT;
+
+    return totalAmount;
+  }
```
nataliaklonowska, line 42: You don't have to change the value of `totalAmount`; instead, return the result of the operation. This way, `totalAmount` can be a constant.
```suggestion
  const totalAmount = days * DAILY_COST;

  if (days >= LONG_TERM) {
     return totalAmount - LONG_TERM_DISCOUNT;
  }
```

nataliaklonowska (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+  const DAILY_COST = 40;
```
mvjl000, line 35: It is a good practice to put such a constants outside of the function. By doing so, you will be able to use then in other places in your code

mvjl000 (changes requested):
You nailed it!

There is just one small improvement 👌

mvjl000 (approved):
Perfect

PR #3486, opened 2023-10-30
task solution

Timi2424:


web3-sorcerer:
Hi Tymon, in order to pass your tests you have to change the line inside test.yml file:
'node-version: [14.x]' instead of [12.x] because as you can see your build was tested for that version.

web3-sorcerer (changes requested):
Your code lacks the ability to calculate the cost of rent when it's lower than 3 days + the thing I mentioned previously about test.yml

Timi2424:
Thank you, I missed these steps, hovewer the code should work fine. It calculates the cost on the beginning and then substract discount accordingly

src/calculateRentalCost.js
mvjl000, line 1: We need some improvements here. Read the checklist and make sure to apply its rules to your code ;)

Especially remember about constants (for values like base price, long term, short term, discounts etc) - those should be in uppercased snake case. It would be also good to put them outside of the function so you could use them in other places in your code.

Instead of using let, create a const and then inside of if statements, simply return the proper calculation

mvjl000 (changes requested):

mvjl000 (approved):
Perfect!

PR #3487, opened 2023-10-30
Develop

mosowiecka:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const dayCost = 40;
+  const weekDiscount = 50;
+  const threeDaysDiscount = 20;
+  const priceWithoutDiscout = days * dayCost;
+  let price = priceWithoutDiscout;
```
Patrykmclaren14: Read the checklist. Name of variables should be Uppercase

mosowiecka: commit [3dd7caa](https://github.com/mate-academy/js_task-transportation-on-vacation/pull/3487/commits/3dd7caa8a62452623264a9a4b300e29df7f248c2)

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+
+  if (days >= 3 && days < 7) {
+    price = price - threeDaysDiscount;
+  }
+
+  if (days >= 7) {
```
Patrykmclaren14: Replace numbers with variables. For example:

```diff

-  if (days >= 3 && days < 7) {
-    price = price - threeDaysDiscount;
-  }

- if (days >= 7) {

-  if (days >= 7) {

+  if (days >= SHORT_TERM && days < LONG__TERM) {
+     price = price - threeDaysDiscount;
+ }
+
+  if (days >= LONG__TERM) {

Comment (Patrykmclaren14) [inline src/calculateRentalCost.js]: You can also use **if** and **else if**.
```

  if (days >= 7) {
    --- code
  } else if (days >= 3) {
    --- code
  }

src/calculateRentalCost.js
```diff
@@ -29,6 +29,21 @@
...
+    price = price - weekDiscount;
```
Patrykmclaren14: You can make a change to make the code clearer.

```
price -= weekDiscount;
```

patrykyo20 (commented):
Read the checklist and based on it, correct the errors, and everything will be correct.

src/calculateRentalCost.js
mosowiecka: Read the checklist, point 4.

Timi2424:
Looks good, but mabye better use camel case for variables in the future

Timi2424 (approved):

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const DAY_COST = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUN = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUN = 20;
```
mvjl000, line 36: I would like you to move those constans outside of the function.

As previous reviewer said, it will allow you to use the values in other functions.

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
+  const priceWithoutDiscount = days * DAY_COST;
+  let price = priceWithoutDiscount;
```
mvjl000, line 38: I think you only need one constant here, let's call it `basePrice` 
```suggestion
  const basePrice = days * DAY_COST;
```
Then, inside of `if` statements you could return the proper calculation like this:
```
```

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    price = price - SHORT_TERM_DISCOUN;
+  }
+
+  if (days >= LONG_TERM) {
+    price = price - LONG_TERM_DISCOUN;
+  }
```
mvjl000, line 46: This works perfectly fine but you can simplify it.
Instead of checking for short term first, check the long term
```suggestion
  if (days >= LONG_TERM) {
  // ...
  }

  if (days >= SHORT_TERM) {
    // ...
  }
```

If the days value is greater than LONG_TERM, it will get inside of this if statement.
If it is not greater than LONG_TERM but greater than SHORT_TERM, it will get inside of second if statement.
If the value is less than both LONG & SHORT TERM, you return `basePrice` at the bottom of the function

mvjl000 (changes requested):

mbruhler (approved):
Approved, but please use proper English words while naming your variables (DISCOUN -> DISCOUNT)

PR #3488, opened 2023-10-30
add solution

olegnovikov10:


mishawk (approved):
Nicely done ✨

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const DAYS_PRICE = 40;
+  const SHORT_DISCOUNT = 20;
+  const LONG_DISCOUNT = 50;
```
roman-mirzoian, line 35: It is better to keep static constants outside the function definition so that they are not created anew each time the function is called.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= LONG_TERM) {
+    return days * DAYS_PRICE - LONG_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM) {
+    return days * DAYS_PRICE - SHORT_DISCOUNT;
+  }
+
+  return days * DAYS_PRICE;
```
roman-mirzoian, line 45: You call `days * DAYS_PRICE` three times - this can be stored in a separate constant and reused.

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Good work 👍

PR #3490, opened 2023-10-30
Solution

marcinzrodlowski:


michal-reszka (changes requested):
Repeating values ​​should be variables. Check readme.md second link.

michal-reszka (changes requested):
the variable should be written in large words

michal-reszka (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (DAYS_PROVIDED < 3) {
+    result = DAYS_PROVIDED * DAILY_PRICE;
+  } else if (DAYS_PROVIDED >= 3 && DAYS_PROVIDED < 7) {
+    result = (DAYS_PROVIDED * DAILY_PRICE) - SMALL_DISCOUNT;
+  } else if (DAYS_PROVIDED >= 7) {
+    result = (DAYS_PROVIDED * DAILY_PRICE) - BIG_DISCOUNT;
+  }
+
+  return result;
```
mvjl000, line 45: You need only two if statements. Remember about the checklist (point 4. - Prefer if with return over if else to simplify later conditions).
Simply check if it is a long term or short term. By the way, create constants for 3 & 7 (SHORT_TERM & LONG_TERM)

```suggestion
  if (DAYS_PROVIDED < 3) {
    result = DAYS_PROVIDED * DAILY_PRICE;
  } else if (DAYS_PROVIDED >= 3 && DAYS_PROVIDED < 7) {
   //example
    return = BASE_PRICE - SMALL_DISCOUNT;
  } else if (DAYS_PROVIDED >= 7) {
    result = (DAYS_PROVIDED * DAILY_PRICE) - BIG_DISCOUNT;
  }
  // You are returning this at the end as a default
  return BASE_PRICE;
```

mvjl000 (changes requested):
Always pay attention to the checklist. Please.

src/calculateRentalCost.js
```diff
@@ -27,8 +27,22 @@
...
+
 function calculateRentalCost(days) {
-  // write code here
+  const BASE_PRICE = days * DAILY_PRICE;
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    return BASE_PRICE - SMALL_DISCOUNT;
+  } else if (days >= LONG_TERM) {
+    return BASE_PRICE - BIG_DISCOUNT;
+  }
```
nataliaklonowska, line 43: checklist point 5 - you don't need to use `else if` when you have `return` inside an `if` statement :)

nataliaklonowska (changes requested):

mbruhler (approved):
Approved

PR #3491, opened 2023-10-30
Solution

michal-reszka:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
-}
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
```
szymanski-marcin, line 32: i would suggest to add one more const with discount for short-term rental and one const with basic price daily.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days < 1) {
+    return 0;
+  } else if (days >= LONG_TERM) {
+    return days * 40 - LONG_TERM_DISCOUNT;
+  } else if (days >= 3) {
+    return days * 40 - 20;
```
szymanski-marcin, line 39: try to create a  "let" variable and update your result in here instead of returning

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
 
+  return days * 40;
+}
```
szymanski-marcin, line 43: return here the variable i mentioned above.

szymanski-marcin (commented):
some minor things, overall its good!

szymanski-marcin (approved):

src/calculateRentalCost.js
JoyCoffeeAddict: Correct! Add another const

src/calculateRentalCost.js
JoyCoffeeAddict: There is nothing wrong with returning here.

JoyCoffeeAddict (changes requested):
Look at checklist points 4 and 5

JoyCoffeeAddict (approved):

PR #3495, opened 2023-10-30
Solution

Sibunas:


mbruhler (changes requested):
Hi,

- use constant values for constant and write them uppercase. In your code 40 is a BASE_PRICE as an example
- the thresholds should also be named (for example SHORT_TERM_DISCOUNT etc.)

When fixed it will be accepted

src/calculateRentalCost.js
```diff
@@ -29,13 +29,16 @@
 
 function calculateRentalCost(days) {
   let cost = 0;
+  const BASE_PRICE = 40;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
 
   if (days >= 3 && days < 7) {
```
kate-prakofyeva, line 36: You can do the same with days:
const LONG_TERM = 7;
const SHORT_TERM = 3;

kate-prakofyeva (approved):

src/calculateRentalCost.js
JoyCoffeeAddict: Agreed, you should use consts for these durations.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (days >= 3 && days < 7) {
+    cost = (days * BASE_PRICE) - SHORT_TERM_DISCOUNT;
+  } else if (days >= 7) {
+    cost = (days * BASE_PRICE) - LONG_TERM_DISCOUNT;
+  } else {
+    cost = days * BASE_PRICE;
+  }
```
JoyCoffeeAddict, line 42: Checklists point 4 and 5, prefer `if` with return over `if else`

JoyCoffeeAddict (changes requested):

JoyCoffeeAddict (approved):

PR #3497, opened 2023-10-30
Solution

MKolaszewska:


anna-lymorenko (approved):
Well done!
For the future, please remember to check your code with checklist (in this case you should avoid hardcode and 'if else' - more info in the checklist)

MKolaszewska:
> Well done! For the future, please remember to check your code with checklist (in this case you should avoid hardcode and 'if else' - more info in the checklist)

Ok, thanks for the feedback, Anna :)

PR #3506, opened 2023-11-02
add task solution

nazardudnyk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days === 7) {
+    return 230;
```
nazarmatsevych, line 32: What is 230, 220 and 100?

Use constant variables for `magic numbers` and calculate the total amount

nazarmatsevych (changes requested):

PR #3511, opened 2023-11-03
add task solution

pawelnowicki87:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,13 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return (days * 40);
+  } else if (days <= 6) {
+    return (days * 40) - 20;
+  }
```
nazarmatsevych, line 35: Since you return a value in each condition, you can create two separate if instead of an else if statement.

Also move "Magic numbers" like "20", "40" and "50" to separate constant variables

nazarmatsevych (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
-  // write code here
+  const price = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+
+  if (days < 3) {
+    return (days * price);
+  }
+
+  if (days <= 6) {
```
volodymyr-soltys97, line 39: Create a variable for these magic number 3 and 6

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

nazarmatsevych (approved):
Well done 👍

PR #3518, opened 2023-11-06
js

nazardudnyk:


src/calculateRentalCost.js
```diff
@@ -26,9 +26,24 @@
...
  */
+const LONG_TERM = 7;
+const SHORT_TERM = 3;
+const LONG_TERM_DISCOUNT = 50;
+const SHORT_TERM_DISCOUNT = 20;
+const DAILY_COST = 40;
 
 function calculateRentalCost(days) {
-  // write code here
+  const TOTAL_AMOUNT = days * DAILY_COST;
```
Viktor-Kost, line 36: usually we write var names using uppercase when this value isn't changed, yet TOTAL_AMOUNT depends on `days` that we accept as an function argument 
```suggestion
  const basicRent = days * DAILY_COST;
```

Viktor-Kost (approved):
GJ!

AbdullahWins (commented):
Well Done!

PR #3522, opened 2023-11-07

monikaiengibarian:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    total = total - 50;
+  }
+
+  if (days >= 3 && days < 7) {
+    total = total - 20;
+  }
+
+  return total;
```
sTorba24, line 41:
```suggestion
  const total = days * 40;

  if (days >= 7) {
    return total - 50;
  }

  if (days >= 3 && days < 7) {
    return total - 20;
  }

  return total;
```

sTorba24: also create a variables explaining 'magical' numbers 40, 50, 20

sTorba24 (changes requested):
Almost done! Lets improve it

monikaiengibarian:
Updated

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  let basePrice = days * PRICE_PER_DAY;
```
VitaliyBodnar, line 37: variable basePrice is not changed anywhere, so it supposed to be a const
```suggestion
  const basePrice = days * PRICE_PER_DAY;
```

VitaliyBodnar (changes requested):

src/calculateRentalCost.js
monikaiengibarian: Changed

nazarmatsevych (approved):
🔥

PR #3527, opened 2023-11-09
Solution

NowakTobiasz:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,14 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const result = {
+    2: 80,
+    3: 100,
+    6: 220,
+    7: 230,
+  };
```
nazarmatsevych, line 36: Interesting but looks like cheating.
In this task you have constant prices so you need to calculate the total amount. Create variables and make some calculations

nazarmatsevych (changes requested):
if you have any questions - feel free to ask them in fe_chat

PR #3538, opened 2023-11-10
Develop

Kozubowicz:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,13 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days < 3) {
+    return days * 40;
+  } else if (days < 7) {
+    return days * 40 - 20;
+  }
+
+  return days * 40 - 50;
```
nazarmatsevych, line 38: Move "magic numbers" like "40",  "50" and "20" to constant variables and use them.

Also create a separate variable for this calculation: days * 40, to avoid code duplication

nazarmatsevych (changes requested):

nazarmatsevych (approved):
👍

PR #3566, opened 2023-11-16
Task solution

ayasinska:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const PRICE_PER_DAY = 40;
+  const LONG_RENT = 7;
+  const LONG_RENT_DISCOUNT = 50;
+  const SHORT_RENT = 3;
+  const SHORT_RENT_DISCOUNT = 20;
+
+  let basePrice = days * PRICE_PER_DAY;
+
+  if (days >= LONG_RENT) {
+    return basePrice - LONG_RENT_DISCOUNT;
```
vlasiuksasha, line 40: The LONG_RENT_DISCOUNT should be subtracted from the basePrice only when the long rent duration is exactly equal to 7. If the rent duration is more than 7, the discount should be more.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  }
+
+  if (days >= SHORT_RENT) {
+    return basePrice - SHORT_RENT_DISCOUNT;
```
vlasiuksasha, line 44: The SHORT_RENT_DISCOUNT should be subtracted from the basePrice only when the short rent duration is exactly equal to 3. If the rent duration is more than 3 but less than 7, the discount should be more.

vlasiuksasha (changes requested):
check the comments

src/calculateRentalCost.js
ayasinska: https://mate.academy/chats/27190

src/calculateRentalCost.js
ayasinska: https://mate.academy/chats/27190

4ksandrey (changes requested):
try to resend task

PR #3583, opened 2023-11-21
CalculateRentalCost

OlgaBelmega:


ZhukSeeker:
Code is too complicated, you can do it much easier without that much of variables

ZhukSeeker (approved):
Code is too complicated, you can do it much easier without that much of variables

KarolinaVlasova (approved):

PR #3591, opened 2023-11-21
Task solution calculaterentalcost

natalialav:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
```
nikolaichuk-artem, line 31: Constants is common to use all-uppercase letters

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+
+  if (days >= longTerm) {
+    return days * pricePerDay - longTermDiscount
+  }
+
+  if (days >= shortTerm) {
+    return days * pricePerDay - shortTermDiscount
+  }
+
+  return days * pricePerDay
```
nikolaichuk-artem, line 45: You can make a variable with (days *  pricePerDay)
For example basePrice

nikolaichuk-artem (approved):
Good Job!
minor comments

KarolinaVlasova (approved):
GJ! But do not push package-lick.json and package.json files, and don't forget about semicolons

PR #3615, opened 2023-11-23
wrote code about rental cost

oskushnir:


lerastarynets (approved):
You're doing a fantastic job, keep going!

PR #3618, opened 2023-11-24
add task solution

vladyslav-zakharov:
1. Made names descriptive.
2. Don't used literals in expressions .
3. Used const.
4. Prefer if with return over if else to simplify later conditions.
5. Didn't add else after if with return - the code after it won't be executed anyway.


nazarmatsevych (approved):
🔥

PR #3620, opened 2023-11-26
 bSolution

przwojwwp:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,13 @@
 
 function calculateRentalCost(days) {
   // write code here
+  if (days >= 7) {
+    return (days * 40) - 50;
+  } else if (days >= 3) {
+    return (days * 40) - 20;
+  } else {
+    return (days * 40);
+  }
```
anastasiiavorobiova, line 38: There are a lot of magic numbers here. It's a bad practice, so read carefully the checklist to create constants with meaningful names https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md

anastasiiavorobiova (changes requested):
Great start! Keep doing!

yevhenii-pyl (approved):
Yep, you got the idea with variable naming.  👍 

One other thing here, is that these constants should be declared outside of the function scope. 

Here, by declaring them with the upper case notation, you are simulating the env variables, thus they can't be function scoped.

FYI, more about env variables: 
https://medium.com/chingu/an-introduction-to-environment-variables-and-how-to-use-them-f602f66d15fa

PR #3629, opened 2023-11-28
Solution

mat12722:


DariuszBPoz (changes requested):
Check Your code with checklist first

Khalandr (approved):
Please consider more clear variables' naming. while "longTermDiscount" is ok - "longlongTermDiscount" - is NOK!. In this particular case it may be: shortTermRent / longTermRent / shortTermDiscount / longTermDiscount

PR #3637, opened 2023-11-29
add task solution

GeorgiusX:
# Boilerplate for JS tasks

**Read [the guideline](https://github.com/GeorgiusX/js_task-guideline/blob/master/README.md) before start**


anastasiiavorobiova (approved):
Awesome work!

PR #3651, opened 2023-12-03
Solution

serjgrm:


nazarmatsevych (changes requested):
Proofread your code with [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  let totalPrice = 0;
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+  const DAILY_CAR_RENT = 40;
+
+  for (let i = 0; i < days; i++) {
+    totalPrice += DAILY_CAR_RENT;
+  }
```
VitaliyBodnar, line 40: can be less
```suggestion
 const totalPrice = days * PRICE_PER_DAY;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    totalPrice -= SHORT_TERM_DISCOUNT;
+  } else if (days >= 7) {
```
VitaliyBodnar, line 44: 7 is a magic number
Consider will be better if you create a enum for magic numbers and use it everywhere

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+  const DAILY_CAR_RENT = 40;
+
+  for (let i = 0; i < days; i++) {
+    totalPrice += DAILY_CAR_RENT;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    totalPrice -= SHORT_TERM_DISCOUNT;
```
VitaliyBodnar, line 43: don't need to change totalPrice in every case, you can just return new data
```suggestion
    return totalPrice - SHORT_TERM_DISCOUNT;
```

VitaliyBodnar (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,26 @@
...
+
+  totalPrice = days * PRICE_PER_DAY;
+
+  if (days >= term.SHORT && days < term.LONG) {
+    return totalPrice - discount.SMALL;
+  } else if (days >= term.LONG) {
+    return totalPrice - discount.BIG;
+  }
+
+  return totalPrice;
```
VitaliyBodnar, line 50: - don't need to change totalPrice 
- else is redundant cause after return the function will be stoped 
- it's better to user UPPER_CASE for static constant 
```suggestion
  const TERM = {
    SHORT: 3,
    LONG: 7,
  };
  const DISCOUNT = {
    SMALL: 20,
    BIG: 50,
  };
  const PRICE_PER_DAY = 40;
  const totalPrice = days * PRICE_PER_DAY;

  if (days >= TERM.SHORT && days < TERM.LONG) {
    return totalPrice - DISCOUNT.SMALL;
  }
  
  if (days >= TERM.LONG) {
    return totalPrice - DISCOUNT.BIG;
  }
  return totalPrice;
```

VitaliyBodnar (changes requested):

Viktor-Kost (approved):

PR #3678, opened 2023-12-10
add solution

kostrubitskii:


YevhenOstrovskiy (approved):
Good job!
You can create a variable to calculate total price without discount.

olya-shyian (approved):
Excellent!💥

PR #3684, opened 2023-12-11
add solution

Xilston:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
   // write code here
+  const ONE_DAY_RENT = 40;
+  const DISCOUNT_SHORT_COUNT = 20;
+  const DISCOUNT_LONG_COUNT = 50;
+  const DISCOUNT_SHORT_SHIFT = 3;
+  const DISCOUNT_LONG_SHIFT = 7;
+
+  const result = days * ONE_DAY_RENT;
+
+  if (days >= DISCOUNT_SHORT_SHIFT && days < DISCOUNT_LONG_SHIFT) {
```
sasholk, line 40: I would move it down, below `LONG_SHIFT` so in that case u don't need to write `&& days < DISCOUNT_LONG_SHIFT` and check in several conditions.

sasholk (approved):
Nice work! 🔥
All checklist points done 👍

src/calculateRentalCost.js
olya-shyian: good point )

olya-shyian (changes requested):
Well done! ☘️
Please update according to Oleksandr's comment.
His suggestion would be more readable

olya-shyian (approved):
Perfect! Good job!💥

PR #3694, opened 2023-12-11
solution

skiper404:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const basePrice = days * PRICE_PER_DAY;
```
dyarechuk, line 37: `const basePrice`
You're using const variable, it would be better to rewrite this variable with uppercase characters and underscores
After changing the variable, replace it in all places

dyarechuk (changes requested):
Good job! 🔥
Please fix your code

dyarechuk (approved):
Well done! 🔥

anstsot (approved):
Great job, approving :sparkles:

PR #3706, opened 2023-12-14
test

OleksandrFEdev:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const carRent = 40;
+  let fullCost = days * carRent;
+  const moreThan7DaySale = 50;
+  const moreThan3DaySale = 20;
+
+  if (days >= 7) {
+    fullCost -= moreThan7DaySale;
+  } else if (days >= 3) {
```
VitaliyBodnar, line 38: 7 and 3 are a magic numbers, so make sense to move them to the constants

src/calculateRentalCost.js
```diff
@@ -28,7 +28,18 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const carRent = 40;
```
VitaliyBodnar, line 31: it's better to use UPPER_CASE for static variables

VitaliyBodnar (changes requested):

src/calculateRentalCost.js
lerastarynets: not fixed

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #3715, opened 2023-12-16
solution

kapesha:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const RENT_PER_DAY = 40;
+  const SHORT_RENT = 3;
+  const SHORT_RENT_DISCOUNT = 20;
+  const LONG_RENT = 7;
+  const LONG_RENT_DISCOUNT = 50;
+  const TOTAL_RENT = days * RENT_PER_DAY;
```
Viktor-Kost, line 36: it's better to write this variable using camelCase since it's calculated value
```suggestion
  const totalRent = days * RENT_PER_DAY;
```

Viktor-Kost (approved):
Great job!

PR #3718, opened 2023-12-18
First task JS Advanced

Ulcja:
First task with rental cars. I've implemented 'calculateRentalCost' function, which returns the total amount for different count of days.


nazarmatsevych (approved):
Awesome 👍

PR #3729, opened 2023-12-21
first task

YuliiaKindruk:
https://github.com/YuliiaKindruk/js_task-transportation-on-vacation

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,20 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const price = days * PRICE_PER_DAY;
+
+  if (days >= 7) {
+    return price - 50;
+  }
+
+  if (days >= 3) {
+    return price - 20;
+  }
+
+  if (days < 3) {
+    return price;
+  }
 }
 
 module.exports = calculateRentalCost;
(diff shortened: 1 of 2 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+    return price - 50;
+  }
+
+  if (days >= 3) {
+    return price - 20;
+  }
+
+  if (days < 3) {
+    return price;
+  }
```
etojeDenys, line 44: rewrite your code following the checklist
image

etojeDenys (changes requested):

YuliiaKindruk:
Hello. When you will have a time to review my second attempt of this task?

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    return basePrice - LONG_TERM_DISCOUNT;
+  }
+
+  if (days>= SHORT_TERM) {
+    return basePrice - SHORT_TERM_DISCOUNT;
+  }
+
+  if (days < SHORT_TERM) {
+    return basePrice;
+  }
```
IvanFesenko, line 48: no need in `let` - you never change this value, use `const` instead. Last if is redundant - you will return base price in all other cases
```suggestion
  const SHORT_TERM_DISCOUNT = 20;
  const basePrice = days * PRICE_PER_DAY;

  if (days >= LONG_TERM) {
    return basePrice - LONG_TERM_DISCOUNT;
  }

  if (days>= SHORT_TERM) {
    return basePrice - SHORT_TERM_DISCOUNT;
  }

  return basePrice;
```

IvanFesenko (changes requested):
Good job, Small remarks

src/calculateRentalCost.js
YuliiaKindruk: ok

anastasiiavorobiova (approved):
Awesome work!

PR #3742, opened 2023-12-27
add task solution

bashtovaA:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days === 7) {
```
Viktor-Kost, line 31: avoid using `magic` numbers. Create variables instead

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+    return 230;
```
Viktor-Kost, line 32: this is hardcode. You should calculate value you need to return depends on an argument you get and discounts

Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const CAR_DAY = 7;
+  const CAR_DAY_SIX = 6;
+  const CAR_DAY_THREE = 3;
+  const CAR_DAY_DISKONT = 230;
+  const CAR_DAY_DISKONT_SIX = 220;
+  const CAR_DAY_DISKONT_THREE = 100;
```
lerastarynets, line 36: don't hardcode these values, you need to count them
ex
`const basePrise = days * PRICE_PER_DAY;`

and then return
`return basePrise - LONG_TERM_DISCOUNT;`

src/calculateRentalCost.js
```diff
@@ -28,7 +28,27 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const CAR_DAY = 7;
+  const CAR_DAY_SIX = 6;
+  const CAR_DAY_THREE = 3;
```
lerastarynets, line 33: these are unclear names of consts, pls prefer something like
```
const LONG_TERM = 7;
const SHORT_TERM = 3;
```

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+  const LONG_TERM = 7;
+  const SIX_TERM = 6;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+  const PRICE_PER_DAY = 40;
+
+  const basePrise = days * PRICE_PER_DAY;
+
+  if (days === LONG_TERM) {
```
anastasiiavorobiova, line 40:
```suggestion
  if (days >= LONG_TERM) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+    return basePrise - LONG_TERM_DISCOUNT;
+  }
+
+  if (days === SIX_TERM) {
+    return basePrise - SHORT_TERM_DISCOUNT;
+  }
```
anastasiiavorobiova, line 46:
```suggestion

```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,28 @@
...
+
+  if (days === SHORT_TERM) {
```
anastasiiavorobiova, line 48:
```suggestion
  if (days >= SHORT_TERM) {
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #3743, opened 2023-12-27
add task solution

ChrisHryts:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_RENT_DAYS = 7;
+  const SHORT_RENT_DAYS = 3;
+  const LONG_DISCONT = 50;
+  const SHORT_DISCONT = 20;
+  const PRICE_BASIC = days * PRICE_PER_DAY;
```
```suggestion
  const priceBasic = days * PRICE_PER_DAY;
```
It is a calculated variable so it shouldn't be in upper case

polosanya (approved):
Nice! 👍

PR #3748, opened 2023-12-31
add task solution

olduma:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const DAY_PRICE = 40;
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+  const TOTAL_PRICE = days * DAY_PRICE;
```
polosanya, line 36: This variable is calculated, so it shouldn't be in upper case
```suggestion
  const totalPrice = days * DAY_PRICE;
```

polosanya (approved):
Good job!

PR #3751, opened 2024-01-02
Solution

Helena-Krivichkaya:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const rentCost = 40;
+  const discountOne = 50;
+  const discountTwo = 20;
+  const cost = days * rentCost;
+
+  if (days < 3) {
+    return cost;
+  }
+
+  if (days >= 3 && days < 7) {
```
volodymyr-soltys97, line 40:
```suggestion
  if (days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+    return cost - discountTwo;
+  }
+
+  if (days >= 7) {
+    return cost - discountOne; 
+  }
```
volodymyr-soltys97, line 46:
```suggestion
  return cost - discountOne; 
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40;
+  const discountOne = 50;
+  const discountTwo = 20;
+  const cost = days * rentCost;
+
+  if (days < 3) {
```
volodymyr-soltys97, line 36: Create a variable for this magic number

src/calculateRentalCost.js
volodymyr-soltys97, line 40: And here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/calculateRentalCost.js
volodymyr-soltys97: Still not fixed from the previous review
image
```suggestion
  const shortTerm = 3;

  if (days < shortTerm) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+    return cost;
+  }
+
+  if (days < 7) {
```
volodymyr-soltys97, line 40: The same as above, create a variable for number and use it here

src/calculateRentalCost.js
```diff
@@ -28,7 +28,20 @@
...
+    return cost - discountTwo;
+  } else {
+    return cost - discountOne;
+  }
```
volodymyr-soltys97, line 44: else is redundant here, you can write only return
```suggestion
  }
  
  return cost - discountOne;
```

volodymyr-soltys97 (changes requested):
Almost done!

nazarmatsevych (approved):
Well done 👍

PR #3752, opened 2024-01-02
Solution

kravets111:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
+  const rentCar = 40;
+
+  if (days >= 7) {
```
Viktor-Kost, line 34: move all `magic` number to variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let result = 0;
```
Viktor-Kost, line 31: instead of variable redeclaring you can just set a default result and return this value for conditions (look an example below)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const rentCar = 40;
+
+  if (days >= 7) {
+    result = rentCar * days - 50;
```
Viktor-Kost, line 35: so here you will just return like this
```suggestion
    return baseRent - PRO_DISCOUNT;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days >= 3 && days < 7) {
+    result = rentCar * days - 20;
+  }
+
+  if (days > 0 && days < 3) {
+    result = rentCar * days;
+  }
+
+  return result;
```
Viktor-Kost, line 46: rewrite this code using variables and default `baseRent`

Viktor-Kost (changes requested):

src/calculateRentalCost.js
Viktor-Kost: isn't fixed. You should move all `magic` numbers to variables

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let discount = 0;
```
Viktor-Kost, line 31: this variable is redundant
You can either just create 4 different variables for `3`, `7`, `20`, `50` or group those variables into one object (advanced variant)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    discount = 50;
+
+    return rentCar - discount;
+  }
+
+  if (days >= 3 && days < 7) {
+    discount = 20;
+
+    return rentCar - discount;
+  }
```
Viktor-Kost, line 44: so you solution will looks like 
```suggestion
  if (days >= SOME_VAR && days < ANOTHER_VAR) {
    return rentCar - ONE_MORE_VAR;
  }
```

Viktor-Kost (changes requested):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+    return rentCar - discountSevenDays;
+  }
+
+  if (days >= threeDays && days < sevenDays) {
+    return rentCar - discountThreeDays;
+  }
+
+  if (days < threeDays) {
+    return rentCar;
+  }
```
IvanFesenko, line 47: no need in last if statement - by default should return basic cost
```suggestion
  return rentCar;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discountSevenDays = 50;
+  const discountThreeDays = 20;
+  const rentCar = 40 * days;
```
IvanFesenko, line 33: Not clear variable name
```suggestion
  const basicPrice = 40 * days;
```

IvanFesenko (approved):
Check my comments

PR #3753, opened 2024-01-02
add task solution

MrWhiteres:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
-  // write code here
+  const rendCostDay = 40;
+
+  const discount = {
+    bases: 20,
+    long: 50,
+  };
+
+  const correctDiscount = days >= 7 ? discount.long
+    : (days < 7 && days > 2) ? discount.bases : 0;
```
Viktor-Kost, line 39: move days `magic` numbers to variables. You can do similar to a `discount` object or even merge both objects

Viktor-Kost (changes requested):
Nice!
A minor fix needed

src/calculateRentalCost.js
MrWhiteres: Didn't quite get it

Viktor-Kost: `7` and `2` are [`magic numbers`](https://www.inf.unibz.it/~calvanese/teaching/04-05-ip/lecture-notes/uni04/node15.html)
So move them to variables.
Example: 
```
  const LONG_TERM = 7;
  const SHORT_TERM = 3;
```

or you can merge with an object you have
Something like:
```
const discounts = {
  base: {
    value: 20,
    term: 3,
  },
  pro: {
    value: 50,
    term: 7,
  }
};
```

P.S. since you've used `days >= 7` it's better to follow this type of condition and use `3` instead of `2`
```
days >= 7 ? discount.long
    : (days < 7 && days >= 3) ? discount.bases : 0;
```

Viktor-Kost (changes requested):

maxim2310 (approved):
Well done

PR #3760, opened 2024-01-04
add transportation on vacation solution

lbihun:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayRentCost = 40;
+  const shortTermDiscount = 20;
+  const longTermDiscount = 50;
+  const rentalCost = oneDayRentCost * days;
+  const shortTerm = rentalCost - shortTermDiscount;
+  const longTerm = rentalCost - longTermDiscount;
+
+  if (days < 3) {
```
volodymyr-soltys97, line 38: Сreate a variable for this magic number

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    return rentalCost;
+  }
+
+  if (days >= 3 && days < 7) {
```
volodymyr-soltys97, line 42: The same as above
```suggestion
  if (days < 7) {
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,24 @@
...
+    return shortTerm;
+  }
+
+  if (days >= 7) {
+    return longTerm;
+  }
```
volodymyr-soltys97, line 48: This if is redundant, so you need to remove it
```suggestion
  return longTerm;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

nazarmatsevych (approved):
👍

PR #3765, opened 2024-01-05
add solution

StageUa:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const lessWeeklyDiscount = 20;
+  const weeklyDiscount = 50;
+
+  if (days < shortTerm) {
+    return costDayCar * days;
+  } else if (days < longTerm) {
+    return costDayCar * days - lessWeeklyDiscount;
+  }
+
+  return costDayCar * days - weeklyDiscount;
```
Viktor-Kost, line 43: 1. move costDayCar * days to a variable
2. when you use `early return` inside `if` you can avoid using `else`
3. first condition should be for longTerm and then for shortTerm

Viktor-Kost (changes requested):

etojeDenys (approved):
awesome

PR #3776, opened 2024-01-09
The solution to the exercise was implemented

alionazherdetska:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,24 @@
...
+  const basicPrice = days * pricePerDay;
+  const lessThanThreeDaysCost = basicPrice;
+  const lessThanSevenDaysCost = days * pricePerDay - threeOrMoreDaysDiscount;
+  const sevenOrMoreDays = days * pricePerDay - sevenOrMoreDaysDiscount;
+
+  if (days < threeDaysStay) {
+    return lessThanThreeDaysCost;
+  } else if (days < sevenDaysStay) {
+    return lessThanSevenDaysCost;
+  }
```
nazarmatsevych, line 47:
```suggestion
  if (days < threeDaysStay) {
    return lessThanThreeDaysCost;
  } 
  
  if (days < sevenDaysStay) {
    return lessThanSevenDaysCost;
  }
```

We don't need to use an else block, because in each if we have a return statement

nazarmatsevych (approved):
Well done, left a minor comment

PR #3779, opened 2024-01-11
done

Fugas252mm:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const lowDiscount = 20;
+  const hiDiscount = 50;
+
+  if (days < midleTerm) {
+    return basePrice;
+  } else if (days >= midleTerm && days < longTerm) {
+    return basePrice - lowDiscount;
+  } else {
+    return basePrice - hiDiscount;
+  }
```
anastasiia-tilikina, line 43: let's get rid of if else chain)

```suggestion
  if (days < midleTerm) {
    return basePrice;
  }
  
    if (days >= midleTerm && days < longTerm) {
      return basePrice - lowDiscount;
    } 
  
  
    return basePrice - hiDiscount;
```

anastasiia-tilikina: when we use  return  inside if we don't need to use else after it

anastasiia-tilikina (changes requested):
GJ

let's just fix our code style a bit

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const BASE_PRICE = days * 40;
```
l4st1m0za, line 31: The 'BASE_PRICE' is calculated before the checks. It should be calculated after validating the number of days.

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
...
+  const MIDL_TERM = 3;
+  const LONG_TERM = 7;
+  const LOW_DISCOUNT = 20;
+  const HI_DISCOUNT = 50;
+
+  if (days < MIDL_TERM) {
+    return BASE_PRICE;
+  }
+
+  if (days >= MIDL_TERM && days < LONG_TERM) {
```
l4st1m0za, line 41: The condition 'days >= MIDL_TERM' is redundant because it's already checked in the previous if condition.

ghost (changes requested):
Good job, some small changes needed

ghost (approved):
Great job!

PR #3827, opened 2024-01-25
add task solution

Poloutinho:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,15 @@
...
-  // write code here
+  if (days < 3) {
+    return days * 40;
+  }
+
+  if (days < 7) {
+    return days * 40 - 20;
+  } else {
+    return days * 40 - 50;
+  }
```
sarakhmen, line 39: what do these constants mean? Please add corresponding const variables for these numbers. Also, pay attention to the code duplication: days * 40 is executed in any case so it's probably better to move it out of any condition

sarakhmen (changes requested):
Good job but let's improve your solution :)

PR #3839, opened 2024-01-29
task completed

Natalii0:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
```
Viktor-Kost, line 31: 1. create variables to all numeric values to avoid `magic` numbers

2. use `const` for a base rent
```
const baseConst = days * baseRent
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 7) {
+    totalCost -= 50;
```
Viktor-Kost, line 34: use early return 
```suggestion
    return baseConst - PRO_DISCOUNT;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  if (days >= 3 && days < 7) {
+    totalCost -= 20;
```
Viktor-Kost, line 38: early return

src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  }
+
+  return totalCost;
```
Viktor-Kost, line 41: if our condition don't match we just return baseCost
```suggestion
  return baseConst;
```

Viktor-Kost (changes requested):

src/calculateRentalCost.js
Natalii0: fixed

src/calculateRentalCost.js
Natalii0: fixed

src/calculateRentalCost.js
Natalii0: fixed

src/calculateRentalCost.js
Natalii0: fixed

Natalii0:
fixed

Natalii0 (commented):
fixed

PR #3861, opened 2024-02-05
Qa jan24 dmytro ped

Elementa523:


falatorry:
It is better to declare the totalCost variable using const since it does not change in the future

falatorry (approved):

Oleksii-Siviolov (approved):
Well done, Dmytro.

PR #3874, opened 2024-02-05
did the task

manch0ffline:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const priceOndeDay = 40;
+
+  if (days >= 7) {
```
Viktor-Kost, line 33: move `7`, `3`, `50`, `20`  to variables to avoid using `magic` numbers 

You can move `days * priceOndeDay` to a variable as well

Viktor-Kost (changes requested):

Viktor-Kost (approved):

PR #3885, opened 2024-02-08
Develop

DimaK88:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,31 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+  const PRICE_PER_DAY = 40;
+
+  if (days >= TERM_SEVEN_DAYS) {
+    return days * PRICE_PER_DAY - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= TERM_SIX_DAYS) {
+    return days * PRICE_PER_DAY - SHORT_TERM_DISCOUNT;
+  }
```
etojeDenys, line 45: why do we need that logic with 6 days, we don't have it in the task description

src/calculateRentalCost.js
```diff
@@ -28,7 +28,31 @@
...
+
+  if (days >= TERM_THREE_DAYS) {
+    return days * PRICE_PER_DAY - SHORT_TERM_DISCOUNT;
+  }
+
+  if (days >= TERM_TWO_DAYS) {
+    return days * PRICE_PER_DAY;
+  }
```
etojeDenys, line 53: same here

etojeDenys (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #3903, opened 2024-02-12
Added task solution

RochiFox:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  let totalSum = COSTS_PER_DAY * days;
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    totalSum -= SHORT_TERM_DISCOUNT;
+  }
+
+  if (days >= LONG_TERM) {
+    totalSum -= LONG_TERM_DISCOUNT;
+  }
```
Nirvacsh, line 45: 5. [Checklist] Better to use if (condition) + return 
image

Nirvacsh (approved):
Good job, small addition can be done to simplify code.

src/calculateRentalCost.js
BudnikOleksii: Good point 👍

BudnikOleksii (changes requested):
Good work 👍 
Check the previous reviewer comment and fix it accordingly

Moroz-Dmytro (changes requested):
Good work 👍 
But it can be simplified

Moroz-Dmytro (approved):
Looks good 👍

PR #3906, opened 2024-02-12
add solution

illya-onyshchuk:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+
+  const basePrice = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return basePrice - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    return basePrice - SHORT_TERM_DISCOOUNT;
+  }
```
vladarskyi, line 45: You can simplify second if like this: `  if (days >= SHORT_TERM) {
  }` 

because earlier you already checked that the number of days is less than LONG_TERM

vladarskyi, line 45: You can simplify second if like this:

```suggestion
  if (days >= SHORT_TERM) {
    return basePrice - SHORT_TERM_DISCOUNT;
  }
```
because earlier you already checked that the number of days is less than LONG_TERM

src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOOUNT = 20;
```
vladarskyi, line 34: small mistake in variable name

vladarskyi (changes requested):
Nice job 👍 
But you can do some improvements

Moroz-Dmytro (approved):
LGTM 👍

PR #3907, opened 2024-02-12
task solution

meljaszuk:
task solution

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -27,20 +27,17 @@
  */
 
 function calculateRentalCost(days) {
-  const dailyCost = 40;
-  const longTermDiscount = 50;
-  const mediumTermDiscount = 20;
+  const DAILY_COST = 40;
+  const LONG_TERM_DISCOUNT = 50;
+  const MEDIUM_TERM_DISCOUNT = 20;
   let rentalCost;
 
-  switch (true) {
-    case days < 3:
-      rentalCost = days * dailyCost;
-      break;
-    case days < 7:
-      rentalCost = days * dailyCost - mediumTermDiscount;
-      break;
-    case days >= 7:
-      rentalCost = days * dailyCost - longTermDiscount;
+  if (days >= 7) {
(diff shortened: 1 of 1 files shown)


yevhenii-pyl (changes requested):
Good job, however, we can make it even better: 

1. Variables naming declaration. Sure, currently you have them provided 'as it is' and camelCase may look perfectly fine in this case. But, if we try to imagine what these numbers really are - we may come to a conclusion that they are nothing more but a 'magic numbers' here: they come from nowhere and have no actual background. 
So let's make them more reasonable by thinking about their nature here. 

Most probably, in the real-life scenario these variables will be provided as: 
a) arguments to a function as a result of some other calculations or API call 
b) as environmental variables. 

In this case, we can't imagine having them from API or other functions, so let's assume they are environmental. In this case, we have to call them in upperCase with underscore as a separator: `dailyCost` => `DAILY_COST`, etc. 

2. `Switch(true)` statement: while it is a valid usage, however it is still debatable and some teams may be quite strict in denying such approach. (in details here: https://medium.com/trabe/using-switch-true-in-javascript-986e8ad8ae4f)
I personally would strongly recommend using a set of `if`'s here. 

3. Switch statements itself are quite error prompting: 
it works only due to placement order and if you really want to make it work, second condition should be `days >= 3 && < 7` since `days < 3` and `days < 7` have logical overlap. 

4. As a rule of thumb, each switch statement has to have a `default` case. 
Not exactly crucial, but even some linters will mark switch without default case as an error.

meljaszuk:
Thank you for the extensive feedback, corrections have been implemented.

yevhenii-pyl (changes requested):
Way better! 
Let's now remove all possible 'magic numbers' from your code. In future, terms of a discount may change, who knows where else we would like to use them. 
So let's move `7` and `3` to constants as well. 

Could you also think how this can be simplified even more? 
What if you firstly calculate a basic cost and then modify it(if necessary) applying a discount?  This may let you to shorten `if => else if => else` condition to just two(?) `if`'s.

yevhenii-pyl (approved):
Nailed it! 🚀

PR #3911, opened 2024-02-12
Solution

hattoriKanto:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayRentCost = 40;
+  const threeDayRentDiscount = 20;
+  const sevenDayRentDiscount = 50;
+  const smallTermRent = 3;
+  const longTermRent = 7;
+  const totalRent = oneDayRentCost * days;
```
DenNice-cloud, line 36: I think, thats okey, but as I see in other works, you should change all variables name from `oneDayRentCost` to `ONE_DAY_RENT_COST` for example.

DenNice-cloud (changes requested):
Everything looks brilliant!
Just little fix with all variables.))

src/calculateRentalCost.js
hattoriKanto: I think that camelCase variables are ok too, but I changed them as you said)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const ONE_DAY_RENT_COST = 40;
+  const THREE_DAYS_RENT_DISCOUNT = 20;
+  const SEVEN_DAYS_RENT_DISCOUNT = 50;
+  const SMALL_TERM_RENT = 3;
+  const LONG_TERM_RENT = 7;
```
Moroz-Dmytro, line 35: Move this variable outside the function

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+  const TOTAL_RENT = ONE_DAY_RENT_COST * days;
```
Moroz-Dmytro, line 36: This variable should be named in camelCase
`UPPER_CASE` -> constants(as above, lines 31-35)
`camelCase` -> calculated value

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const ONE_DAY_RENT_COST = 40;
+  const THREE_DAYS_RENT_DISCOUNT = 20;
+  const SEVEN_DAYS_RENT_DISCOUNT = 50;
```
Moroz-Dmytro, line 33:
```suggestion
  const SMALL_RENT_DISCOUNT = 20;
  const LONG_RENT_DISCOUNT = 50;
```

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days < SMALL_TERM_RENT) {
+    return TOTAL_RENT;
+  }
+
+  if (days >= LONG_TERM_RENT) {
+    return TOTAL_RENT - SEVEN_DAYS_RENT_DISCOUNT;
+  }
+
+  return TOTAL_RENT - THREE_DAYS_RENT_DISCOUNT;
```
Moroz-Dmytro, line 46: Better to change logic in `if`s to make `return TOTAL_RENT;` as a default return(the last one)
And `TOTAL_RENT - THREE_DAYS_RENT_DISCOUNT` should be in `if` block

Moroz-Dmytro (changes requested):
Good work 👍 
But it can be improved

Moroz-Dmytro (approved):
LGTM 👍

PR #3923, opened 2024-02-12
Develop

Oleksandr2108:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,22 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TIME = 7;
+  const SHORT_TIME = 3;
+  const LONG_TIME_DISCOUNT = 50;
+  const SHORT_TIME_DISCOUNT = 20;
```
roman-mirzoian, line 36: Static constants should be kept outside the function so that they are not recreated during multiple calls to the function.

roman-mirzoian (approved):
Well done, but pay attention to the comment 👍

PR #3925, opened 2024-02-13
add task solution

IRONH4MMER:


src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
 
 function calculateRentalCost(days) {
   // write code here
+  const basePrice = 40;
```
polosanya, line 32: Better name it `dayPrice` as it is a price for one day

src/calculateRentalCost.js
```diff
@@ -29,6 +29,23 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const finalPrice = days * basePrice;
```
polosanya, line 38: The name is a bit confusing. It is not a final price as it doesn't include discounts

polosanya (changes requested):

BudnikOleksii (approved):
👍

PR #3928, opened 2024-02-13
add task solution

FranbowWNS:
I'm not sure what happened but vscode somehow pushed changes as my secondary GH account. i'm not sure how bcause i never loginned into vscode from my secondary account.

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -29,14 +29,16 @@
 
 function calculateRentalCost(days) {
   const BASIC_RENTAL_COST = 40;
+  const SHORT_RENTAL_TERM = 3;
   const SMALL_DISCOUNT = 20;
+  const LONG_RENTAL_TERM = 7;
   const BIG_DISCOUNT = 50;
 
   let totalCost = days * BASIC_RENTAL_COST;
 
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
+  const BASIC_RENTAL_COST = 40;
+  const SMALL_DISCOUNT = 20;
+  const BIG_DISCOUNT = 50;
+
+  let totalCost = days * BASIC_RENTAL_COST;
+
+  if (days >= 3 && days < 7) {
+    totalCost = totalCost - SMALL_DISCOUNT;
+  } else if (days >= 7) {
+    totalCost = totalCost - BIG_DISCOUNT;
```
lerastarynets, line 40: pls move 3 and 7 to consts as well

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

FranbowWNS:
and also, do you know how to fix this strange vscode thing, git account that i'm using is this one, but all git actions happening from another account.
image

FranbowWNS:
I implemented requested fixes

PR #3934, opened 2024-02-14
Solution

mfcastilho:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+
+  if (days >= 3 && days < 7) {
+    return (rentCost * days) - 20;
+  }
+
+  if (days >= 7) {
+    return (rentCost * days) - 50;
+  }
+
+  return (rentCost * days);
```
jv-aquino, line 41: Also the (rentCost * days) is a recurring operation, maybe you could, like the values, store it somewhere for more readability

jv-aquino (changes requested):
Nice logic but with these new tips you can improve your code even more and make an amazing solution!

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40;
+  const totalToPay = rentCost * days;
+
+  if (days >= 7) {
```
jv-aquino, line 34: The conditional value (7) here is still being hardcoded, which is not a good practice

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40;
+  const totalToPay = rentCost * days;
```
jv-aquino, line 32: Consts should be uppercase

src/calculateRentalCost.js
```diff
@@ -28,7 +28,22 @@
...
+
+  if (days >= 7) {
+    const discount = 50;
```
jv-aquino, line 35: You could also declare the discounts before the ifs, but it's ok

jv-aquino (changes requested):
Your code improved but there's still improvements to make

jv-aquino (approved):
Great job! Your code really improved a lot and you are doing so much progress! Nice, Mario!

PR #3943, opened 2024-02-16
solution

Davwidos:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,10 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const discount = days >= 3 ? days >= 7 ? 50 : 20 : 0;
```
TideTwo, line 31: The checklist suggests that you don't use literals in expressions (point 2), instead of writing `3` create a variable above it: ```const SMALL_DISCOUNT_REQUIREMENT = 3;``` and use that in your expression, so others understand what the number means.

TideTwo (changes requested):

TideTwo (approved):
From what I understand, you ONLY need named variables for expressions (`if (*expression here*)`) and I don't think there is anything against the tertiary operator (` ? : `) but this works fine, and I don't see any other issues.

choeqq (approved):

PR #3946, opened 2024-02-17
add solution test

thisIvanRepo:
tessts passed

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -28,7 +28,15 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
+
+  if (days >= 3) {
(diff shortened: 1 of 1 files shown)


nazarmatsevych (changes requested):
Great work but proofread your code with [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const minDiscount = 20;
+  const maxDiscount = 50;
+
+  if (days >= 7) {
```
etojeDenys, line 35: better to move 7 and 3 magic numbers to some constants too

etojeDenys (approved):
good job

PR #3953, opened 2024-02-18
Solution

konradwasilewski444:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,17 @@
...
+  if (days >= 0) {
+    if (days < 3) {
+      return Number(days) * 40;
+    }
+
+    if (days < 7) {
+      return Number(days) * 40 - 20;
+    }
+
+    return days * 40 - 50;
```
choeqq, line 40: please try to add variables when using some specific numbers, when someone will do your code review, they also need to know what those numbers mean. In our case it's obvious, but typically it's a good practice

choeqq (changes requested):

choeqq (changes requested):
nice, typically we use `UPPERCASED_VALUES` for constants like numbers

choeqq (approved):

PR #3959, opened 2024-02-19
Solution

OlesPartyka:


TideTwo (approved):

src/calculateRentalCost.js
```diff
@@ -28,7 +28,21 @@
  */
 
 function calculateRentalCost(days) {
-  // write code here
+  const shortTerm = 3;
```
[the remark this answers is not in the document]
asakevych: I agree with @TideTwo , it's a good practice to use capitalised letters and snake case for consts

asakevych (changes requested):
Good job, a small change is required :wink:

src/calculateRentalCost.js
```diff
@@ -28,17 +28,17 @@
...
 
-  if (days < shortTerm) {
+  if (days < SHORT_TERM) {
     return days * rentalPricePerDay;
   }
 
-  if (days < longTerm) {
+  if (days < LONG_TERM) {
     return (days * rentalPricePerDay) - threeDaysBonus;
   }
```
jv-aquino, line 43: You could simplify both of these IFs with a simpler return using the ternary operator (?), creating a tighter and smoother code.
Could you come up with this improvement?

jv-aquino (commented):
Nice variable naming and general structure! Your code is clean, easily modifiable and really seams like code from real world codebases! There's just one nice thing that you could use to simplify it that I commented about

jv-aquino (approved):

src/calculateRentalCost.js
OlesPartyka: in my opinion the first option is more understandable

PR #3964, opened 2024-02-19
solution

marinatea:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const totalCost = costPerDay * days;
+
+  if (days >= 7) {
+    return totalCost - 50;
+  } else if (days >= 3 && days < 7) {
+    return totalCost - 20;
+  }
```
OlesPartyka, line 14: Put the numbers in variables, so it will be clear what is happening in the code

OlesPartyka (approved):

src/calculateRentalCost.js
choeqq: yes, I agree with @OlesPartyka, variables should explain what 50 / 7 / 3 means to a person who never read this code before, also when declaring constant like 40 you should use `UPPERCASED_NAME`

choeqq (changes requested):

choeqq (approved):

PR #3967, opened 2024-02-19
solution

kubsen:


choeqq (changes requested):
please add constant variables for all those hardcoded numbers, we'd like to be as descriptive as possible to a person who has never seen this piece of code. Please remember the names should be `UPPERCASED` when declaring constant variables

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const DAYCOST = 40;
+  const WEEKLENGTH = 7;
+  const WEEKOFF = 50;
+  const THREEDAYS = 3;
+  const THREEDAYSOFF = 20;
```
asakevych, line 11: please check once more the checklist file, you have examples there, use snake_case to make your constants better readable and also a good practice it to have them outside of the function, in the beginning of the file

asakevych (changes requested):
Good job! small change is needed :wink:

asakevych (changes requested):
Good job! small change is needed :wink:

asakevych (approved):
Nice! :fire:

PR #3968, opened 2024-02-19
Solution

luishenriquetp:
Else if added


src/calculateRentalCost.js
```diff
@@ -28,7 +28,23 @@
...
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const basePrice = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return basePrice - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM) {
```
jv-aquino, line 43: There's a better practice here instead of using two plain IF's, can you guess what is it?

jv-aquino (changes requested):
Good job naming the consts and nice code overall!

src/calculateRentalCost.js
luishenriquetp: Switch?! Or you mean to use Else?

luishenriquetp: Hello João. I think I could solve every point you mentioned. Any doubts, please let me know. Thank you.

EdPirro (approved):
Great job, @luishenriquetp! Keep it up 🚀

PR #3970, opened 2024-02-20
add task solution

Ditiselit:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const numberOfDays = days;
+  const priceDay = 40;
+  let totalPrice = priceDay * numberOfDays;
+  const shortTerm = 3;
+  const shortTermDiscount = 20;
+  const longTerm = 7;
+  const longTermDiscount = 50;
```
lerastarynets, line 13: transOnVacChlist

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

PR #3971, opened 2024-02-20
Solution

bartKarb:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let result = days * 40;
+
+  if (days >= 7) {
+    result = result - 50;
```
Kacper7001, line 10: result -= 50; i think its better and for sure shorter

Kacper7001 (changes requested):
I think its better to create const "dailyRate" or some, to keep the price of daily rental, instead of containing this value in results.

src/calculateRentalCost.js
bartKarb: Code shortened as above

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
```
Kacper7001, line 7: const x = 40;
let result = days *  x

this example is better for wide project, if something is gonna change you need to fix it in one line "const x = 40;"

Kacper7001 (changes requested):

src/calculateRentalCost.js
bartKarb: Transformed 40 into const

  const dailyRate = 40;
  let result = days * dailyRate;

Kacper7001 (approved):

choeqq (changes requested):
We should create a constant for each of those hardcoded numbers, please remember that the name of the variables should be `UPPERCASED`

choeqq (changes requested):
Please make sure to adjust your solution to my latest comment above ⬆️

bartKarb:
I just adjusted the task solution for uppercase constants convention

choeqq (approved):

PR #3974, opened 2024-02-20
Solution

sonyasapsan:


sarakhmen (approved):
Good job, but you can improve your solution if you want ;)

PR #3990, opened 2024-02-24
Add task solution

deezhizyu:
Also consider updating eslint in repository, i was getting `TypeError: Error while loading rule 'prettier/prettier': context.getPhysicalFilename is not a function` in tests without update


nazarmatsevych (approved):
Awesome 🔥

PR #4008, opened 2024-03-01
Solutionv2

kgGugala:


src/calculateRentalCost.js
```diff
@@ -28,7 +28,19 @@
...
 
 function calculateRentalCost(days) {
-  // write code here
+  const standardPrice = 40;
+  const shortTerm = 3;
+  const midTermDiscount = 20;
+  const longTerm = 7;
+  const longTermDiscount = 50;
+
+  if (days < shortTerm) {
```
yevhenii-pyl, line 37: Let's think how can we simplify the condition here.

yevhenii-pyl (changes requested):
GJ 👍 
Let's adjust variable declarations and simplify the condition.

kgGugala:
I just made new commit, however I'm not 10% sure that you expected that kind of simplification. If this is not what you expected please give me any hint so I can figure it out.

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
...
+const STANDARD_PRICE = 40;
+const SHORT_TERM = 3;
+const MID_TERM_DISCOUNT = 20;
+const LONG_TERM = 7;
+const LONG_TERM_DISCOUNT = 50;
+
 function calculateRentalCost(days) {
-  // write code here
+  if (days < SHORT_TERM) {
+    return days * STANDARD_PRICE;
```
yevhenii-pyl, line 38: Can't it be a default case? 

Like just `return days * STANDARD_PRICE` if no conditions were met?

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
...
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    return days * STANDARD_PRICE - MID_TERM_DISCOUNT;
```
yevhenii-pyl, line 42: Even better, why don't you create a const for standardTerm or something that will be responsible for  this calculation? 

`days * STANDARD_PRICE`

yevhenii-pyl (changes requested):
Just one more step :)

kgGugala:
Please ignore my last commit as I didn't see your one comment - I'm already working on it.

src/calculateRentalCost.js
```diff
@@ -27,8 +27,24 @@
...
 
+const STANDARD_PRICE = 40;
+const SHORT_TERM = 3;
+const MID_TERM_DISCOUNT = 20;
+const LONG_TERM = 7;
+const LONG_TERM_DISCOUNT = 50;
+
 function calculateRentalCost(days) {
-  // write code here
+  const STANDARD_TERM = days * STANDARD_PRICE;
```
yevhenii-pyl, line 37: And this is just an internal variable that you use inside the function. It's not supposed to 'mimic' env variable.
This one should follow standard naming camelCase convention.

Don't perceive this as a jest, at this point it's just really important for you to grab the basics.

yevhenii-pyl (changes requested):
One very very last fix and I'll let you be ✊

yevhenii-pyl (approved):
🥇

PR #4020, opened 2024-03-04
Solution

ArtemDemeniev:


metrakonon (approved):
Well done! Everything looks good.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
-  // write code here
-}
+  const MEDIUM_TERM = 3;
+  const MEDIUM_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const rentalCost = 40;
+
+  if (days < MEDIUM_TERM) {
+    return rentalCost * days;
```
polosanya, line 14: Let's create a variable for `rentalCost * days` to avoid duplication

polosanya (changes requested):
Almost perfect!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
-  // write code here
-}
+  const MEDIUM_TERM = 3;
+  const MEDIUM_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const rentalCost = 40;
+  const finalCost = rentalCost * days;
+
+  if (days < MEDIUM_TERM) {
```
polosanya, line 14: I think it's better to check for a long term, then for medium term and then just return basic price without any discounts

polosanya (approved):
Well done!

PR #4025, opened 2024-03-05
Solution

Allanpt:


jv-aquino (changes requested):
You are improving a lot, but there's some things that could be improved in your solution here

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40;
+
+  if (days < 3) {
+    return rentalCost * days;
+  } else if (days < 7) {
```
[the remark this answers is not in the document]
Allanpt: And What about numbers into 'return'? Like '-20' and '-50'? Is it ok, or I have to change this hard code too?

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+  const RENTALCOST = 40 * days;
+  const SHORT_DAYS = 3;
+  const LONG_DAYS = 7;
+
+  if (days < SHORT_DAYS) {
+    return RENTALCOST;
+  }
+
+  if (days < LONG_DAYS) {
+    return RENTALCOST - 20;
```
jv-aquino, line 16: the discounts could also be added as consts, if you may change this, it would be an even cleaner code and more modifiable

jv-aquino (approved):
Good job! There's a single thing missing that I ask you to implement afterwards, but for now good job! (also nice variable naming!)

src/calculateRentalCost.js
Allanpt: Ok ill change this, thx bro

PR #4027, opened 2024-03-05
Add solution

olafchuszno:


yevhenii-pyl (changes requested):
Solid! 

Let's just extract these constants from the function:  
```
LONG_TERM
LONG_TERM_DISCOUNT
MID_TERM
MID_TERM_DISCOUNT
PRICE_PER_DAY
```

You don't want them to be function scoped, since you got the idea with naming here, you probably want to simulate the environmental variables behaviour here.

src/calculateRentalCost.js
```diff
@@ -3,8 +3,25 @@
...
+
+const LONG_TERM = 7;
+const LONG_TERM_DISCOUNT = 50;
+const MID_TERM = 3;
+const MID_TERM_DISCOUNT = 20;
+const PRICE_PER_DAY = 40;
+
 function calculateRentalCost(days) {
-  // write code here
+  const BASE_PRICE = days * PRICE_PER_DAY;
```
yevhenii-pyl, line 14: Ah, and just overlooked it here: this variable should be just camelCased. 
It's just a regular function scoped variable, we should follow the standard naming convention. 

What's the main idea behind uppercased variables here:
with UPPER_CASED variables you are just mimicking the behaviour we have with env variables that are stored at a very special .env file and consumed within the components like

`const { API_GATEWAY_URL } = process.env`

yevhenii-pyl (approved):
Good job! Just pay attention to the variables naming convention, probably you also just overlooked it in process.

PR #4029, opened 2024-03-05
add calculateRentalCost function

metrakonon:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const baseCost = 40;
```
mrkovaliv, line 7: rename all your constants to uper snake case, it's a good practice to use this naming for constants
`baseCost -> BASE_COST`

mrkovaliv (changes requested):
Nice work 👍  but there are some improvements

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+  const shortTermDiscount = 20;
+  const longTermDiscount = 50;
+  const shortTermThreshold = 3;
+  const longTermThreshold = 7;
+
+  if (days < shortTermThreshold) {
```
mrkovaliv, line 13: you can skip one condition

```
  if (days < shortTermThreshold) {
    return baseCost * days;
  }

  if (days >= longTermThreshold) {
    return baseCost * days - longTermDiscount;
  }
  
  return baseCost * days - shortTermDiscount;
```

mrkovaliv (approved):
Looks good 👍 👍 👍

PR #4034, opened 2024-03-08
Develop

AvramenkoMarina:
1. Make names descriptive, so they explain the purpose of variables and functions.
2. Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.
3. Prefer `const` over `let` where possible, to avoid unintentional changes.
4. Prefer `if` with `return` over `if else` to simplify later conditions. 

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,6 +21,6 @@ calculateRentalCost(7); // 230
 ```
 
 ## Instructions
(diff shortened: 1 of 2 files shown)


anastasiiavorobiova (approved):
Awesome work!

PR #4046, opened 2024-03-12
solution

aMtiIV:


yevhenii-pyl (changes requested):
From the code perspective it works just fine! 
But let's adjust it from code style perspective:

1. Let's avoid 'magic numbers' by defining simulated env varibles outside the function scope, like:
```
const DISCOUNT_MID_TERM = 3;
const DISCOUNT_MID_VALUE = 20;
```

2. Suppose, there will be other options for discount - this may make else-if syntax a bit nasty. What if we use switch statement here?

BogdanMaliuta (approved):
Works good, but why your variables in upper case?

PR #4051, opened 2024-03-14
add task solution

DaniilManakov:


lerastarynets (approved):
You're on the right track!

PR #4058, opened 2024-03-17
add solution

ihorromaniuk:


nazarmatsevych (changes requested):
Switch looks redundant here, use if instead and move 'magic numbers' like '20', '40' and so on to variables 
https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md

nazarmatsevych (approved):
👍

PR #4062, opened 2024-03-17
Add Javascript First Solution :)

vinasxz:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
+
+  if (days >= 7) {
```
jv-aquino, line 9: The conditional days should be stored in a way to facilitate changes - also remember the naming guidelines (check 1) when you do this
☝️
[checklist 2](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+    totalCost -= 50;
```
jv-aquino, line 10: you should also store the discount in a better manner

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+  } else if (days >= 3) {
+    totalCost -= 20;
```
jv-aquino, line 12: Also, you **could/might** use a direct return after the conditionals

[and remember checklist 4 and 5](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

jv-aquino (changes requested):
Good code but we can make it better!

src/calculateRentalCost.js
vinasxz: Wow, i'm guessing i dind't saw that. Fixed already, following the checklist now.

PR #4068, opened 2024-03-18
Solution

aPetrashyshyn:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentalCost = 40 * days;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 7;
```
TarasHoliuk, line 10: wrong value here

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+
+  if (days >= LONG_TERM) {
+    return rentalCost - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
+    return rentalCost - SHORT_TERM_DISCOUNT;
+  }
```
TarasHoliuk, line 17:
```suggestion
  if (days >= LONG_TERM) {
    return rentalCost - LONG_TERM_DISCOUNT;
  } 
  
  if (days >= SHORT_TERM) {
    return rentalCost - SHORT_TERM_DISCOUNT;
  }
```
with `return` inside `if` you don't need `else`. Such :arrow_up: code is easier to read

TarasHoliuk (changes requested):

nazarmatsevych (approved):
👍

PR #4070, opened 2024-03-19
Solution

HigorPedroso:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
  */
 function calculateRentalCost(days) {
   // write code here
```
jv-aquino, line 7: i think you forgot to remove this 😅

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const basePrice = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return basePrice - LONG_TERM_DISCOUNT;
+  }
+
```
jv-aquino, line 19: generally keep the ifs together if they have the same idea, so no empty lines necessary here

jv-aquino (approved):
Awesome code!!!!!! You got everything right at the first try, that's a really good sign Higor! Keep up your good work!

(i left two really minor commentary if you may want to change but it's nothing that affects the beauty of this code here!)

PR #4075, opened 2024-03-19
Solution

ropanefarias:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,33 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const RENT_PER_DAY = 40;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
```
jv-aquino, line 11: nice variables naming!

jv-aquino (changes requested):
Your code is organized and you are using some of the best practices, but you are doing two mistakes:
1. You are using too much ifs: you could just use 2 in fact
2. You forgot some of the [checklist (4 and 5)](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

pedro-ruas (approved):
Nice work, Rodrigo!
If I may give you an advice, we **usually** use uppercase variable declarations for variables declared **outside** of function scope.
This is because  a variable, even a *const*, declared insede a function, will exist only in that function and will be generated and deleted whenever that function starts and ends.

Now, a **const** declared *outside* a function will be created and stored once and then only be accessed whenever the function runs.

In that way, we usually use the uppercase notation for **final const variables** declared *outside* a function. Got it?

PR #4083, opened 2024-03-21
https://github.com/AlexRomant/js_task-transportation-on-vacation.git

AlexRomant:


Ivan-Shvorak (changes requested):
You need to add only required function, please don't touch tests -
"Implement calculateRentalCost function that returns the total amount for different number of days."

AlexRomant (commented):
was changed the information in functions

AlexRomant (commented):
Plese check my Homework

PR #4091, opened 2024-03-24
Solution

karpinska006:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
+  const MIDDLE_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const MIDDLE_TERM = 3;
+  let price = days * REGULAR_COST;
+
+  if (days >= LONG_TERM) {
+    price -= LONG_TERM_DISCOUNT;
+  } else if (days >= MIDDLE_TERM) {
+    price -= MIDDLE_TERM_DISCOUNT;
+  }
```
TarasHoliuk, line 18: In such cases, you can immediately return (and avoid `else if` - it's easier to read code without it):

```suggestion
  if (days >= LONG_TERM) {
    return price - LONG_TERM_DISCOUNT;
  }
  
  if (days >= MIDDLE_TERM) {
    return price - MIDDLE_TERM_DISCOUNT;
  }
```

TarasHoliuk (changes requested):
GJ :+1: 
Added a small suggestion how to make code easier to read :slightly_smiling_face:

etojeDenys (approved):

PR #4097, opened 2024-03-25
Solution

yuliaPel:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DEY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  let totalAmount = PRICE_PER_DEY * days;
```
polosanya, line 12: Make it const instead of let. Also, the name is a bit confusing as it's not a totalAmount. Discounts are not included

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
+
+  if (days >= LONG_TERM) {
+    return totalAmount - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM) {
+    return (totalAmount = totalAmount - SHORT_TERM_DISCOUNT);
```
polosanya, line 19: What's the purpose of reassigning the variable? Just return the result

polosanya (changes requested):
Almost done!

polosanya (approved):
Well done! 👍

PR #4099, opened 2024-03-25
Solution

odunchyk:


yurii-filiushkin (approved):
GREAT JOB!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM_DISCOUNT = 20;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const RentalCost = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return RentalCost - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
```
undeletable, line 16: We don't need `else if` here. If first conditional on line 14 is true, this code would never be executed because of return. Refer to item 5 of task checklist.

undeletable (changes requested):
The solution is fine, but needs a small change according to the checklist. See my comment to the code.

mshepel (approved):
Well done

undeletable (approved):
Well done!

PR #4101, opened 2024-03-25
add task solution

MarchAlbion:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const LONG_TERM_DISCOUNT = 50;
```
yanamomot, line 8: Fine! But, you can do this task more easily :)

yanamomot (approved):

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
...
+  const THREE_DAYS_DISCOUNT = 3;
```
mrkovaliv, line 9: move on all constants outside the function

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+  const SEVEN_DAYS_DISCOUNT = 7;
+  const PRICE_FOR_ONE_DAY = 40;
+
+  let TOTAL_PRICE = PRICE_FOR_ONE_DAY * days;
```
mrkovaliv, line 14: use const -> let
just `totalPrice` because TOTAL_PRICE isn't constant

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
...
+
+  if (days >= SEVEN_DAYS_DISCOUNT) {
+    TOTAL_PRICE -= LONG_TERM_DISCOUNT;
```
mrkovaliv, line 17: just return value

`return totalPrice -  LONG_TERM_DISCOUNT;
`

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
...
+  }
+
+  if (days >= THREE_DAYS_DISCOUNT && days < SEVEN_DAYS_DISCOUNT) {
+    TOTAL_PRICE -= SHORT_TERM_DISCOUNT;
```
mrkovaliv, line 21: the same

mrkovaliv (changes requested):
Nice work 👍 👍 👍 

Added some improvements

DanilWeda (approved):
Good job. 
Approved. 
In the future - you can restore files ( about src/calculateRentalCost.test.js) so that they do not fall into the PR. You can check the git commands, but be careful!
Thanks)

mrkovaliv (approved):

PR #4103, opened 2024-03-25
solution js first task vs debug

Olek-91:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
+  const MIN_TERM = 3;
+  const MIN_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+  const renCost = DAY_PRICE * days;
+
+  if (days >= MIN_TERM && days < LONG_TERM) {
+    return renCost - MIN_TERM_DISCOUNT;
+  }
+
+  if (days >= 7) {
```
iAlfreX, line 18: Don't use literals in expressions (HARDCODE).

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const DAY_PRICE = 40;
+  const LONG_TERM = 7;
+  const MIN_TERM = 3;
+  const MIN_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+  const renCost = DAY_PRICE * days;
+
+  if (days >= MIN_TERM && days < LONG_TERM) {
```
iAlfreX, line 14: You can put this condition after the second one. And change it to just days >= MIN_TERM

iAlfreX (changes requested):
Good job! Just make some corrections

src/calculateRentalCost.js
Olek-91: do it, thanks)))

src/calculateRentalCost.js
Olek-91: fixed))))

src/calculateRentalCost.js
Olek-91: As you can see, this was done on purpose))

iAlfreX (approved):
Good job :)

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const DAY_PRICE = 40;
```
mrkovaliv, line 7: move on all constants outside the function

mrkovaliv (approved):
Great job 👍 👍 👍

PR #4104, opened 2024-03-25
create calculateRentalCost func

iAlfreX:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+  const BASE_PRICE = PRICE_PER_DAY * days;
```
Olek-91, line 12: [checklist.md] 2. Do not use literals in HARDCODE expressions. Instead, use constants that explain these values. Yes, but for me it is better to use a 'camelCase' for the result of the expressions, since it is not just a value. In general, the function reads all clear. Good job!!!

Olek-91 (approved):
GOOD JOB!!!

src/calculateRentalCost.js
iAlfreX: It makes sense. I'm agree with you, thanks!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
```
mrkovaliv, line 7: move on all constants outside the function

src/calculateRentalCost.js
mrkovaliv, line 12: better `BASE_PRICE` -> `basePrice`, BASE_PRICE isn't constant

mrkovaliv (approved):
Nice work 👍 👍 👍

PR #4117, opened 2024-03-26
Solution

yarynaburenko:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let price = days * 40;
```
l4st1m0za, line 7:
```suggestion
  let price = days * 40;
```

40 is magic number here

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+
+  if (days >= 7) {
```
l4st1m0za, line 9: 7 is magic number here

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+    price -= 50;
```
l4st1m0za, line 10: 50 is magic number here

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+  } else if (days >= 3) {
```
l4st1m0za, line 11: You need to avoid all magic numbers!

ghost (changes requested):
Good job, but improvements needed

Esceype (approved):
Well done!

PR #4123, opened 2024-03-28
Solution

luyx0:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
```
mrkovaliv, line 7: use const here

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
...
+
+  if (days >= 7) {
+    totalCost -= 50;
```
mrkovaliv, line 10: just return totalCost - 50;

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
+
+  if (days >= 7) {
```
mrkovaliv, line 9: move on all values to constants and move on them outside the function:
example
```
  const SHORT_TERM = 3;
```

mrkovaliv (changes requested):
Nice work 👍  but there are some improvements)))

src/calculateRentalCost.js
polosanya: Not fixed. Create variables for all 'magic numbers' (40, 7, 3)

src/calculateRentalCost.js
```diff
@@ -3,8 +3,19 @@
...
+const SHORT_TERM_DISCOUNT = 20;
+
 function calculateRentalCost(days) {
-  // write code here
+  const totalCost = days * 40;
+
+  if (days >= 7) {
+    return totalCost - WEEKLY_DISCOUNT;
+  } else if (days >= 3) {
+    return totalCost - SHORT_TERM_DISCOUNT;
```
polosanya, line 15:
```suggestion
    return totalCost - WEEKLY_DISCOUNT;
  }
  
  if (days >= 3) {
    return totalCost - SHORT_TERM_DISCOUNT;
```
`else` is redundant here as the code will not go further after `return` anyway

polosanya (changes requested):
Almost done!

src/calculateRentalCost.js
```diff
@@ -3,8 +3,22 @@
...
  */
+const DAILY_RATE = 40;
+const WEEKLY_DISCOUNT = 50;
+const SHORT_TERM_DISCOUNT = 20;
+const WEEK = 7;
+const THREE_DAYS = 3;
+
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * DAILY_RATE;
```
mrkovaliv, line 13: use `const` here

 const totalCost = days * DAILY_RATE;

mrkovaliv (approved):
Nice work 👍 👍 👍

PR #4136, opened 2024-04-01
Solution

Ganger2525:


src/calculateRentalCost.js
```diff
@@ -13,9 +13,9 @@ function calculateRentalCost(days) {
   const basePrice = days * PRICE_PER_DAY;
 
   if (days >= LONG_TERM) {
-    return basePrice - LONG_TERM_DISCOUNT;
+   return basePrice - LONG_TERM_DISCOUNT;
   } else if (days >= SHORT_TERM) {
```
Mian3450, line 17: [Check list]: DON'T add `else` after `if` with the `return` - the code after it won't be executed anyway.

BAD EXAMPLE:

if (condition) {
  return x;
} else {
  return y;
}

GOOD EXAMPLE:

if (condition) {
  return x;
}

return y;

Mian3450 (changes requested):
GJ, but needs some fixes

natalliiaa (approved):
LGTM!

PR #4174, opened 2024-04-09
implemented task

vadymhrnk:


src/calculateRentalCost.js
nazar-medushevskyi, line 1: GJ🔥 You're using constants for various values which is good. Consider adding descriptive names for these constants to improve readability

nazar-medushevskyi (approved):
GJ🔥

PR #4188, opened 2024-04-14
Solution

ekaterinapopova1382:


Khalandr (changes requested):
Ekaterina, You have a good approach, but please create additional variables which will contain discounts ad rewrite if-else statement using those variables.

Khalandr (approved):
Well done

PR #4189, opened 2024-04-15
Solution

Nivon35:


Khalandr (changes requested):
This is a good try, but please declare two variables for short term and long term discounts and rewrite if-else statement with them

Khalandr (approved):
GJ

PR #4190, opened 2024-04-15
Solution

Steiniak:


Khalandr (changes requested):
This is a good approach, but please declare additional variables for short term and long term discounts and rewrite if-else statement with them.

Steiniak (commented):
fixed

Khalandr (approved):
GJ

PR #4191, opened 2024-04-15
Solution

MarcinG0:


Khalandr (changes requested):
This is a good try, but please create two variables for short term and long term discounts instead of reassigning discount variables few times.

PR #4193, opened 2024-04-15
Solution

Kryspinox:


Khalandr (changes requested):
Damian, this is a good solution, but kindly ask You to refactor Your variables, using camelCase naming convention. (the same as basePrice)

Khalandr (approved):
GJ

PR #4195, opened 2024-04-15
Solution

ElaOz:


Khalandr (changes requested):
While technically Your code works fine and provide correct solution , there is a space for some good practices. First of all, You have some constantas, like day price or discounts. Please, declare variables, assign those values and rewrite Your if-else statement using variables.

ElaOz:
Please review my changes.

Khalandr (approved):
GJ

PR #4199, opened 2024-04-16
Solution

MariaSzynaka:


Khalandr (changes requested):
please declare two variables for short term and long term discounts and rewrite if-else statement

PR #4205, opened 2024-04-17
Solution

BagietaCyganeta:


BogdanMaliuta (approved):
Solution looks good, next time you can use return in your if for remove else

PR #4238, opened 2024-04-30
add task solution

jacube1:


BogdanMaliuta (approved):
Nice work!
You can optimize your code by removing if/else, and using return in if

PR #4240, opened 2024-05-01
add solution

santana1337:


nazarmatsevych (changes requested):
Good start but proofread your code with [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

PR #4255, opened 2024-05-08
Solution

WGoliasz:


yevhenii-pyl (changes requested):
Good job! 👍 

Just a small note: since you are mimicking environmental variables with uppercase naming( which is absolutely right here!), you should declare these variables outside the function scope. 

Later, working with real .env file, you will have to import them: 
`const { CLIENT_PHONE, PASSWORD, TEST_API_URL, HOST, SERVICE_PORT } = process.env;`

Also, 'AMOUNT' variable is not meant to be taken from the .env, it's a dynamic variable created inside your `calculateRentalCost` function.

So let's adjust two things:

1. Declare variables that are mimicking env vars `AMOUNT_FOR_DAY... BIG_DISCOUNT` at the global scope (outside the function);
2. Use correct naming notation for the `AMOUNT` variable.

WGoliasz:
Is it okay now? :)

yevhenii-pyl (approved):
Aaaand you got it!
Now it's 100% accurate 🥇

PR #4277, opened 2024-05-16
Solution

Kasprukov:


RuslanHnatyshyn (changes requested):
Good job)
Fix your code.

Kasprukov:
> Good job) Fix your code.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
+  const carCostPerDay = 40;
+  const mediumRentTermDiscount = 20;
+  const longRentTermDiscount = 50;
+  const price = carCostPerDay * days;
+
+  if (days < 3) {
+    return price;
+  }
+
+  if (days < 7) {
```
dinaminahi, line 16: Let's add 3 and 7 to constants and  just a heads-up that it's a good practice in JavaScript to use SNAKE_CASE (all uppercase) for constant variables.
image

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const carCostPerDay = 40;
+  const mediumRentTermDiscount = 20;
+  const longRentTermDiscount = 50;
+  const price = carCostPerDay * days;
+
+  if (days < 3) {
+    return price;
+  }
```
dinaminahi, line 14: This is default price that can be returned in the end without check if other two weren't true
So firstly let's check if `days >= 7`
then if `days >= 3`
and then return default

dinaminahi (changes requested):
Good job 👍 
Let's improve your code a little bit

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const longTermDays = 7;
+  const shortTermDays = 3;
+  const carCostPerDay = 40;
+  const mediumRentTermDiscount = 20;
+  const longRentTermDiscount = 50;
```
roman-mirzoian, line 11: Here it is better to use contacts:

image

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
+  const price = carCostPerDay * days;
```
roman-mirzoian, line 12: I also recommend using a naming that will be more understandable.
```suggestion
  const basePrice = carCostPerDay * days;
```

src/calculateRentalCost.js
roman-mirzoian, line 11: Here you can simplify the naming a little.

roman-mirzoian (changes requested):
Well done, but pay attention to this comments.

src/calculateRentalCost.js
Kasprukov: @roman-mirzoian 
  const longTerm = 7;
  const shortTerm = 3;
  const costPerDay = 40;
  const mediumDiscount = 20;
  const longDiscount = 50;
  const basePrice = costPerDay * days;
  
that is better?)

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const longTerm = 7;
+  const shortTerm = 3;
```
roman-mirzoian, line 8: Example:
```suggestion
  const LONG_TERM = 7;
  const SHORT_TERM = 3;
```

roman-mirzoian (changes requested):
It's better, but you probably didn't understand the previous comments. It was about the fact that constants are written with all capital letters.

Pay attention to the comment, I gave an example there.

roman-mirzoian (approved):
Well done!

PR #4280, opened 2024-05-16
Add solution

RuslanHnatyshyn:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const payPerDay = 40;
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+
+  if (days < 3) {
```
FaiHamid, line 12: 2. [checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.

This refers to the number of days for the discount.

src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const payPerDay = 40;
```
FaiHamid, line 8: Also for constants that we use  for explanation values a different syntax will be used, where all letters are capitalized, and instead of a space - an underscore `_` .

FaiHamid (changes requested):

FaiHamid (changes requested):
Good job)
Please, fix your code.

FaiHamid (approved):

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
  */
 function calculateRentalCost(days) {
   // write code here
```
roman-mirzoian, line 7:
```suggestion
```

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
...
+  const PAY_PER_DAY = 40;
+  const BIG_DISCOUNT = 50;
+  const SMALL_DISCOUNT = 20;
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+
+  if (days < SHORT_TERM) {
+    return days * PAY_PER_DAY;
```
roman-mirzoian, line 15: ``days * PAY_PER_DAY;``
This logic is repeated and you can transfer it to a variable so as not to duplicate the code.

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const PAY_PER_DAY = 40;
+  const BIG_DISCOUNT = 50;
+  const SMALL_DISCOUNT = 20;
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+
+  if (days < SHORT_TERM) {
```
roman-mirzoian, line 14: You can also make the code more beautiful.
Remove this condition and return the standard price after the other two conditions.

roman-mirzoian (changes requested):
Nice work, but let's improve your code.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PAY_PER_DAY = 40;
+  const SUM_PER_DAY = days * PAY_PER_DAY;
```
roman-mirzoian, line 8: It is better not to use a constant here: Constants are for values that remain unchanged throughout the execution of the program. Because SUM_PER_DAY depends on other variables, its value can vary depending on the context.

roman-mirzoian (changes requested):
Well done, but pay attention to this comment.

roman-mirzoian (approved):
Good work, but pay attention to the comment that was left for you.
Also tip for future - static constants can be kept in a file outside of a function.

Also, be careful when naming this constant using camelCase should start with a lowercase letter:
your version:
`const SumPerDay = days * payPerDay;`
correct version:
`const sumPerDay = days * payPerDay;`

RuslanHnatyshyn:
Thank you for your comments.

PR #4283, opened 2024-05-16
add task solution

V-Shut:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
+  const longTerm = 7;
+  const shortTermDisc = 20;
+  const longTermDisc = 50;
+  const noDisc = price * days;
+
+  if (days >= shortTerm && days < longTerm) {
+    return noDisc - shortTermDisc;
+  } else if (days >= longTerm) {
+    return noDisc - longTermDisc;
+  } else {
```
demkoveronika, line 18: 5. [Cheklist] DON'T add else after if with return - the code after it won't be executed anyway.

demkoveronika (changes requested):
Good job! 
Please fix your code

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const shortTerm = 3;
+  const longTerm = 7;
+  const shortTermDisc = 20;
+  const longTermDisc = 50;
+  const noDisc = price * days;
+
+  if (days >= shortTerm && days < longTerm) {
```
dinaminahi, line 14: Let's firstly check if days >= longTerm and in second if you will just need to check if days >= shortTerm

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const price = 40;
+  const shortTerm = 3;
+  const longTerm = 7;
+  const shortTermDisc = 20;
+  const longTermDisc = 50;
+  const noDisc = price * days;
```
dinaminahi, line 12: we don't usually use use shortcuts for names, so let's rename it to something like `totalCost`
And just a heads-up that it's a good practice in JavaScript to use SNAKE_CASE (all uppercase) for constant variables.
image

dinaminahi (changes requested):
Good job 👍  Let's improve it a little bit

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE = 40;
+  const SHORT_TERM = 3;
+  const longTerm = 7;
```
roman-mirzoian, line 9: Why don't you use a constant here.
```suggestion
  const LONG_TERM = 7;
```

roman-mirzoian (approved):
Well done, and please pay attention to this comment.

PR #4286, opened 2024-05-16
Solution

OSpyrydonov:


Stasandrushchak (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const totalPrice = days * PRICE_PER_DAY;
```
roman-mirzoian, line 13: I think it is better not to call the ``totalPrice`` but the ``basePrice``

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
roman-mirzoian, line 11: Future tip - static constants can be stored in a file outside the function:
Because if we call the function multiple times, these constants will be created multiple times in the middle of the function, and if you move them outside the function, they will be created once regardless of the number of function calls.

roman-mirzoian (approved):
Well done!

PR #4287, opened 2024-05-16
Solution

LitviVit:


OSpyrydonov (approved):
Good work!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
roman-mirzoian, line 11: Future tip static constants can be kept in a file outside of a function:

roman-mirzoian (approved):
Well done!

PR #4288, opened 2024-05-16
Solution

DSB-18:


StsKushnir (approved):
Good job! Excellent!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const DAY_TERM = 40;
```
roman-mirzoian, line 7: Here the naming is not clear, it is better to use the following:
```suggestion
  const PRICE_PER_DAY = 40;
```

roman-mirzoian (approved):
Well done, and please pay attention to this comment.

PR #4290, opened 2024-05-16
Solution

OleksiiKnihin:
Solution js_task-transportation-on-vacation

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -3,8 +3,22 @@
  *
  * @return {number}
  */
-function calculateRentalCost(days) {
-  // write code here
+function calculateRentalCost(numberOfDays) {
+  const basePrice = 40;
+  const MIDDLE_TERM = 3;
+  const LONG_TERM = 7;
+  const MIDDLE_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (numberOfDays >= LONG_TERM) {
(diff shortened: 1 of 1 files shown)


DSB-18 (approved):
Well done!

src/calculateRentalCost.js
```diff
@@ -3,8 +3,22 @@
  *
  * @return {number}
  */
-function calculateRentalCost(days) {
-  // write code here
+function calculateRentalCost(numberOfDays) {
+  const basePrice = 40;
```
roman-mirzoian, line 7: Сonstant can also be used here

src/calculateRentalCost.js
```diff
@@ -3,8 +3,22 @@
...
+  const MIDDLE_TERM = 3;
+  const LONG_TERM = 7;
+  const MIDDLE_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (numberOfDays >= LONG_TERM) {
+    return numberOfDays * basePrice - LONG_TERM_DISCOUNT;
```
roman-mirzoian, line 14: expression ``numberOfDays * basePrice`` can be used as a variable to avoid duplicating code

roman-mirzoian (changes requested):
Good job but let's improve it.

roman-mirzoian (approved):
Well done!

Also tip for future - static constants can be kept in a file outside of a function.

PR #4293, opened 2024-05-17
Solution

mhryvko:


src/calculateRentalCost.js
```diff
@@ -9,7 +9,7 @@ function calculateRentalCost(days) {
   const LONG_TERM_DISCOUNT = 50;
   const SHORT_TERM = 3;
   const SHORT_TERM_DISCOUNT = 20;
-  let basePrice = days * PRICE_PER_DAY;
+  const basePrice = days * PRICE_PER_DAY;
```
OleksiiKnihin, line 12: You need to use `let` because this variable is changes.

OleksiiKnihin (approved):
Good job!
But you need need to fix a small mistake

P.S. Yes, I saw. In the first variant of your solution you did right) watch the video "Code Review on Github" from 2 min 42 sec.
P.S.S. I hope, I'm right)))

roman-mirzoian (approved):
Well done! 
Tip for future - static constants can be kept in a file outside of a function.

PR #4298, opened 2024-05-18
Solution

sa4ok-1:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyCost = 40;
+  let totalCost = days * dailyCost;
+
+  if (days >= 7) {
```
ikocherovets, line 10: 2.[CHECKLIST] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

BAD EXAMPLE:
if (numberOfDays >= 7) {
  return basePrice - 50;
}

GOOD EXAMPLE:
const LONG_TERM = 7;


The same problem with the line number 12.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
...
+    totalCost = totalCost - 50;
+  } else if (days >= 3) {
```
ikocherovets, line 12: 4. [CHECKLIST] Prefer if with return over if else to simplify later conditions.

Same with the line number 11, use `return` after `if` inside a function.

ikocherovets (changes requested):
Good job mate!
There are some little adjustments, so please fix them inside your code :)

roman-mirzoian (approved):
Well done!

PR #4301, opened 2024-05-19
Solution

Kateryna99:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
...
-  // write code here
+  const longTermDiscountDays = 7;
+  const shortTermDiscountDays = 3;
+  const carPricePerDay = 40;
+  const carPricePerRent = days * carPricePerDay;
+
+  if (days >= longTermDiscountDays) {
+    return carPricePerRent - 50;
+  } else if (days >= shortTermDiscountDays && days < longTermDiscountDays) {
+    return carPricePerRent - 20;
```
polosanya, line 15:
```suggestion
  if (days >= longTermDiscountDays) {
    return carPricePerRent - 50;
  }
  
  if (days >= shortTermDiscountDays && days < longTermDiscountDays) {
    return carPricePerRent - 20;
```
`else` is redundant here as the code will not go further after return anyway

src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const longTermDiscountDays = 7;
+  const shortTermDiscountDays = 3;
+  const carPricePerDay = 40;
+  const carPricePerRent = days * carPricePerDay;
+
+  if (days >= longTermDiscountDays) {
+    return carPricePerRent - 50;
```
polosanya, line 13: Let's create variables for prices too

polosanya (changes requested):
Almost done!

PR #4303, opened 2024-05-20
add task solution

pushyshyn:


roman-mirzoian (approved):
Well done! Also tip for future - static constants can be kept in a file outside of a function.

PR #4311, opened 2024-05-25
Solution

YaroslavKolesnyk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyCost = 40;
+  const manyDays = 7;
+  const fewDays = 3;
+
+  if (days >= manyDays) {
+    return dailyCost * days - 50;
```
polosanya, line 12: Let's create a variable for `dailyCost * days` to avoid duplication 
Also, variables for `50` and `20` are required

polosanya (changes requested):
Nice! Let's improve your code a bit)

PR #4317, opened 2024-05-27
Add vocation cost calculator

GoVolodya:
Created a function to calculate the cost of vocation car rent cost.

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -4,22 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  const dailyRentalRate = 40;
-  const smallDiscount = 20;
-  const bigDiscount = 50;
-  const longTermRental = 7;
-  const shortTermRental = 3;
-  const vacationRentalCost = days * dailyRentalRate;
+  const DAILY_RENTAL_RATE = 40;
+  const SMALL_DISCOUNT = 20;
+  const BIG_DISCOUNT = 50;
+  const LONG_TERM_RENTAL = 7;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -5,6 +5,20 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const cost = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
```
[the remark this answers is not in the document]
etojeDenys: image
better to fix it

etojeDenys (changes requested):
rewrite your code following this checklist

https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md

GoVolodya:
Sorry, I have missed the check list. Will know it's there for the future.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRentalRate = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
+  const longTermRental = 7;
+  const shortTermRental = 3;
```
etojeDenys, line 11: name these constants in uppercase

etojeDenys (changes requested):

src/calculateRentalCost.js
GoVolodya: Yes, that is a good idea, so when we have more code we can see the difference between variables and constants.

maxim2310 (approved):
Well done!!!

PR #4318, opened 2024-05-28
Solution2

YaroslavKolesnyk:


Pashahu1:
You can change your constant writing on:

const DAILY_COST = 40;
const MANY_DAYS = 7;
const FEW_DAYS = 3;
const MANY_DAYS_DISCOUNT = 50;
const FEW_DAYS_DISCOUNT = 20;

this be a good solution 🥇

PR #4320, opened 2024-05-29
add calculateRentalCost function in solution

Daniel-spec-dell:
Test passed.


lerastarynets (approved):
Keep up the great work!

PR #4322, opened 2024-05-30
My solution

karaniewskiruslan:
My solution to task

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -5,6 +5,10 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const rentCost = 40;
+  const rentOff = days < 3 ? 0 : days >= 3 && days >= 7 ? 50 : 20;
+
+  return rentCost * days - rentOff;
 }
 
 module.exports = calculateRentalCost;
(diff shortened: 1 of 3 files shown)


src/calculateRentalCost.js
```diff
@@ -5,6 +5,10 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const rentCost = 40;
```
nataliaklonowska, line 8: use UPPERCASE for constant values (rentCost=> RENT_COST)

src/calculateRentalCost.js
```diff
@@ -5,6 +5,10 @@
...
+  const rentOff = days < 3 ? 0 : days >= 3 && days >= 7 ? 50 : 20;
```
nataliaklonowska, line 9: [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md): "Don't use literals in expressions (HARDCODE). Instead use constants explaining this values"

nataliaklonowska (changes requested):
To make your code more reusable and improve readability you should change numbers to constants that explain it's purpose

BogdanMaliuta (approved):
Solution looks good, but dont forget remove comments from your code

PR #4334, opened 2024-06-04
[Feature] added rental cost calculation

Arthur-Tolkachov:
## Description
Added rental cost calculation function


vadiimvooo (approved):
Well done

PR #4339, opened 2024-06-06
Kate Hubko

KaterynaHubko:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,19 @@
...
+  if (days === 6) {
+    return 220;
+  }
+
+  if (days === 3) {
+    return 100;
+  }
+
+  return 80;
 }
```
4ksandrey, line 21: your code will not work if prices will changes, please revise video for this task you have a solution) use it

4ksandrey (changes requested):
check the comment

Julia14012000 (approved):
Good job!

PR #4370, opened 2024-06-18
Solution

Voronaolga:


Sh1r09 (approved):
You need to change your returns because they work only with specific values.

annakachurets (approved):
gj

PR #4372, opened 2024-06-18
Solution

MarynaZhukovska:


yurayagu:
Good job)

AdolfWhiteCat (approved):
good job!

Huskov-Anton (changes requested):
you should create variables and assign values ​​to them and then use these variables in your code - not hardcoded values (check solution on platform )

PR #4379, opened 2024-06-18
task completed using switch cases

ketinng:


lerastarynets (approved):
You've got this, keep pushing forward!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,14 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  switch (true) {
```
smachylooo, line 7: Better to use `if` because it will exist only when you need. Now you will always do `switch`, which could play with you bad joke in larger project and also it's overload memory.

smachylooo (changes requested):

PR #4385, opened 2024-06-20
Solution

Diegocgamarra:


gitlherme (changes requested):
Hello, @Diegocgamarra!

First of all, great job!

The AI consider some points that make sense. Can you please check it?

I think that you can avoid to use else if statement. You can create just 2 if statements and use another return as default. Can you please check if it make sense?

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
+  const longTermDiscount = 50;
+  const subtotal = basePrice * days;
+
+  if (days < shortTerm) {
+    return subtotal;
+  } else if (days >= shortTerm && days < longTerm) {
+    return subtotal - shortTermDiscount;
+  } else if (days >= longTerm) {
+    return subtotal - longTermDiscount;
+  }
```
gitlherme, line 20: Here you can simplify your code to be more maintanable and readable if you use something like

```js

if (days < shortTerm) {
    return subtotal;
}

if (days >= shortTerm && days < longTerm) {
    return subtotal - shortTermDiscount;
}

return subtotal - longTermDiscount;

```

If your code enter in one of the ifs, the return of the if will be returned. Else, will be returned the default return, that is the subtotal - longTermDiscount. Make sense?

But your code is great! Good job! 🎉

gitlherme (approved):
Great job! Just a comment about the code that can be used in the next exercises. =)

PR #4408, opened 2024-07-01
Solution

DarynaRiabova:


Khalandr (changes requested):
While provided solution is technically correct - please try to avoid hardcoding any values. Please refactor Your code, so every value would be stored in a variable.

Khalandr (approved):
well done

PR #4410, opened 2024-07-01
Solution

YuliiaKosenchuk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,24 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const COST_PER_DAY = 40;
+  const TOTAL_COST = days * COST_PER_DAY;
```
etojeDenys, line 8:
```suggestion
  const totalCost = days * COST_PER_DAY;
```
it's not a constant

etojeDenys (approved):

src/calculateRentalCost.js
YuliiaKosenchuk: exactly! There may be a different number of days.

PR #4415, opened 2024-07-02
Solution

iyehorova:


NavI-Rap0 (approved):
good job!

src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
  *
  * @return {number}
  */
+const DAY_RENT = 40;
```
ArturErmolenko, line 6:
```suggestion
const dayRent = 40;
```

src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
...
+const LONG_TERM = 7;
```
ArturErmolenko, line 7:
```suggestion
const longTerm = 7;
```
use CamelCase for const

ArturErmolenko (approved):
GJ 👍

PR #4416, opened 2024-07-02
Solution

igor-romasiuk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const basePrice = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return basePrice - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
```
AndGrachov, line 17: 4. [checklist] Prefer if with return over if else to simplify later conditions.
5. [checklist] DON'T add else after if with return - the code after it won't be executed anyway.

Better just to use another `if`.

AndGrachov (changes requested):
Great job!
But please fix your code.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```
Moroz-Dmytro, line 11: Would be better to move these variables outside the function

Moroz-Dmytro (approved):
Looks good 👍

PR #4417, opened 2024-07-02
calcRentCost

NavI-Rap0:


src/calculateRentalCost.js
```diff
@@ -1,10 +1,29 @@
+'use strict';
+
 /**
  * @param {number} days
  *
  * @return {number}
  */
 function calculateRentalCost(days) {
   // write code here
```
piyopiyo0, line 9: not critical, but you can remove the comment here :)

src/calculateRentalCost.js
```diff
@@ -1,10 +1,29 @@
...
+  const DAY_COST = 40;
+  const MANY_DAYS = 7;
+  const SOME_DAYS = 3;
+  const MANY_DAYS_DISCOUNT = 50;
+  const FEW_DAYS_DISCOUNT = 20;
+
+  const daysCost = DAY_COST * days;
```
piyopiyo0, line 16: I would suggest a more precise name for the final variable (like totalDaysCost) (but not really sure anyway :D)

piyopiyo0 (approved):
Good job! 👍

mrkovaliv (approved):
👍 👍 👍

PR #4428, opened 2024-07-02
Finish

NIKaragu:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  let amount = 0;
+
+  for (let i = 0; i <= days - 1; i++) {
+    amount += 40;
+  }
+
+  if (days >= 7) {
```
Teollan, line 14: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining these values.

for this and some other lines

Teollan (changes requested):
Great job!

Fix some minor issues, to make it perfect)

src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  let amount = 0;
+
+  for (let i = 0; i <= days - 1; i++) {
```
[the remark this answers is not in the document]
NIKaragu: Yea, you're right. But about `const amount = days * amountPerDay;` you are not, 'cause we can't (accordingly to rule of correct tone) determine modifiable variables as a constant.

Teollan (approved):
🔥

src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const PAY_PER_DAY = 40;
+  const DAYS_FOR_MAX_DISCOUNT = 7;
+  const DAYS_FOR_MIN_DISCOUNT = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const MIN_TERM_DISCOUNT = 20;
```
Moroz-Dmytro, line 12: Would be better to move them outside the function

src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
...
+
+  if (days >= DAYS_FOR_MAX_DISCOUNT) {
+    return PAY_PER_DAY * days - LONG_TERM_DISCOUNT;
```
Moroz-Dmytro, line 15: Create variable for `PAY_PER_DAY * days` and reuse in all places

Moroz-Dmytro (changes requested):
Good work 👍 
Small comments below

src/calculateRentalCost.js
NIKaragu: done ✔️

BudnikOleksii (approved):
👍

PR #4436, opened 2024-07-02
add task solution

Yurii-Yaremko:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,29 @@
...
+  let COUNT_MONEY = 0;
+  const PRICE_ONE_DAY = 40;
+  const TOTAL_AMOUNT = PRICE_ONE_DAY * days;
+  const DISCOUNT_THREE_DAYS = 20;
+  const DISCOUNT_SEVEN_DAYS = 50;
+  const THREE_DAY = 3;
+  const SEVEN_DAY = 7;
+
+  if (days < THREE_DAY) {
+    COUNT_MONEY = TOTAL_AMOUNT;
```
Mazurok001, line 17: Good job! But without an additional variable `CAUNT_MONEY` you can return `return` the result directly in the `if` block, also you can use it in all your code, like item 4.[checklist]. The code works in your case too.

Mazurok001 (approved):
Good job!

src/calculateRentalCost.js
BudnikOleksii: > Good job! But without an additional variable `CAUNT_MONEY` you can return `return` the result directly in the `if` block, also you can use it in all your code, like item 4.[checklist]. The code works in your case too.

Good point!
A few more suggestions:
- move constants such as `PRICE_ONE_DAY`, etc outside the function
- Variables should be in camelCase - `const totalAmount = PRICE_ONE_DAY * days;`

BudnikOleksii (changes requested):
Good work 👍 
Check the comments and fix in accordance

BudnikOleksii (approved):
👍

PR #4451, opened 2024-07-08
Solution

nataSokolnikova:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
  *
  * @return {number}
  */
+const minDaysForDiscount = 3;
+const maxDaysForDiscount = 7;
+const baseDiscountAmount = 20;
+const largeDiscountAmount = 50;
+const dailyRentalRate = 40;
```
volodymyr-soltys97, line 10: You need to move these variables inside the function, as they are only used within the function

src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
...
+
 function calculateRentalCost(days) {
-  // write code here
+  if (days < minDaysForDiscount) {
+    return days * dailyRentalRate;
+  }
+
+  if (days >= minDaysForDiscount && days < maxDaysForDiscount) {
```
volodymyr-soltys97, line 17:
```suggestion
  if (days < maxDaysForDiscount) {
```

src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
...
+const maxDaysForDiscount = 7;
+const baseDiscountAmount = 20;
+const largeDiscountAmount = 50;
+const dailyRentalRate = 40;
+
 function calculateRentalCost(days) {
-  // write code here
+  if (days < minDaysForDiscount) {
+    return days * dailyRentalRate;
+  }
```
volodymyr-soltys97, line 15: This condition is redundant, you can immediately return after two ifs

src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
...
+const minDaysForDiscount = 3;
+const maxDaysForDiscount = 7;
+const baseDiscountAmount = 20;
+const largeDiscountAmount = 50;
+const dailyRentalRate = 40;
+
 function calculateRentalCost(days) {
-  // write code here
+  if (days < minDaysForDiscount) {
+    return days * dailyRentalRate;
```
volodymyr-soltys97, line 14: Сreate a variable for this logic `days * dailyRentalRate` and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Good job 👍

PR #4455, opened 2024-07-09
solution

adonch:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
+  const RENT_PRICE = 40;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
```
volodymyr-soltys97, line 13: Redundant
```suggestion
  if (days < LONG_TERM) {
```

volodymyr-soltys97 (approved):
Good job 👍

src/calculateRentalCost.js
adonch: But we have "if you rent the car for `3` or more days" from task

PR #4468, opened 2024-07-11
Solved the task

Pokhvalenkos:


etojeDenys (changes requested):
image
pls check the checklist

src/calculateRentalCost.js
```diff
@@ -5,6 +5,19 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const BASE_PRICE = days * 40;
```
etojeDenys, line 12:
```suggestion
  const basePrice = days * 40;
```
it's not a constant

src/calculateRentalCost.js
```diff
@@ -5,6 +5,19 @@
...
+
+  if (days >= LONG_TERM) {
+    return BASE_PRICE - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
+    return BASE_PRICE - SHORT_TERM_DISCOUNT;
+  }
```
etojeDenys, line 18:
```suggestion
  if (days >= LONG_TERM) {
    return BASE_PRICE - LONG_TERM_DISCOUNT;
  }
  
   if (days >= SHORT_TERM) {
    return BASE_PRICE - SHORT_TERM_DISCOUNT;
  }
```
else is redundant here, cause we use return in our if statement

etojeDenys (changes requested):

etojeDenys (approved):

PR #4471, opened 2024-07-12
Solution

MK030191:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+  const DISCOUNT_AFTER_7_DAYS = 50;
+
+  if (days <= 2) {
+    return days * EVERY_DAY_COST;
+  }
+
+  if (days <= 6) {
+    return days * EVERY_DAY_COST - DISCOUNT_AFTER_3_DAYS;
+  } else {
+    return days * EVERY_DAY_COST - DISCOUNT_AFTER_7_DAYS;
```
volodymyr-soltys97, line 18: Create a variables for 2 and 6, and use only 2 if conditions, without else

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

vadiimvooo (approved):
Well done

PR #4498, opened 2024-07-22
Solution

kamilbykow:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,25 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const RENT_COST = 40;
```
Zibi95, line 13: Variable names have to bring meaning about the value that it is keeping so you should be here as descriptive as you can. 

Here is how i would call those variables

``` 
const LONG_TERM_RENTAL_DAYS_THRESHOLD = 7;
const LONG_TERM_RENTAL_DISCOUNT_AMOUNT = 50;
const SHORT_TERM_RENTAL_DAYS_THRESHOLD = 3;
const SHORT_TERM_RENTAL_DISCOUNT_AMOUNT = 20;
const DAILY_RENTAL_COST = 40;
```

Zibi95 (approved):
Great job with implementation

PR #4517, opened 2024-07-31
Develop

Mr36Tomek:


BogdanMaliuta (changes requested):
avoid using if/else, just use if and return, it`s better for readability.Also remove comments from code

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  // The car coste calculator
+  let result = days * 40;
+
+  if (days >= '7') {
```
natalia-klonowska, line 10: [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md): Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
+    result -= '50';
+
+    return result;
```
natalia-klonowska, line 13: you can change result to const and return the outcome of operation directly
also then change naming of result to be more descriptive ;)

natalia-klonowska (changes requested):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,24 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  let price = days * 40;
+
+  if (days >= 7) {
+    price -= LONG_TERM_DISCOUNT;
+
+    return price;
+  }
+
+  if (days >= 3) {
```
Zibi95, line 18: Those numbers are called magic numbers. Avoid using magic numbers and keep them in variables with descriptive name.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,24 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  // The car coste calculator
```
Zibi95, line 7: Delete comments before creating PR

src/calculateRentalCost.js
```diff
@@ -4,7 +4,24 @@
...
+  if (days >= 7) {
+    price -= LONG_TERM_DISCOUNT;
+
+    return price;
+  }
+
+  if (days >= 3) {
+    price -= SHORT_TERM_DISCOUNT;
+
+    return price;
```
Zibi95, line 21: You can simply write `return price - SHORT_TERM_DISCOUNT`

Zibi95 (approved):
Good job logic works so I can approve this task for you but please read the comments and apply those changes.

PR #4518, opened 2024-08-01
solution

VitaliiNez:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+  const LONG_TERM_DISCOUNT = 50;
+  const BASE_PRICE = days * PRICE_PER_DAY;
+
+  switch (true) {
+    case days >= LONG_TERM:
+      return BASE_PRICE - LONG_TERM_DISCOUNT;
+    case days >= SHORT_TERM:
+      return BASE_PRICE - SHORT_TERM_DISCOUNT;
+    default:
+      return BASE_PRICE;
```
etojeDenys, line 20: better to use if/else here, because you are not checking the exact value

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const BASE_PRICE = days * PRICE_PER_DAY;
```
etojeDenys, line 12: it's not a constant, because it use days in the calculation

etojeDenys (approved):

PR #4524, opened 2024-08-02
rentalCost solution

MaksymMohyla:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,28 @@
...
+  const oneDayCost = 40;
+  const shortTerm = 3;
+  const shortTermDiscount = 20;
+  const longTerm = 7;
+  const longTermDiscount = 50;
+  let basePrice = 0;
+
+  for (let i = 0; i < days; i++) {
+    basePrice += oneDayCost;
+  }
```
etojeDenys, line 16: why do you need the loop here, you can just multiply those numbers

src/calculateRentalCost.js
```diff
@@ -4,7 +4,28 @@
...
+
+  switch (true) {
+    case days >= longTerm:
+      basePrice -= longTermDiscount;
+      break;
+
+    case days >= shortTerm:
+      basePrice -= shortTermDiscount;
+      break;
```
etojeDenys, line 25: better to use multiple if statements here instead of switch case, because you are not checking some exact value

src/calculateRentalCost.js
```diff
@@ -4,7 +4,28 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayCost = 40;
+  const shortTerm = 3;
+  const shortTermDiscount = 20;
+  const longTerm = 7;
+  const longTermDiscount = 50;
```
etojeDenys, line 11: all constants should be in uppercase

etojeDenys (changes requested):

etojeDenys (approved):
amazing

PR #4558, opened 2024-08-10
Solution

Yexelay:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const longTermDiscount = 50;
+  const normalTermDiscount = 20;
+  const rentalCost = 40;
+  const longTerm = 7;
+  const normalTerm = 3;
```
etojeDenys, line 11: all constants better to name in UPPER_CASE

etojeDenys (changes requested):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const LONG_TERM_DISCOUNT = 50;
+  const NORMAL_TERM_DISCOUNT = 20;
+  const RENTAL_COST = 40;
+  const LONG_TERM = 7;
+  const NORMAL_TERM = 3;
+  const TOTAL_COST = days * RENTAL_COST;
```
etojeDenys, line 12: it's not a constant

etojeDenys (approved):
good job

src/calculateRentalCost.js
Yexelay: AI bot suggestions don't always need to be applied?

PR #4562, opened 2024-08-12
Solution

joaoguilherme000:
- [DEMO LINK] (https://github.com/joaoguilherme000/Mate-Tarefa1/blob/master/src/calculateRentalCost.js)
- [TEST LINK] (https://github.com/joaoguilherme000/Mate-Tarefa1/blob/master/src/calculateRentalCost.test.js)


pedro-ruas (approved):
João, I see you understood the task and knew how to solve it perfectly.
However, I also believe the guides made by the AI Mentor are valid, specially the ones about using "magic numbers".
In the future, try saving those numbers to constants that make sense and allow the code to be understandable on it's own.
The ternary is also ok tu use, but an if statement could also make it more clear in the future.

PR #4570, opened 2024-08-15
task solution

Dorosh90:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+  const LONG_TERM = 7;
+  const SHORT_TERM_DISCOUNT = 20;
+  const LONG_TERM_DISCOUNT = 50;
+  const BASE_PRICE = days * 40;
+
+  if (days < SHORT_TERM) {
+    return BASE_PRICE;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
```
TarasHoliuk, line 17: Good comment, apply the suggested solution:
image

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+    return BASE_PRICE - SHORT_TERM_DISCOUNT;
+  }
+
+  if (days >= LONG_TERM) {
+    return BASE_PRICE - LONG_TERM_DISCOUNT;
+  }
```
TarasHoliuk, line 23:
```suggestion
  return BASE_PRICE - LONG_TERM_DISCOUNT;
```
As it was said in AI comment `if` block is redundant here - you can just return

TarasHoliuk (changes requested):

etojeDenys (approved):
good job

PR #4577, opened 2024-08-19
Solution

Ensenkei:
Implemented function according to the task and checklist.md

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -7,12 +7,14 @@ function calculateRentalCost(days) {
   const TOTAL_AMOUNT = days * 40;
   const LONG_TERM_DISCOUNT = 50;
   const SHORT_TERM_DISCOUNT = 20;
+  const LONG_DAYS = 7;
+  const SHORT_DAYS = 3;
 
-  if (days >= 7) {
+  if (days >= LONG_DAYS) {
     return TOTAL_AMOUNT - LONG_TERM_DISCOUNT;
   }
 
-  if (days >= 3) {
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const TOTAL_AMOUNT = days * 40;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  if (days >= 7) {
```
etojeDenys, line 11: better to move 7 and 3 also to some constants

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const TOTAL_AMOUNT = days * 40;
```
etojeDenys, line 7: it's not a constant

etojeDenys (approved):
good job

src/calculateRentalCost.js
Ensenkei: Shure, thank you!

src/calculateRentalCost.js
Ensenkei: I've tried to use "let" and this error in VS

Ensenkei: Just watched video with Solution of this task and i get what you mean. Thank you!

PR #4586, opened 2024-08-21
Solution

davicatarino:


pedro-ruas (approved):
Good job Davi!
For future reference, if there is a `return` inside your `if`, there is no need to use a `else` clause. Because once the `if` passes, nothing will run afterwards.
So your code would work the same if you only removed all `else` words.
Also, `numberOfDays` isn't needed, because it's only another variable holding the value passed on `days`, you could remove it without any issue and save some memory.

PR #4598, opened 2024-08-25
Add solution

stasLiutyww:


YuliiaShatkovska (approved):
Good job!

polosanya (changes requested):
Almost perfect! Pay attention that `else if` is redundant as the code will not go further after return anyway

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const DEF_PRICE = 40;
+  const MID_TERM = 3;
+  const MID_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
```
Moroz-Dmytro, line 11: Better to move such constants outside the function

Moroz-Dmytro (approved):
Looks good 👍 
Only small comment below

PR #4599, opened 2024-08-25
Solution

gordgelin:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,24 @@
...
+  const sum = days * COST_PER_DAY;
+
+  switch (true) {
+    case days >= DISCOUNT_LVL2_MIN_DAYS:
+      return sum - DISCOUNT_LVL2;
+    case days >= DISCOUNT_LVL1_MIN_DAYS:
+      return sum - DISCOUNT_LVL1;
+    default:
+      return sum;
+  }
```
sarakhmen, line 24: you have code duplication here and also there is a more concise and proper way to do this logic :)

sarakhmen (changes requested):

src/calculateRentalCost.js
gordgelin: ok, let's have some fun with improved solution :)

Elena-Bruyako (changes requested):
Not Java course HW

gordgelin:
yes i know it is not Java HW, i just decided to look through some FS modules, it's just my main course set to Java

Elena-Bruyako (changes requested):
sorry, can't check you HW

PR #4604, opened 2024-08-26
Solution

Nikname2303:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+    return result;
+  }
+
+  if (days >= 3) {
+    result -= 20;
+
+    return result;
+  }
+
+  return result;
```
nklimovych, line 21: I believe it would be better to use `switch(true)` in this case.

nklimovych (approved):

PR #4605, opened 2024-08-26
task has been solved

coldmelnyk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const shortTerm = 2;
```
nactyai, line 7: write your variables like this
const LONG_TERM = 7;

checklist[2]

nactyai (changes requested):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const SHORT_TERM = 2;
+  const LONG_TERM = 7;
+  const PRICE = 40;
+  let sale = 0;
```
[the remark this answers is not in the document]
coldmelnyk: but in a second condition I reassign it if its LONG_TERM

nactyai (approved):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const MIN_DAYS_FOR_DISCOUNT = 3;
+  const LONG_TERM = 7;
+  const PRICE = 40;
+  let sale = 20;
+
+  if (days >= MIN_DAYS_FOR_DISCOUNT && days < LONG_TERM) {
+    return days * PRICE - sale;
+  }
```
VolodymyrKirichenko, line 14: u can do the `Long` condition first, then u won't have to do 2 checks here)
That is, first we check for a larger number

```suggestion
  if (days >= LONG_TERM) {
    return days * PRICE - sale;
  }
```

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
+
+  if (days >= LONG_TERM) {
+    sale = 50;
+
+    return days * PRICE - sale;
```
VolodymyrKirichenko, line 19: u can move it to variable, something like - `defaultPrice = days + price` and then decrease this value

VolodymyrKirichenko (changes requested):
Good, but let's make ur code more better 👍

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const MIN_DAYS_FOR_DISCOUNT = 3;
+  const LONG_TERM = 7;
+  const PRICE = 40;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
```

BudnikOleksii (approved):
👍

PR #4607, opened 2024-08-26
Solution

Valeraom:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,29 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const ONE_DAY_COST = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  let totalAmount = 0;
```
YuriiMykhailenko, line 13: You can initialise this variable with calculation like in lline 27, then make if section shorter, just return `tot-amount - discount`.

YuriiMykhailenko (approved):
Great Job!!! 
Just consider my comment in the future.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,29 @@
...
+
+  if (days >= LONG_TERM) {
+    totalAmount = ONE_DAY_COST * days - LONG_TERM_DISCOUNT;
```
VolodymyrKirichenko, line 16: u can make a variable for it, something like:
`defaultPrice = ONE_DAY_COST * days`

src/calculateRentalCost.js
VolodymyrKirichenko, line 13: but why? u can return `defaultPrice - discount` immediately. U don't need a variable for this

src/calculateRentalCost.js
```diff
@@ -4,7 +4,29 @@
...
+
+    return totalAmount;
+  }
+
+  if (days >= SHORT_TERM) {
+    totalAmount = ONE_DAY_COST * days - SHORT_TERM_DISCOUNT;
+
+    return totalAmount;
```
VolodymyrKirichenko, line 24: u can use this option
```suggestion
    return DEFAULT_PRICE - SHORT_TERM_DISCOUNT;
```

VolodymyrKirichenko (changes requested):
Overall good, but let's make ur code better 👍

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const ONE_DAY_COST = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
```

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+
+  const DEFAULT_PRICE = ONE_DAY_COST * days;
```
BudnikOleksii, line 13:
```suggestion
  const currentPrice = ONE_DAY_COST * days;
```
Variable names should be in camelCase

BudnikOleksii (approved):
Good work 👍

PR #4610, opened 2024-08-26
task solution

Ivanavi808:
solution

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -5,6 +5,7 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  return days < 3 ? days * 40 : days < 7 ? days * 40 - 20 : days * 40 - 50;
 }
 
 module.exports = calculateRentalCost;
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -1,26 +1,20 @@
-/**
- * @param {number} days
- *
- * @return {number}
- */
 function calculateRentalCost(days) {
-  // write code here
-  // return days < 3 ? days * 40 : days < 7 ? days * 40 - 20 : days * 40 - 50;
+  const PRICE = 40;
+  const LONGTERMDISCOUNT = 50;
+  const LONGTERM = 7;
+  const SHORTTERMDISCOUNT = 20;
+  const SHORTTERM = 3;
+  const TOTALCOST = days * PRICE;
 
-  const price = 40;
-  const longTermDiscount = 50;
-  const longTerm = 7;
-  const shortTermDiscount = 20;
-  const shortTerm = 3;
-  const totalCost = days * price;
-
-  if (days >= longTerm) {
-    return totalCost - longTermDiscount;
-  } else if (days >= shortTerm) {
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -5,6 +5,7 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  return days < 3 ? days * 40 : days < 7 ? days * 40 - 20 : days * 40 - 50;
```
Sholudyvyy, line 8: This code is very hard. You need use if construction for this.

Sholudyvyy, line 8: [CheckStyle] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
BAD EXAMPLE:
if (numberOfDays >= 7) {
  return basePrice - 50;
}
GOOD EXAMPLE:
const LONG_TERM = 7;

}

dokrod (changes requested):
Good work.👍 Correct what Stanislav told you.

src/calculateRentalCost.js
```diff
@@ -5,6 +5,22 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  // return days < 3 ? days * 40 : days < 7 ? days * 40 - 20 : days * 40 - 50;
```
BudnikOleksii, line 8: Remove all comments

src/calculateRentalCost.js
```diff
@@ -5,6 +5,22 @@
...
+
+  const price = 40;
+  const longTermDiscount = 50;
+  const longTerm = 7;
+  const shortTermDiscount = 20;
+  const shortTerm = 3;
```
BudnikOleksii, line 14: Constants should be in UPPER_CASE, and move them outside the function

src/calculateRentalCost.js
```diff
@@ -5,6 +5,22 @@
...
+  const totalCost = days * price;
+
+  if (days >= longTerm) {
+    return totalCost - longTermDiscount;
+  } else if (days >= shortTerm) {
+    return totalCost - shortTermDiscount;
+  }
```
BudnikOleksii, line 21:
```suggestion
  }
  
  if (days >= SHORT_TERM) {
    return totalCost - SHORT_TERM_DISCOUNT;
  }
```
Avoid `else if`

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments should be fixed

src/calculateRentalCost.js
```diff
@@ -1,10 +1,20 @@
...
- *
- * @return {number}
- */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE = 40;
+  const LONGTERMDISCOUNT = 50;
+  const LONGTERM = 7;
+  const SHORTTERMDISCOUNT = 20;
+  const SHORTTERM = 3;
```
VolodymyrKirichenko, line 6: move these variables outside the function so that they are not recreated when the function is called

VolodymyrKirichenko (approved):
Good job 😮‍💨

PR #4614, opened 2024-08-26
on-vacation task

EustasTheMonk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
```
VVladLytvyn, line 7: good job, but you have something to fix: you could change names for `const` by `midTerm` to `MID_TERM`. it's not a mistake but there some reasons to to like that: 

1. you can easily notice constants in code and distinguish them from variables.
2. if you see that it's constant, you will not try accidentally change it.
3. and also too many developers agree with this rule, and write like that.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
+  const midTerm = 3;
+  const longTerm = 7;
+  const longTermDiscount = 50;
+  const midTermDiscount = 20;
+  const totalCost = pricePerDay * days;
```
VVladLytvyn, line 12: you should change `const totalCost` to `let totalCost`; 
change const to let - because `days` is a changing variable;

VVladLytvyn (changes requested):
👍 nice job, please fix your code

VVladLytvyn (approved):

Moroz-Dmytro (changes requested):
Fix small comment below 😉

polosanya (approved):
Well done!

PR #4626, opened 2024-08-28
Solution

Hobudo23:
"Solution"

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -4,7 +4,10 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
+
   let TAX_PER_DAY = 40;
+  let LONG_TERMIN_DISCOUNT = 50;
+  let MIDDLE_TERMIN_DISCOUNT = 20;
   let LONG_TERMIN = 7;
   let MIDLE_TERMIN = 3;
 
@@ -15,9 +18,9 @@ function calculateRentalCost(days) {
   if (days < MIDLE_TERMIN) {
     return days * TAX_PER_DAY;
   } else if (days >= MIDLE_TERMIN && days < LONG_TERMIN) {
-    return days * TAX_PER_DAY - 20;
+    return days * TAX_PER_DAY - MIDDLE_TERMIN_DISCOUNT;
   } else {
-    return days * TAX_PER_DAY - 50;
+    return days * TAX_PER_DAY - LONG_TERMIN_DISCOUNT;
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let res = 0;
+
+  for (let i = 0; i < days; i++) {
```
VolodymyrKirichenko, line 9: u don't need a loop here)
use mathematic.
If we know the number of days and the price for 1 day - we can know the base price for all vacation, right?

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+    res += 40;
+  }
+
+  if (days >= 3 && days < 7) {
+    res -= 20;
+  }
+
+  if (days >= 7) {
+    res -= 50;
```
VolodymyrKirichenko, line 18: use constants for values ​​that never change.
for example:
```suggestion
    return basePrise - LONG_TERM_DISCOUNT;
```

VolodymyrKirichenko, line 18: move all `constants` outside the function

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  let res = 0;
+
+  for (let i = 0; i < days; i++) {
+    res += 40;
+  }
+
+  if (days >= 3 && days < 7) {
```
VolodymyrKirichenko, line 13: Знімок екрана 2024-09-02 о 15 01 28

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let res = 0;
```
VolodymyrKirichenko, line 7: at the end of all comments, u will understand that u do not need this variable, because u will return the result immediately after the condition in your `if`

VolodymyrKirichenko (changes requested):
Task is done, but let's prove it)
1) U don't need loop for this task 🥹 
2) Check [Checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md), look on my comments and fix ur code.

Hobudo23:
I change values by variables

djtyrf312 (approved):
Approved

PR #4629, opened 2024-08-29

Tulenty:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
```
volodymyr-soltys97, line 12: Create a variable for base price and use it everywhere
```suggestion
  const SHORT_TERM_DISCOUNT = 20;
  const basePrice = days * cost;
  
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code, also, you need to fix AI comments

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
```
etojeDenys, line 7: it's a constant

etojeDenys (approved):
lgtm

PR #4630, opened 2024-08-29
Develop

YevhenRDoIT:
Add js file with function to calculate rental cost

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -12,15 +12,17 @@ function calculateRentalCost(days) {
   const SHORT_TERM = 3;
   const SHORT_TERM_DISCOUNT = 20;
 
+  const basePrice = days * PRICE_PER_DAY;
+
   if (days >= LONG_TERM) {
-    return days * PRICE_PER_DAY - LONG_TERM_DISCOUNT;
+    return basePrice - LONG_TERM_DISCOUNT;
   }
 
   if (days >= SHORT_TERM) {
-    return days * PRICE_PER_DAY - SHORT_TERM_DISCOUNT;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
```
volodymyr-soltys97, line 14: Сreate a variable for base price and use it everywhere 
```suggestion
  const SHORT_TERM_DISCOUNT = 20;
  const BASE_PRICE = days * PRICE_PER_DAY;
  
```

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+  if (days >= LONG_TERM) {
+    return days * PRICE_PER_DAY - LONG_TERM_DISCOUNT;
```
volodymyr-soltys97, line 16:
```suggestion
    return BASE_PRICE - LONG_TERM_DISCOUNT;
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/calculateRentalCost.js
YevhenRDoIT: ok

YevhenRDoIT:
I have created a new variable for base price and use it everywhere

volodymyr-soltys97 (approved):
Great! 🔥

PR #4644, opened 2024-09-02
task solution

W3zzie:


BogdanMaliuta (changes requested):
AI wrote a good suggestions for you, please take a look at them

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
-}
+  const dailyRate = 40;
+  const longTermDiscount = 50;
+  const shortTermDiscount = 20;
+  const longTermThreshold = 7;
+  const shortTermThreshold = 3;
```
Zibi95, line 11: For such variables we should use UPPERCASE. Using UPPERCASE for constant variables is a common convention in many programming languages to indicate that these values should not change during the execution of the program.
```suggestion
  const DAILY_RATE = 40;
  const LONG_TERM_DISCOUNT = 50;
  const SHORT_TERM_DISCOUNT = 20;
  const LONG_TERM_THRESHOLD = 7;
  const SHORT_TERM_THRESHOLD = 3;
```

Zibi95 (approved):
Good job. Code works and looks fine. I have left one comment for future reference.

PR #4650, opened 2024-09-03
Solution

maksimvayder:


Daria-afk (commented):
Look at the checklist, it describes what not to use else

maksimvayder:
The checklist only shows a bad example of using the else if structure, in which return is immediately used after else, but does not say that it cannot be used.

vlukashuk:
gj

vlukashuk (approved):

KarolVlasova (approved):

PR #4658, opened 2024-09-03
Solution

Lyudmila2024:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
```
djtyrf312, line 6: you need to write a function with logic inside but not the hard coded values

djtyrf312 (changes requested):
write a function with logic but not the hard coded values

KarolVlasova (approved):
approve, but do not push .json files

PR #4659, opened 2024-09-03
Solution

MarianaQAaug2024:


djtyrf312 (changes requested):
write the function with logic but not the hardcoded values

PR #4666, opened 2024-09-04
Solution

Anton777148:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const SHORT_DISCOUNT = 20;
+  const LONG_DISCOUNT = 50;
+
+  if (days >= LONG_TERM) {
+    return days * PRICE - LONG_DISCOUNT;
```
volodymyr-soltys97, line 14: You can create a variable for base price and use it everywhere
```suggestion
    return BASE_PRICE - LONG_DISCOUNT;
```

volodymyr-soltys97 (approved):
Good job! 👍

PR #4667, opened 2024-09-04
Solution Strakh Valeriia

vs-qaaug:


djtyrf312 (changes requested):
you do not need to do anything in the test files. You had to do everything in the calculateRentalCost file. Remove file with tests

vs-qaaug (commented):
Deleted file with tests

PR #4675, opened 2024-09-04
solution

MarianaQAaug2024:


src/calculateRentalCost.js
```diff
@@ -1,10 +1,20 @@
 /**
  * @param {number} days
- *
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRate = 40; // Вартість оренди за день
```
djtyrf312, line 6:
```suggestion
  const dailyRate = 40;
```
remove all comments from the file. It is not a good practice to comment the code

djtyrf312 (changes requested):
remove all comments from pr

djtyrf312 (changes requested):
remove extra file calculateRentalCost copy.js

djtyrf312 (changes requested):
replied in the chat on the platform

KarolVlasova (approved):

PR #4681, opened 2024-09-05
MySoluution

iryna112:


KarolVlasova (changes requested):
you hardcoded your function. Use variables and math methods

djtyrf312 (approved):
approved

PR #4697, opened 2024-09-10
Solution

MichaelK-jpg:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const everyDayCarRentCost = 40;
+  let discount = 0;
+
+  if (days >= 3 && days < 7) {
```
natalia-klonowska: [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md): Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
image

src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const everyDayCarRentCost = 40;
+  let discount = 0;
```
natalia-klonowska: create two const variables for possible discounts instead of reassigning this one and then you can move return inside if statement using specific discount

natalia-klonowska (changes requested):

natalia-klonowska (approved):
Great work 🔥

PR #4707, opened 2024-09-14
Solution

Kuxim:


src/calculateRentalCost.js
```diff
@@ -1,10 +1,16 @@
...
- * @param {number} days
- *
- * @return {number}
- */
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
+
+  if (days >= 7) {
+    totalCost -= 50;
```
TarasHoliuk, line 5:
```suggestion
    return totalCost -= BIG_DISCOUNT;
```
you can immediately return in each case and use `const` for totalCount variable

TarasHoliuk (changes requested):
There are good and fair comments by the ai-mentor, please, apply them

Kuxim:
Please check how I changed the code.

src/calculateRentalCost.js
```diff
@@ -1,10 +1,22 @@
...
+  // Применяем скидки и возвращаем стоимость
+  if (days >= LONG_TERM_RENTAL_DAYS) {
+    return totalCost - BIG_DISCOUNT;
+  }
+
+  if (days >= MID_TERM_RENTAL_DAYS) {
+    return totalCost - SMALL_DISCOUNT;
+  }
+
+  return totalCost; // Если арендовано меньше 3 дней скидка не применяется
```
etojeDenys, line 19: Remove comment. Try to write obvious code that explains all we need without comments)

etojeDenys (approved):

PR #4709, opened 2024-09-14
Solution

1luki9901:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
+
+  if (days >= 7) {
+    totalCost -= 50;
```
natalia-klonowska, line 10: you can return result right away and then you don't need to change value of totalCost
```suggestion
    return totalCost - 50;
```

natalia-klonowska (changes requested):
Look at [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) to better understand how to fix issues I've marked ;)

src/calculateRentalCost.js
```diff
@@ -4,12 +4,16 @@
...
 
-  if (days >= 7) {
-    totalCost -= 50;
-  } else if (days >= 3) {
-    totalCost -= 20;
+  if (days >= week) {
+    return totalCost - longRentalDiscount;
+  } else if (days >= threeDays) {
+    return totalCost - shortRentalDiscount;
   }
```
natalia-klonowska, line 17: [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md): don't add `else` after `if` with `return` - the code after it won't be executed anyway.

natalia-klonowska (changes requested):
just needs one more adjustment :)

Zibi95 (approved):
LGTM!

PR #4750, opened 2024-09-25
my solution

minimal005:
my code

### DIFF
--- a/src/calculateRentalCost.js
+++ b/src/calculateRentalCost.js
@@ -1,10 +1,21 @@
+'use strict';
+
 /**
  * @param {number} days
  *
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const costForDay = 40;
+  let discount = 0;
+
+  if (days >= 7) {
+    discount = 50;
+  } else if (days >= 3) {
+    discount = 20;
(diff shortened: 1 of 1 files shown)


src/calculateRentalCost.js
```diff
@@ -1,10 +1,28 @@
...
  *
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const COST_FOR_DAY = 40;
+  const LITTLE_DAYS = 3;
+  const LITTLE_DISCOUNT = 20;
+  const BIG_DAYS = 7;
+  const BIG_DISCOUNT = 50;
```
BudnikOleksii, line 13: Move constants outside the function

src/calculateRentalCost.js
```diff
@@ -1,10 +1,28 @@
...
+
+  const TOTAL_COST = days * COST_FOR_DAY;
```
BudnikOleksii, line 15:
```suggestion
  const totalCost = days * COST_FOR_DAY;
```
Variables should be in camelCase, only constants that are not dynamic values in UPPER_CASE

BudnikOleksii (approved):
Good work 👍

PR #4768, opened 2024-10-01
Solution

OksanaW8:


src/calculateRentalCost.js
eule1, line 1: for the name of constants, variables and functions better to use camelCase

eule1 (changes requested):
pls use camelCase

OksanaW8:
The task checklist requires a code like mine.

PR #4770, opened 2024-10-02
Solution

Krrampuss:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const everyDayRent = 40;
+  let totalSum = everyDayRent * days;
+
+  if (days >= 7) {
+    totalSum = everyDayRent * days - 50;
+  } else if (days >= 3) {
+    totalSum = everyDayRent * days - 20;
```
BudnikOleksii, line 13: - 7, 50, 3, 20 are magic values
image
- Prefer if with return over if else to simplify later conditions.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const everyDayRent = 40;
+  let totalSum = everyDayRent * days;
```
BudnikOleksii, line 8: Prefer const over let where possible, to avoid unintentional changes.

BudnikOleksii (changes requested):
Good start, but don't forget about the [checklist](https://github.com/Krrampuss/js_task-transportation-on-vacation/blob/develop/checklist.md)

src/calculateRentalCost.js
BudnikOleksii: > Prefer const over let where possible, to avoid unintentional changes.

Not fixed

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const everyDayRent = 40;
+  const discountFor7Days = 50;
+  const discountFor3Days = 20;
+  const minDaysFor7DayDiscount = 7;
+  const minDaysFor3DayDiscount = 3;
```
BudnikOleksii, line 11: Constants should be in UPPER_CASE, and you can move them outside the function
image

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
+  let totalSum = everyDayRent * days;
+
+  if (days >= minDaysFor7DayDiscount) {
+    totalSum = everyDayRent * days - discountFor7Days;
+  } else if (days >= minDaysFor3DayDiscount) {
+    totalSum = everyDayRent * days - discountFor3Days;
+  }
```
BudnikOleksii, line 18:
```suggestion
  if (days >= minDaysFor7DayDiscount) {
    return totalSum - discountFor7Days;
  }
  
  if (days >= minDaysFor3DayDiscount) {
    return totalSum - discountFor3Days;
  }
```
- Avoid else if and
- You don't need to assign the value, just return it
image

BudnikOleksii (changes requested):
Good, just a few moments still should be fixed

BudnikOleksii (approved):
Looks good 🚀

PR #4788, opened 2024-10-14
Solution

OkMoroz:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const DAILY_COST = 40;
+  const MANY_DAYS = 7;
+  const BIG_DISCOUNT = 50;
+  const FEW_DAYS = 3;
+  const SMALL_DISCOUNT = 20;
```

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+
+  let totalCost = DAILY_COST * days;
```
BudnikOleksii, line 13:
```suggestion
  const totalCost = DAILY_COST * days;
```
Prefer const over let

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+
+  if (days >= MANY_DAYS) {
+    return (totalCost = totalCost - BIG_DISCOUNT);
```
BudnikOleksii, line 16:
```suggestion
    return totalCost - BIG_DISCOUNT;
```
You don't need to assign the value, just return it

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

BudnikOleksii (approved):
Looks good 🚀

PR #4825, opened 2024-10-26
add solution

schyrva:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const basePrice = days * PRICE_PER_DAY;
```
chelsea7smile, line 13: -better use `let` for variable `basePrice`;
[Checklist] 3. Prefer `const` over `let` where possible, to avoid unintentional changes.

chelsea7smile (changes requested):
Good job!
Let fix  `const` to `let`:)

src/calculateRentalCost.js
```diff
@@ -10,7 +10,7 @@ function calculateRentalCost(days) {
   const SHORT_TERM = 3;
   const SHORT_TERM_DISCOUNT = 20;
 
-  const basePrice = days * PRICE_PER_DAY;
+  let basePrice = days * PRICE_PER_DAY;
```
chelsea7smile, line 13: Cool!

Check this line at the end also - `return basePrice;`
- because ` return PRICE_PER_DAY * days;` looks way better (ai mentor also give this comment)

chelsea7smile (changes requested):

chelsea7smile (approved):

src/calculateRentalCost.js
BudnikOleksii: > Cool!

Why? If you don't need to change variable use const

Prefer const over let where possible, to avoid unintentional changes.

BudnikOleksii (approved):
Good work 👍

src/calculateRentalCost.js
chelsea7smile: Yeah, I've got confused, because Misha in video and also AI said this :
"The current logic applies only one discount, either long-term or short-term, based on the number of days. If the task requires cumulative discounts (e.g., both long-term and short-term discounts for rentals longer than 7 days), you need to adjust the logic to apply both discounts where applicable. Otherwise, ensure this logic aligns with the task requirements."
-and based on this I decided that `let` is possible if discounts are added/changed

> > Cool!
> 
> Why? If you don't need to change variable use const
> 
> Prefer const over let where possible, to avoid unintentional changes.

PR #4826, opened 2024-10-26
solution

petro-boiarenko:


src/calculateRentalCost.js
```diff
@@ -1,10 +1,24 @@
...
-  // write code here
+  const dayCosts = 40;
+  let totalAmount = days * dayCosts;
+  const longTermLoan = 7;
+  const longTermDiscount = 50;
+  const mediumTermLoan = 3;
+  const mediumTermDiscount = 20;
+
+  if (days >= longTermLoan) {
+    totalAmount -= longTermDiscount;
```
schyrva, line 10: You can simplify this by using this expression `return totalAmount - longTermDiscount;`

src/calculateRentalCost.js
```diff
@@ -1,10 +1,24 @@
-/**
- * @param {number} days
- *
- * @return {number}
- */
 function calculateRentalCost(days) {
-  // write code here
+  const dayCosts = 40;
```
schyrva, line 2: It is a good practice to constants with uppercase and underline like this `const DAY_COST = 40;`

schyrva (approved):
Goof Job!
Everything is good, but you can simplify return expression and rename constants :)

src/calculateRentalCost.js
BudnikOleksii: > You can simplify this by using this expression `return totalAmount - longTermDiscount;`

Good point

BudnikOleksii (changes requested):
Good work 👍 
Check the comments and improve your solution

PR #4828, opened 2024-10-26
Develop

chelsea7smile:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+
+  if (days >= LONG_TERM) {
+    return basePrice - DISCOUNT_LONG_TERM;
+  }
+
+  if (days >= SHORT_TERM) {
+    return basePrice - DISCOUNT_SHORT_TERM;
+  }
+
+  return PRICE_PER_DAY * days;
```
yuliiazubenko, line 23: You could return `basePrice` from line 13  instead of returning  `PRICE_PER_DAY * days` on line 23

yuliiazubenko (changes requested):
Your code looks good!
I suggest changes only on line 23

yuliiazubenko (commented):
Perfect! :)

yuliiazubenko (approved):

StasSokolov1 (approved):
Looks good to me, I'm approving it🔥

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const DISCOUNT_LONG_TERM = 50;
+  const SHORT_TERM = 3;
+  const DISCOUNT_SHORT_TERM = 20;
```

VolodymyrKirichenko (approved):
Well done 🔥

PR #4830, opened 2024-10-27
add solution

yuliiazubenko:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
-}
+  const RCICE_PER_DAY = 40;
```
maxsabo, line 7: поправ назву замість RCICE на PRICE

maxsabo (changes requested):
Все добре, тільки виправ маленьку помилку :)

maxsabo (approved):

src/calculateRentalCost.js
yuliiazubenko: дякую!

maxsabo (approved):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
...
 function calculateRentalCost(days) {
-  // write code here
-}
+  const PRICE_PER_DAY = 40;
+
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+
+  const MEDIUM_TERM = 3;
+  const MEDIUM_TERM_DISCOUNT = 20;
```
VolodymyrKirichenko, line 13: Move CONSTANTS outside the function.

VolodymyrKirichenko (approved):
Well done 🔥 
But review my comment and fix it

PR #4833, opened 2024-10-27
resolving task

Daniil-102:


pogorielova (approved):
Great job!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
```

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+
+  let basePrice = days * PRICE_PER_DAY;
```
VolodymyrKirichenko, line 13: u don't `let` variable in this place. it never changes

VolodymyrKirichenko (approved):
Well done 🔥 
But check my comments

PR #4834, opened 2024-10-27
Solution

pogorielova:


Eeoneishin:
good job

Eeoneishin (approved):
good job, approve

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const RENTAL_COST = 40;
+  const SHORT_DISCOUNT_PERIOD = 3;
+  const LONG_DISCOUNT_PERIOD = 7;
+  const SHORT_DISCOUNT = 20;
+  const LONG_DISCOUNT = 50;
```

VolodymyrKirichenko (approved):
Well done 🔥 
But check my comment and fix it.

src/calculateRentalCost.js
pogorielova: Thank you, fixed

PR #4836, opened 2024-10-27
added solution

alechsva:


kami2693 (approved):

kami2693:
Nice

kami2693 (approved):

AnderZerfall (approved):
Great job! However, consider using variables for constant values. For example, since the base price for rent is $40 per day, it would be beneficial to define it as `const price = 40`. This way, if you need to change the price in the future, you won't have to search for all instances of "40" in your code.

Additionally, I would simplify the if-else statement a bit. You can adjust it like this:

```javascript
if (days >= 3) {
  if (days >= 7) {
    total -= 50;
  } else {
    total -= 20;
  }
}

return total;
```

Anyway, as everyone says, "don't touch it if it works." 

Nice work!

PR #4839, opened 2024-10-28
first

Jion353:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const COST = 40;
+  const LONG_RENT = 7;
+  const SHORT_RENT = 3;
+  let discount = 0;
+
+  if (days >= LONG_RENT) {
```
so3r, line 12: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values. 
You should add constants for long term discount, and short term discount.

and next lines

src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
...
+    discount = 50;
+  } else if (days >= SHORT_RENT) {
```
so3r, line 14: 4. [checklist] Prefer `if` with `return` over `if else` to simplify later conditions.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const COST = 40;
+  const LONG_RENT = 7;
+  const SHORT_RENT = 3;
+  const DISCOUNT_LONG = 50;
+  const DISCOUNT_SHORT = 20;
+
+  if (days >= LONG_RENT) {
+    return days * COST - DISCOUNT_LONG;
```
BudnikOleksii, line 14: You do `days * COST` 3 times, create a variable

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+  } else if (days >= SHORT_RENT) {
```
BudnikOleksii, line 15:
```suggestion
  } 
  
  if (days >= SHORT_RENT) {
```

src/calculateRentalCost.js
BudnikOleksii: > 4. [checklist] Prefer `if` with `return` over `if else` to simplify later conditions.

Good point

BudnikOleksii (approved):
Good job 👍 
Approved, but left a few suggestions

PR #4840, opened 2024-10-28
solution

Yevhen-Filimonov:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,19 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const price = 40;
+  const x = price * days;
```
Mariagosp, line 9: 1. [checklist] Make names descriptive, so they explain the purpose of variables and functions.

You can use `RENT_COST` or something else instead of `x`.

src/calculateRentalCost.js
```diff
@@ -5,6 +5,19 @@
...
+  if (days >= 7) {
```
Mariagosp, line 10: 2. [checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

Mariagosp (changes requested):
Good job!))
Please fix your code

src/calculateRentalCost.js
```diff
@@ -5,6 +5,19 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const price = 40;
```
VolodymyrKirichenko, line 8: 1) use CONSTANTS for all the static values
```suggestion
  const PRICE_PER_DAY = 40;
```
2) Move CONSTANTS outside the function, so that they don't recreate each time the function is called

Look at the [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
VolodymyrKirichenko, line 9: what the x?idk..
change the variable name
```suggestion
  const basePrice = price * days;
```

src/calculateRentalCost.js
VolodymyrKirichenko: good point 🔥 
do it!
use [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) for it

VolodymyrKirichenko (changes requested):
Good work, but u need to check the  [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) and make some modifies

src/calculateRentalCost.js
```diff
@@ -5,6 +5,19 @@
...
+  const x = price * days;
+  if (days >= 7) {
+    return  x - 50;
```
VolodymyrKirichenko, line 11: change all the Magic numbers
```suggestion
    return  basePrice - LONG_TERM_DISCOUNT;
```

src/calculateRentalCost.js
```diff
@@ -4,7 +4,26 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  let PRICE_ALL = PRICE_PER_DAY * days;
```
BudnikOleksii, line 8:
```suggestion
  const currentPrice = PRICE_PER_DAY * days;
```
Prefer const over let
And variable names should be in camelCase

src/calculateRentalCost.js
```diff
@@ -4,7 +4,26 @@
...
+  const SHORT_DISCOUNT = 20;
+  const LONG_DISCOUNT = 50;
+  const SHORT = 3;
+  const LONG = 7;
```
BudnikOleksii, line 12:
```suggestion
  const SHORT_TERM = 3;
  const LONG_TERM = 7;
```
Short and long what?

src/calculateRentalCost.js
```diff
@@ -4,7 +4,26 @@
...
+
+  if (days >= LONG) {
+    return  PRICE_ALL - LONG_DISCOUNT;
+  }
+
+  if (days >= SHORT) {
+    return PRICE_ALL - SHORT_DISCOUNT;
+  }
+
+  if (days < SHORT) {
```
BudnikOleksii, line 23: You don't need this if, you can just return in this case

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const currentPrice = PRICE_PER_DAY * days;
+  const SHORT_DISCOUNT = 20;
+  const LONG_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const LONG_TERM = 7;
```
VolodymyrKirichenko, line 12: 1) Move CONSTANTS outside the function.
2) I think basePrice better describes this variable

```suggestion
  const basePrice = PRICE_PER_DAY * days;
```

VolodymyrKirichenko (approved):
Well done 🔥 
But plz, check my comment and fix it 🥹

PR #4842, opened 2024-10-28
Solution

Mariagosp:


VitaliiShc (changes requested):
Good job!
Please, fix your code.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const RENT_COST = 40;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const BASE_PRICE = days * RENT_COST;
```
VitaliiShc, line 13: Const `BASE_PRICE` shouldn't be in HARDCORE_CASE.
This constant is used and matters only inside this function.  In HARDCORE_CASE conventionally constants are named when considered 'frozen' or 'global'.
Such a notation shows that this constant CANNOT be re-declared.
Usually, they are defined at the beginning of the document in the global scope and are not nested in a function (Mikhail's example is only a particularity of this specific task). Unfortunately, Mikhail did not explain what this notation is for.
And the constants in this task have a business meaning, and in fact, their change is possible only at the direction of the business, that is why these constants are considered 'frozen' or 'global'.  `BASE_PRICE` doesn't apply to these - it's a calculated value.

VitaliiShc (changes requested):
upd comments

BudnikOleksii (approved):
Good job 👍

PR #4843, opened 2024-10-28
add task solution

taniabarkovskya:


Ilya-Ilchuk (approved):
Good job!👍

src/calculateRentalCost.js
```diff
@@ -4,7 +4,25 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const CAR_RENT = 40;
+
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
```
VolodymyrKirichenko, line 13: Move CONSTANTS outside the function.

VolodymyrKirichenko (approved):
Well done 🔥 
But check my comment

PR #4844, opened 2024-10-28
Solution

so3r:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,22 @@
...
+const LONG_TERM_DISCOUNT = 50;
+const MID_TERM_DISCOUNT = 20;
+
 function calculateRentalCost(days) {
-  // write code here
+  const basePrice = 40 * days;
+
+  if (days >= LONG_TERM) {
+    return basePrice - LONG_TERM_DISCOUNT;
+  } else if (days >= MID_TERM) {
```
LiliiaVol, line 17: Here will better use if without else

LiliiaVol:
Good work!

src/calculateRentalCost.js
LiliiaVol: 5. DON'T add `else` after `if` with `return` - the code after it won't be executed anyway.

LiliiaVol (changes requested):

LiliiaVol (approved):
Good job!

src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
  *
  * @return {number}
  */
+
+const LONG_TERM = 7;
+const MID_TERM = 3;
```
VolodymyrKirichenko, line 8: maybe will better 🤔 ?
```suggestion
const SHORT_TERM = 3;
```

VolodymyrKirichenko (approved):
Well done 🔥

PR #4853, opened 2024-10-28
calculate rental amount

yzhyhaliuk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40;
+  let totalSum = 0;
+
+  for (let day = 1; day <= days; day++) {
+    totalSum = totalSum + pricePerDay;
+  }
```
maxim2310, line 12: i think you can use multiplying for this)

src/calculateRentalCost.js
```diff
@@ -4,7 +4,20 @@
...
+
+  if (days >= 7) {
+    return totalSum - 50;
```
maxim2310, line 15: looks like magic number, create variable for all discount

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

VitaliyBondarenko1982 (approved):
Looks good

PR #4856, opened 2024-10-28
Solution

SemenVodolazskij:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyCost = 40;
+  let totalCost = days * dailyCost;
+
+  if (days >= 7) {
```
TargoniyAnatoliy, line 10: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
BAD EXAMPLE:
if (numberOfDays >= 7) {
  return basePrice - 50;
}
GOOD EXAMPLE:
const LONG_TERM = 7;

}

TargoniyAnatoliy (changes requested):
Good job!

TargoniyAnatoliy (approved):
Good job!

PR #4870, opened 2024-11-01
Solution

naz4ik:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,23 @@
  *
  * @return {number}
  */
+
 function calculateRentalCost(days) {
-  // write code here
+  const longTerm = 7;
```
anastasiiavorobiova, line 8:
```suggestion
  const LONG_TERM = 7;
```
Consider using the checklist https://github.com/naz4ik/js_task-transportation-on-vacation/blob/develop/checklist.md to complete the task and for variables naming

anastasiiavorobiova (changes requested):
Great work! Almost done!

PR #4923, opened 2024-11-12
Updated calculateRental

Clyacsa:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const priceADay = 40;
+  const LONG_TERM = 7;
```
Wlonix, line 8: const name should be write in camel case
Rechange everywhere

Wlonix (changes requested):
check comment

src/calculateRentalCost.js
Clyacsa: I don't see the mistakes

Clyacsa (commented):
I don't see the mistakes

PR #4925, opened 2024-11-13
Solution

NTelekhovska:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,20 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const priceADay = 40;
```
Bulavskyi, line 8: It is recommended to write constant values in this way: PRICE_A_DAY

Bulavskyi (approved):

PR #4950, opened 2024-11-25
Solution

AnastasiiaFedurko:


Elena-Bruyako (changes requested):
send to review again, please

src/calculateRentalCost.js
```diff
@@ -5,6 +5,18 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const price = 40;
```
Sanderfuck, line 8:
```suggestion
  const pricePerDay = 40;
```

src/calculateRentalCost.js
```diff
@@ -5,6 +5,18 @@
...
+  let total;
+
+  if (days < 3) {
+    total = price * days;
+  } else if (days >= 3 && days <= 6) {
+    total = price * days - 20;
+  } else if (days >= 7) {
+    total = price * days - 50;
```
Sanderfuck, line 16: add constants for strange numbers

src/calculateRentalCost.js
```diff
@@ -5,6 +5,18 @@
...
+  }
+
+  return total;
```
Sanderfuck, line 19:
```suggestion
  return PRICE_PER_DAY * days - discount;
```

Sanderfuck: Try to think about action make here not in every if

Sanderfuck (changes requested):

PR #4996, opened 2024-12-12
Solution

azhadko98:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRate = 40;
+  const discount7Days = 50;
+  const discount3Days = 20;
+  const longTermDays = 7;
+  const midTermDays = 3;
+
+  if (days >= longTermDays) {
+    return days * dailyRate - discount7Days;
```
volodymyr-soltys97, line 14: Create a variable for this logic and use it everywhere
```suggestion
  const midTermDays = 3;
  const basePrice = days * dailyRate;

  if (days >= longTermDays) {
    return basePrice - discount7Days;
```

volodymyr-soltys97 (changes requested):
Good job! 👍 
Let's make your code better

src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRate = 40;
+  const discount7Days = 50;
+  const discount3Days = 20;
+  const longTermDays = 7;
+  const midTermDays = 3;
```
etojeDenys, line 11: it's constants, better to name them in upper case

etojeDenys (approved):

PR #5007, opened 2024-12-17
Solution calculateRentalCost

VitaliiHutsalo:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const priseForDay = 40;
```
alexander-ignatow, line 7:
```suggestion
  const priceForDay = 40;
```
Mind the grammar :)

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+  const bigDiscont = 50;
+  const smallDiscont = 20;
+  const longTerm = 7;
+  const shortTerm = 3;
+
+  if (days >= longTerm) {
+    return days * priseForDay - bigDiscont;
+  } else if (days >= shortTerm && days < longTerm) {
+    return days * priseForDay - smallDiscont;
+  }
```
alexander-ignatow, line 17: You can simplify this

1. No need in `else if` as you `return` from the first `if` block
2. no need to check `longTerm` condition in the second block^ as it is always `true`

I think you should have something like this:

```suggestion
  if (days >= longTerm) {
    return days * priseForDay - bigDiscont;
  } 
  
  if (days >= shortTerm) {
    return days * priseForDay - smallDiscont;
  }
```

alexander-ignatow (approved):
It works, so let's move on 👍 
Please mind the grammar and be more careful with conditions - do not overcomplicate your code

PR #5121, opened 2025-01-20
task solution

KrawczykDamiann:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_OF_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  let rentalCost = days * PRICE_OF_DAY;
```
natalia-klonowska, line 13: you can change it to const if you use this variable instead of changing it

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+
+  if (days >= LONG_TERM) {
+    rentalCost -= LONG_TERM_DISCOUNT;
```
natalia-klonowska, line 16: you can return this value without changing rentalCost

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+  } else if (days >= SHORT_TERM) {
```
natalia-klonowska, line 17: you can change else if to if after using return inside previous if

natalia-klonowska (changes requested):

src/calculateRentalCost.js
```diff
@@ -10,15 +10,15 @@ function calculateRentalCost(days) {
   const SHORT_TERM = 3;
   const SHORT_TERM_DISCOUNT = 20;
 
-  let rentalCost = days * PRICE_OF_DAY;
```
natalia-klonowska: it's better to leave restalCost variable to not repeating days * PRICE_OF_DAY 3 times ;) but change it to const

natalia-klonowska (changes requested):
Great :D just one more fix

natalia-klonowska (approved):

PR #5122, opened 2025-01-20
Solution

KateShepel:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,26 @@
...
+
+  if (days < LONG_TERM) {
+    totalCost = totalCost - SMALL_DISCOUNT;
+
+    return totalCost;
+  }
+
+  totalCost = totalCost - BIG_DISCOUNT;
+
+  return totalCost;
```
alexander-ignatow, line 26: No need to reassign `totalCost`
Just return what you need

```suggestion
  if (days < LONG_TERM) {
    return totalCost - SMALL_DISCOUNT;
  }

  return totalCost - BIG_DISCOUNT;
```

If you do this - you can change 
` let totalCost = days * PRICE_OF_RENT;` to 
` const totalCost = days * PRICE_OF_RENT;`

alexander-ignatow (approved):
Please mind some issues - but it is all good overall 👍

PR #5215, opened 2025-02-07
Solution

R-Ohirok:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const BASE_PRICE = 40;
+  const LONG_TERM = 3;
+  const LONG_TERM_DISCOUNT = 20;
+  const VERY_LONG_TERM = 7;
+  const VERY_LONG_TERM_DISCOUNT = 50;
```
alexander-ignatow, line 12: constants should go outside the function - otherwise they will be re-declared on each function call

alexander-ignatow (approved):
LGTM

PR #5216, opened 2025-02-07
Implement rental cost calculation based on rental days

dvdmsk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const PRICE_PER_DAY = 40;
+  const DISCOUNT_3_DAYS = 20;
+  const DISCOUNT_7_DAYS = 50;
+
+  const totalPrice = PRICE_PER_DAY * days;
+
+  if (days < 3) {
```
etojeDenys: it's better to move 3 and 7 magic numbers to constants too

etojeDenys (approved):

PR #5228, opened 2025-02-10
Solution

vladrlex:


a3ex77 (approved):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const DAILY_RATE = 40;
+  const DISCOUNT_LONG_TERM = 50;
+  const DISCOUNT_MEDIUM_TERM = 20;
+
+  let totalCost = days * DAILY_RATE;
+
+  if (days >= 7) {
```
a3ex77, line 13: Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
BAD EXAMPLE:
if (numberOfDays >= 7) {
  return basePrice - 50;
}
GOOD EXAMPLE:
const LONG_TERM = 7;

}

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+    totalCost -= DISCOUNT_LONG_TERM;
+  } else if (days >= 3) {
```
a3ex77, line 15: DON'T add else after if with return - the code after it won't be executed anyway.
BAD EXAMPLE:
if (condition) {
  return x;
} else {
  return y;
}
GOOD EXAMPLE:
if (condition) {
  return x;
}

return y;

a3ex77 (changes requested):
Good job! Initially, I didn’t understand how to do the check correctly. Added comments, make small changes. And everything is fine!

PR #5230, opened 2025-02-10
Solution

a3ex77:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayCost = 40;
+  let totalCost = days * oneDayCost;
+
+  if (days >= 7) {
```
daniil-trusov: 2. [checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values:
```
- BAD EXAMPLE:
    ```js
    if (numberOfDays >= 7) {
      return basePrice - 50;
    }
    ```
    - GOOD EXAMPLE: 
    ```js
    const LONG_TERM = 7;

    }
    ```
```
Also check this for next lines

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
...
+    totalCost -= 50;
```
daniil-trusov: It's possible to return value just here instead of assign and return it later (no need to fix, just note). Also applicable for the next check.

daniil-trusov (changes requested):
Good job, but need some minor fixes, please review comments.

src/calculateRentalCost.js
```diff
@@ -5,15 +5,17 @@
...
+  const smallDiscount = 20;
+  const totalCost = days * oneDayCost;
 
   if (days >= 7) {
-    totalCost -= 50;
+    return totalCost - bigDiscount;
   } else if (days >= 3) {
-    totalCost -= 20;
+    return totalCost - smallDiscount;
+  } else {
```
daniil-trusov: 5. [checklist] DON'T add `else` after `if` with `return` - the code after it won't be executed anyway.
```
    - BAD EXAMPLE:
    ```js
    if (condition) {
      return x;
    } else {
      return y;
    }
    ```
    - GOOD EXAMPLE: 
    ```js
    if (condition) {
      return x;
    }

    return y;
    ```
```

src/calculateRentalCost.js
daniil-trusov: > 2. [checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values:
> 
> ```
> - BAD EXAMPLE:
>     ```js
>     if (numberOfDays >= 7) {
>       return basePrice - 50;
>     }
>     ```
>     - GOOD EXAMPLE: 
>     ```js
>     const LONG_TERM = 7;
>     const LONG_TERM_DISCOUNT = 50;
> 
>     if (numberOfDays >= LONG_TERM) {
>       return basePrice - LONG_TERM_DISCOUNT;
>     }
>     ```
> ```
> 
> Also check this for next lines

Wasn't fixed in last commit, please fix

daniil-trusov (changes requested):
Some values were rewritten as constants - it's good, but some values remain hardcoded in 'if' blocks, please fix this. 
Also there's minor problem with 'else' after 'if' with 'return', please check comments

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+  const bigDiscount = 50;
+  const smallDiscount = 20;
+  const longTerm = 7;
+  const shortTerm = 3;
+
+  const totalCost = days * oneDayCost;
+
+  if (days >= longTerm) {
+    return totalCost - bigDiscount;
+  } else if (days >= shortTerm) {
```
daniil-trusov: The 'else' after statement with return has no effect on program, so it can be deleted

daniil-trusov (approved):

PR #5232, opened 2025-02-10
Solution

nataliia-v:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let discount = 0;
+
+  if (days >= 7) {
```
dimqee: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

and for the next lines

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let discount = 0;
```
dimqee, line 7: 3. [Checklist] Prefer const over let where possible, to avoid unintentional changes.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+
+  if (days >= 7) {
+    discount = 50;
+  } else if (days >= 3) {
```
dimqee: 4. [Checklist] Prefer if with return over if else to simplify later conditions.

dimqee (changes requested):
Гарна робота! 
Трішки треб зарефакторити код :)

src/calculateRentalCost.js
nataliia-v: I change discount variable, in this case we can use only let

src/calculateRentalCost.js
nataliia-v: why in this case you think is better use only if() but not  else if ()?
I think  else if () is better because I count discount like in switch case. 
This rule it up to you. It is not restriction

src/calculateRentalCost.js
nataliia-v: updated ✅

dimqee (approved):
Greate job!

VenomMB (approved):

PR #5235, opened 2025-02-10
add solution

glebanty19:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  const cost = 40;
+  const DAY_WITHOUT_DISCOUNT = 2;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const MIDDLE_TERM_DISCOUNT = 20;
+
+  if (days <= DAY_WITHOUT_DISCOUNT) {
```
radykhovska: if you write your conditions from bigger condition to smaller you will need them only 2.

consider this solution:
```
if (days >= LONG_TERM) {
  }

  if (days >= AVERAGE_TERM) {
    return basePrice - AVERAGE_TERM_DISCOUNT;
  }

  return basePrice;
```

radykhovska (changes requested):
Good job!
Please consider alternative solution

radykhovska (approved):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const costPerDay = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const MIDDLE_TERM = 3;
+  const MIDDLE_TERM_DISCOUNT = 20;
```
alexander-ignatow, line 11: Makes sense to move this constants outside the function body

alexander-ignatow (approved):
👍

PR #5245, opened 2025-02-10
Solution

OksanaKuziv:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRate = 40;
```
Iwantschenko, line 7: For const variables it is better to use this name format 'DAILY_RATE'

Iwantschenko (approved):

PR #5248, opened 2025-02-11
add solution to task

VladRoz95:
Solution to task 'Transportation on Vacation'


MariiaKapustkina (approved):
It's look good! I'm not sure but you may not have used to "Term" just 'days'. Also you have mistake in word "Turm". Everything else looks like good)

PR #5252, opened 2025-02-11
add task JS

abaskova:


Olehgoon093:
Все добре

Olehgoon093 (approved):

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const RENT_CAR_COST_PER_DAY = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const MIDDLE_TERM = 6;
+  const SHORT_TERM_DISCOUNT = 20;
```
alexander-ignatow, line 12: Where possible keep such constants outside the function body

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+  let totalCost = RENT_CAR_COST_PER_DAY * days;
```
alexander-ignatow, line 13: why not 
```suggestion
  const totalCost = RENT_CAR_COST_PER_DAY * days;
```
?

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+
+  if (days >= LONG_TERM) {
+    totalCost -= LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM && days <= MIDDLE_TERM) {
+    totalCost -= SHORT_TERM_DISCOUNT;
+  }
+
+  return totalCost;
```
alexander-ignatow, line 21: you can get fix ai comment and make your code cleaner by adding the return right from the if statement:

```suggestion
  if (days >= LONG_TERM) {
    return totalCost - LONG_TERM_DISCOUNT;
  } 
  
  else if (days >= SHORT_TERM) {
    return totalCost - SHORT_TERM_DISCOUNT;
  }
  
  return totalCost;
```

Thus you can make `totalConst` constant

alexander-ignatow (approved):
No need to fix your solution - but here are some potential improvements to your code

src/calculateRentalCost.js
alexander-ignatow: As a bonus - you don't really need MIDDLE_TERM with such fix

PR #5259, opened 2025-02-12
solution

Wasd227:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const baseCost = 40;
+  const discountFor3Days = 20;
+  const discountFor7Days = 50;
+  const mediumTerm = 3;
+  const longTerm = 7;
```
alexander-ignatow, line 11: Such constant have to be outside the function body

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+
+  if (days >= longTerm) {
+    return baseCost * days - discountFor7Days;
+  } else if (days >= mediumTerm) {
+    return baseCost * days - discountFor3Days;
+  }
```
alexander-ignatow, line 17: No need to use `else if` here - as long as you have returns from inside the blocks you may keep it more simple:

```suggestion
  if (days >= longTerm) {
    return baseCost * days - discountFor7Days;
  } 
  
  if (days >= mediumTerm) {
    return baseCost * days - discountFor3Days;
  }
```

alexander-ignatow (commented):
Let's move on, but please, mind my comments

One more important thing - for the future please use meaningful commit names

alexander-ignatow (approved):
Let's move on, but please, mind my comments

One more important thing - for the future please use meaningful commit names

PR #5271, opened 2025-02-13
Tests

andrei-pcsantos:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days === 7) {
```
DanielCamposDevX: Remember to not use literals in expressions, instead use constants explaining this values example:
const FULL_WEEK = 7
if(days === FULL_WEEK)
Do that for the rest of the code

DanielCamposDevX (changes requested):
Remember to not use literals in expressions, instead use constants explaining this values example:
const FULL_WEEK = 7
if(days === FULL_WEEK)
Do that for the rest of the code

DanielCamposDevX (approved):

PR #5343, opened 2025-03-04
Solution

AnastasiiaOlina2025:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const basePrice = days * 40;
```
olantonenko: 2. [checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
BAD EXAMPLE:
if (numberOfDays >= 7) {
  return basePrice - 50;
}
GOOD EXAMPLE:
const LONG_TERM = 7;

}
and for the next lines

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+
+  if (days < 3) {
+    return basePrice;
+  } else if (days >= 3 && days <= 6) {
```
olantonenko: 5.[checklist] DON'T add else after if with return - the code after it won't be executed anyway.
BAD EXAMPLE:
if (condition) {
  return x;
} else {
  return y;
}
GOOD EXAMPLE:
if (condition) {
  return x;
}

return y;

olantonenko (changes requested):
Please fix your code:)

src/calculateRentalCost.js
AnastasiiaOlina2025: Done

src/calculateRentalCost.js
AnastasiiaOlina2025: Done

PR #5351, opened 2025-03-04
Solution Implement calculateRentalCost function with discounts

VikaDora:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let total = days * 40;
```
TetianaMate: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead, use constants explaining this values

and for the next lines

TetianaMate (changes requested):
Good job!!!
Please fix code

PR #5353, opened 2025-03-04
Calculate rental code

Tanialvk22:


Niklesov (changes requested):
Its not an answear, you should do it with for() loop. Returning exact test values will not work with other "days" values. What would happed if I would test this code with 8 days value?

PR #5359, opened 2025-03-04
Solution

Viktoria11Slichenko:


Kiborg1-git:
GJ

Kiborg1-git (approved):
GJ
But you can write 'return only once after function. And makes some let for 'day'

PR #5360, opened 2025-03-04
Solution

IlliaYeskov:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,16 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const dailyRate = 40;
+  let totalCost = days * dailyRate;
+
+  if (days >= 7) {
```
AndrewRadz: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.
And for the next lines

src/calculateRentalCost.js
```diff
@@ -5,6 +5,16 @@
...
+    totalCost -= 50;
+  } else if (days >= 3) {
```
AndrewRadz: 4. [Checklist] Prefer 'if' with 'return' over 'if else' to simplify later conditions.

AndrewRadz (changes requested):
GJ, but need some fixes.

djtyrf312 (approved):
Good job!

PR #5363, opened 2025-03-04
Solution

viktorvmm:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,14 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
-}
+  let total = days * 40;
```
djtyrf312: I do not understand what means 40, 7, 50, 3, 20? create constants with understandable names for these numbers

djtyrf312 (changes requested):

natalliiaa (approved):

PR #5364, opened 2025-03-04
Solution

RomanchukOleksandr:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,17 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
```
djtyrf312, line 7: I do not understand what 7, 40, 50 etc means. Create a variables with understandable names for these numbers

djtyrf312 (changes requested):
check comments and pls try to remove changed package and package-lock files from your pr

src/calculateRentalCost.js
RomanchukOleksandr: Added constants instead of magic numbers. Removed changes from package JSON

natalliiaa (approved):
Approved
Do not push package.json and package-lock.json to GitHub next time

PR #5373, opened 2025-03-05
Solution

DJPoMaHbl4:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let totalCost = days * 40;
```
Tanialvk22: Check [checklist]: 2. Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

and for the next lines

src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
+
+  if (days >= 3 && days < 7) {
```
Tanialvk22: Check [checklist]: 4. Prefer if with return over if else to simplify later conditions.

Tanialvk22 (changes requested):
Please check comments

PR #5411, opened 2025-03-12
Solution

ShramkoOleksandr:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRate = 40;
+  let totalCost = days * dailyRate;
+
+  if (days >= 7) {
```
natalliiaa: Create variables for all values that could be changed:
const LONG_TERM = 7;

}

fix it where it is needed

natalliiaa (changes requested):

natalliiaa (approved):
good job

PR #5413, opened 2025-03-12
Solution

LiDemerdzhy:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const baseCost = days * 40;
+  const discountOne = 20;
+  const discountTwo = 50;
```
Oleksii-Siviolov, line 9: Not the best naming for the variables, but let it be.

Oleksii-Siviolov (approved):
Looks good.

src/calculateRentalCost.js
LiDemerdzhy: I couldn't think of something else that would be universal and understandable. Will try to replace with better ones

Stepico (approved):
Good job, but the naming is still not the best, you could call them smallDiscount and bugDiscount, or just like that discount20 and discount50

PR #5432, opened 2025-03-18
Solution

Ihcaniso:


Ihcaniso (commented):
I have made the following updates to the code based on the review comments:

1. **Reordered the condition checks**:
   - I moved the `if (days >= LONG_TERM)` condition **before** the `if (days >= SHORT_TERM)` condition to ensure that the long-term discount is applied correctly when renting for 7 or more days.
   - This ensures that the **long-term discount** is applied first, and the **short-term discount** is applied only if the rental duration is less than 7 days but more than 2 days.

2. **Retested the solution** to ensure that the correct discount is applied for both short and long-term rentals.

Please review the changes and let me know if further modifications are needed. Thanks!

Ihcaniso (commented):
I have made the following updates to the code based on the review comments:

Reordered the condition checks:

I moved the if (days >= LONG_TERM) condition before the if (days >= SHORT_TERM) condition to ensure that the long-term discount is applied correctly when renting for 7 or more days.
This ensures that the long-term discount is applied first, and the short-term discount is applied only if the rental duration is less than 7 days but more than 2 days.
Retested the solution to ensure that the correct discount is applied for both short and long-term rentals.

Please review the changes and let me know if further modifications are needed. Thanks!

PR #5463, opened 2025-03-26
solved

aket0r:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+
+  if (days >= LONG_TERM) {
+    return PRICE_PER_DAY * days - LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= MIDDLE_TERM) {
+    return PRICE_PER_DAY * days - MIDDLE_TERM_DISCOUNT;
+  }
+
+  return PRICE_PER_DAY * days;
```
volodymyr-soltys97, line 21: You can create a variable for base price and use it everywhere
```suggestion
  const PRICE_PER_DAY = 40;
  const BASE_PRICE = PRICE_PER_DAY * days;

  if (days >= LONG_TERM) {
    return BASE_PRICE - LONG_TERM_DISCOUNT;
  }

  if (days >= MIDDLE_TERM) {
    return BASE_PRICE - MIDDLE_TERM_DISCOUNT;
  }

  return BASE_PRICE;
```

volodymyr-soltys97 (approved):
Good job 👍

PR #5471, opened 2025-03-27
Solution

Ruslan-oss-max0:


src/calculateRentalCost.js
```diff
@@ -1,10 +1,27 @@
+'use strict';
+
 /**
  * @param {number} days
  *
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyCarRentalCost = 40;
```
anastasiiavorobiova, line 9: If you have a constant with some particular value it's better to use uppercase letters (it's code convention)
`DAILY_RENTAL_COST` etc.

anastasiiavorobiova (approved):

PR #5509, opened 2025-04-06
done task

BogdanBondar-fw:


RTSaitama:
checklist prohibits to use else-if construction. better, use if-return constructions only.

src/calculateRentalCost.js
```diff
@@ -5,12 +5,14 @@
  */
 function calculateRentalCost(days) {
   const pricePerDay = 40;
+  const smallDiscount = 20;
+  const bigDiscount = 50;
   let totalCost = days * pricePerDay;
```
RTSaitama: we have comment to use const instead of let if we can.
let totalCost = days * pricePerDay;
you can use 'const totalCost = days * pricePerDay; ' here.

src/calculateRentalCost.js
```diff
@@ -5,12 +5,14 @@
...
 
   if (days >= 3 && days <= 6) {
-    totalCost -= 20;
+    totalCost -= smallDiscount;
   } else if (days >= 7) {
```
RTSaitama: checklist prohibits to use else-if construction. better, use if-return constructions only.

DON'T add else after if with return - the code after it won't be executed anyway.

RTSaitama (changes requested):
Well Done! Fix plz this 2 moments.

PR #5510, opened 2025-04-06
added solution

RTSaitama:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,23 @@
  *
  * @return {number}
  */
+
 function calculateRentalCost(days) {
-  // write code here
+  const rentPerDay = 40;
+  const rentPeriod = days;
+  const shortTerm = 3;
+  const LongTerm = 7;
```
BudnikOleksii: Constants that never changes should be in UPPER_CASE and you can move them outside the function

src/calculateRentalCost.js
```diff
@@ -3,8 +3,23 @@
...
+  let discount = 50;
```
BudnikOleksii: Prefer const over let where possible, to avoid unintentional changes.

src/calculateRentalCost.js
```diff
@@ -3,8 +3,23 @@
...
+
+  if (rentPeriod < LongTerm) {
+    discount = 20;
```
BudnikOleksii: Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
image

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution

BudnikOleksii (changes requested):
.

src/calculateRentalCost.js
RTSaitama: @BudnikOleksii ty for review, I haven't come across this rule yet.  Done. (Hope so)

src/calculateRentalCost.js
```diff
@@ -3,8 +3,25 @@
  *
  * @return {number}
  */
+
+const SHORTTERM = 3;
+const LONGTERM = 7;
+const rentperday = 40;
```
BudnikOleksii:
```suggestion
const SHORT_TERM = 3;
const LONG_TERM = 7;
const RENT_PER_DAY = 40;
```
Constants should be in UPPER_SNAKE_CASE

src/calculateRentalCost.js
BudnikOleksii: > Prefer const over let where possible, to avoid unintentional changes.

Not fixed

src/calculateRentalCost.js
BudnikOleksii: Not fixed

BudnikOleksii (changes requested):
A few comments still not fixed, check them

src/calculateRentalCost.js
RTSaitama: @BudnikOleksii  check plz, remade logic with no let

src/calculateRentalCost.js
```diff
@@ -1,10 +1,23 @@
...
- */
+const SHORT_TERM = 3;
+const LONG_TERM = 7;
+const RENT_PER_DAY = 40;
+const LONG_TERM_DISCOUNT = 50;
+const SHORT_TERM_DISCOUNT = 20;
+
 function calculateRentalCost(days) {
-  // write code here
+  const rentPeriod = days;
```
ArturErmolenko: why? you can use `days`
```suggestion
```

src/calculateRentalCost.js
```diff
@@ -1,10 +1,23 @@
...
+    return RENT_PER_DAY * rentPeriod - LONG_TERM_DISCOUNT;
+  }
+
+  if (rentPeriod >= SHORT_TERM) {
+    return RENT_PER_DAY * rentPeriod - SHORT_TERM_DISCOUNT;
+  }
+
+  if (rentPeriod < SHORT_TERM) {
+    return RENT_PER_DAY * rentPeriod;
+  }
```
ArturErmolenko: Function should always have return
What happened if all ifs are false?

Screenshot 2025-04-24 at 14 49 00


```suggestion
    return RENT_PER_DAY * rentPeriod;
```

ArturErmolenko (changes requested):
Almost done 🔥 Let's make your code better ;)

BudnikOleksii (approved):
Good work 👍

PR #5520, opened 2025-04-08
Update calculateRentalCost.js

mslol21:


mslol21:
  let price = days * 40;

  if (days >= 7) {
    price -= 50;
  } else if (days >= 3) {
    price -= 20;
  }

  return price;
}

PR #5539, opened 2025-04-15
Solution

r-marianna:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = days * 40;
+  const longTerm = 7;
+  const longTermOff = 50;
+  const middleTerm = 3;
+  const middleTermOff = 20;
+
```
YesupenkoVladislav, line 12: Constants should be declared at module level (not inside function)  Otherwise looks good!

YesupenkoVladislav (changes requested):

src/calculateRentalCost.js
r-marianna: > Constants should be declared at module level (not inside function) Otherwise looks good!

Generally speaking, yes, but for this one particular example, it doesn't matter. Even in the mentor's solution, we have almost the same answer
https://i.imgur.com/M1RuWYk.png

src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const rentCost = days * PRICE_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return rentCost - LONG_TERM_OFF;
+  }
+
+  if (days >= SHORT_TERM) {
+    return rentCost - SHORT_TERM_DISCOUNT;
```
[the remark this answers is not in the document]
r-marianna: Are you sure?)))

natalliiaa (approved):
Good job!

PR #5543, opened 2025-04-16
Refactor: replace magic numbers with constants in rental cost calculation

YesupenkoVladislav:
Replaced hardcoded values with descriptive constants (DAILY_RATE, SHORT_TERM_DISCOUNT, etc.)
Simplified conditional logic for discounts


natalliiaa (changes requested):
Your solution looks good, but you've pushed too many changed files. Your PR should include only the file you've changed by yourself (src/calculateRentalCost.js), all other files should be removed from the PR
Please fix it

natalliiaa (changes requested):
17 files are still present in your PR instead of 1 file
Your PR should contain calculateRentalCost.js only

PR #5545, opened 2025-04-16
Solution

a3ex77:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+  const longTerm = 7;
+  const shortTerm = 3;
+
+  if (days >= longTerm) {
+    return oneDayCost * days - discountSevenDays;
+  } else if (days >= shortTerm && days < longTerm) {
+    return oneDayCost * days - discountThreeDays;
+  } else {
+    return oneDayCost * days;
+  }
```
ArturErmolenko: it's hard to read code with else else if
can we get rid of else
smth like

```
if() {
return
}

if() {
return
}

return
```

ArturErmolenko (approved):

PR #5562, opened 2025-04-21
Solution calculateRentalCost task

AndSLASH:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,26 @@
...
+  const discountMax = 50;
+
+  const minDays = 3;
+  const maxDays = 7;
+
+  const totalCost = basePrice * days;
+
+  let discountPrice = 0;
+
+  if (days >= minDays && days < maxDays) {
```
nbryk, line 16: 4.  [checklist] Prefer if with return over if else to simplify later conditions.

Try changing these conditions with:
 if (condition) {
 `return` the-result-of-the-solution;
 }
 
 if (different condition) {
 `return` another-solution;
 }

nbryk (changes requested):
Good job!
Please fix your code.

nbryk (approved):
Good job!
You can move forward confidently.

BudnikOleksii (approved):

PR #5563, opened 2025-04-21
solution

nbryk:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,24 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const numberOfDays = days;
+  const basePrice = 40;
```
yelyzavetaz: should use SCREAMING_SNAKE_CASE for constants that represents fixed values.

yelyzavetaz (changes requested):
The logic is correct and clean,  just a minor naming adjustment needed. Well done!

yelyzavetaz (approved):
Great work - keep it up! The code is clean and functional.
Thanks for your contribution ฅ^>⩊<^ ฅ

Moroz-Dmytro (approved):

PR #5566, opened 2025-04-22
Solution

Dikanb:


Ayume-beep (approved):
It seems to me that everything could be done simpler here without using variables. But everything works and that's the main thing!🎉

PR #5568, opened 2025-04-22
Solution

OksanaKuziv:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
...
  */
+const COST_PER_DAY = 40;
+const LONG_TERM = 7;
+const LONG_TERM_DISCOUNT = 50;
+const SHORT_TERM = 3;
+const SHORT_TERM_DISCOUNT = 20;
+
 function calculateRentalCost(days) {
-  // write code here
+  const result = days * COST_PER_DAY;
```
hryhoriikorsun, line 13: If you use `const`, you sdould naming in CapsLock like: "RESULT". Or you can just use `let`.

hryhoriikorsun (changes requested):
Good job!👍 Please fix your code

src/calculateRentalCost.js
OksanaKuziv: Hey! Thanks for your feedback.
I just wanted to clarify a few things about const and naming conventions:

Using const doesn't automatically mean that the variable name should be in all caps.

All-caps  is usually reserved for global, fundamental constants — values that are truly fixed and universal.

In this case, result is a local, intermediate value inside the function. It's common practice to use camelCase for such variables, even if they are declared with const.

const simply ensures the binding won't change, but it doesn't require a specific naming style.

Also, there is no need to switch to 'let' unless we expect the value to be reassigned, which is not the case here.

So using const result in camelCase is both correct and aligned with best practices.

But thanks again for pointing it out — always good to double-check and discuss coding style!

ArturErmolenko (approved):
Well done 🔥

PR #5571, opened 2025-04-22
solution

Ayume-beep:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
```
DenisPitsul: Please, read check list: https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md
2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

and for next lines.

DenisPitsul (changes requested):

src/calculateRentalCost.js
```diff
@@ -1,10 +1,19 @@
-/**
- * @param {number} days
- *
- * @return {number}
- */
+const cost = 40;
```
DenisPitsul, line 1: Move all constants inside method, because it doesn't use anywhere except in this method.

src/calculateRentalCost.js
```diff
@@ -1,10 +1,19 @@
...
+const discount = 20;
+const bigDiscount = 50;
```
DenisPitsul, line 3: Use Upper snake case naming for constants. For example:
From bigDiscount to BIG_DISCOUNT.

DenisPitsul (changes requested):

PR #5572, opened 2025-04-22
Add task solution

hryhoriikorsun:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,26 @@
...
+  let counter = 0;
+  const PRICE_ONE_DAY = 40;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCAUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCAUNT = 50;
+
+  for (let i = 0; i < days; i++) {
+    counter += PRICE_ONE_DAY;
+  }
```
meryam99: 4. Prefer `if` with `return` over `if else` to simplify later conditions.

meryam99 (approved):

PR #5578, opened 2025-04-22
Solution

meryam99:


src/calculateRentalCost.js
```diff
@@ -6,10 +6,18 @@
 function calculateRentalCost(days) {
   const price = 40 * days;
```
DanilAt191, line 7: 2. [checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

Create a new constant variable (like LONG_TERM or SHORT_TERM_DISCOUNT) with a value of `40` and then multiply it by the `days` in the `price`

DanilAt191 (approved):
In general, good job. The code writing is great

PR #5584, opened 2025-04-22
Solution

koliiis:


jasialazyrka (approved):
Nice work!

Maybe consider declaring variables in the order they are used in the function. This could make the flow a bit easier to follow.

PR #5591, opened 2025-04-23
added task solve

torry455:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const oneDayCost = 40;
+  let discount = 0;
```
BudnikOleksii, line 13: Prefer const over let where possible, to avoid unintentional changes.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
...
+
+  if (days >= 3 && days <= 6) {
+    discount = 20;
+  }
+
+  if (days >= 7) {
+    discount = 50;
+  }
```
BudnikOleksii: - Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
image
-

BudnikOleksii (changes requested):
Good work 👍 
But don't forget about checklist, left a few comments in accordance

BudnikOleksii (approved):

PR #5592, opened 2025-04-23
add func

olegll:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,22 @@
...
  */
+const discountSmall = 20;
+
+const discountBig = 50;
+
+const rent = 40;
+
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
```
NicssLucik02, line 13: Don't use literals in expressions (HARDCODE). Instead use constants explaining this values.   
const LONG_TERM = 7;

NicssLucik02 (changes requested):
All good, but the checklist requires creating a constant for the number of rental days.

PR #5594, opened 2025-04-24
Solution-2

valchuk-ivan:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,18 @@
  *
  * @return {number}
  */
+// //
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRate = 40;
+  let total = days * dailyRate;
+
+  if (days >= 7) {
```
meryam99, line 11: 2. Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
const LONG_TERM = 7;

src/calculateRentalCost.js
```diff
@@ -3,8 +3,18 @@
...
+    total -= 50;
+  } else if (days >= 3) {
```
meryam99, line 13: Prefer `if` with `return` over `if else` to simplify later conditions.

meryam99 (approved):
Good job! Seems like you forgot a little bit to go over the checklist, but the solution is right. Hope AI will approve it. Anyways, it's not hard to correct the mistakes and move on :)

PR #5616, opened 2025-04-29
solution

Unlie9:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+      break;
+    }
+
+    case days >= 7: {
+      price -= 50;
+      break;
+    }
+  }
+
+  return price;
```
BudnikOleksii, line 21: Please, fix your code in accordance with a [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md) 

`switch/case` is a bad option, it fits when you have fixed conditions, also it should always have a `default` case

BudnikOleksii (changes requested):
Please, fix your solution in accordance with a [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

PR #5749, opened 2025-06-25
add solution

GSTVdev:


pedro-ruas (commented):
Daniel, your logic is valid and it works. but there is a topic we should point out for you to fix.
Whenever we pass numbers different than 1 or 0, we should specify the origin of that number. When not, we call it "Magic Numbers", a number that makes the code work, but nobody knows where it comes from.

Fixing it is very simple, take those numbers and save them in variables that make sense. For example:
```
const costPerDay = 40;
const total = days * costPerDay;
```
Now we can say that the number 40 is the cost for a day of rent.

pedro-ruas (changes requested):
Fix the magic numbers form my previous comment

pedro-ruas (changes requested):
Why did you fix only ONE of the magic numbers?
Where does the numbers 50, 3, 7 and 20 come from?
If the code is fully in english, why are there new variables written in portuguese?

PR #5762, opened 2025-07-01
feat: add solution

BlackUserSide:


alexander-ignatow (approved):
Good, but the number literals you use are the "magic numbers".
You should avoid them by using properly named constants, to explain the meaning of each number.

For instance:
```
 const DISCOUNT_3_DAYS = 20;
 const DISCOUNT_7_DAYS = 50;
// and so on
```

PR #5769, opened 2025-07-05
Solution

MikeKorotych:


pbahrynovskyi (changes requested):
Everything works. In the topic they demanded to create special variables for numbers of the days, so I advice you to add theese names.

pbahrynovskyi (approved):
Everything works great. Variables are added.

PR #5780, opened 2025-07-08
Solution

yanaovcharyk:


DavidHurmanchuk (approved):
Great job! But I think you should write constants like in the example. Here is an example =
yours : oneDayPrice
example : ONE_DAY_PRICE

PR #5782, opened 2025-07-08
add solution

Chips666:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,28 @@
...
+    price = PRICE_FOR_DAY * days - LONG_DAY_DISCOUNT;
+
+    return price;
+  }
+
+  if (days > SHORT_TERM) {
+    price = PRICE_FOR_DAY * days - SHORT_DAY_DISCOUNT;
+
+    return price;
+  } else {
```
jeolinna, line 24: [checklist] 5. DON'T add `else` after `if` with `return` - the code after it won't be executed anyway.

jeolinna (approved):
GOOD JOB!!!!!🎉🎉🎉

PR #5788, opened 2025-07-09
Solution

bebelekele:


derpedcatto (approved):
Everything looks good.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
+  const SHORT_TERM_DISCOUNT = 20;
+  const RENT_COST = 40;
+
+  if (days >= LONG_TERM) {
+    return days * RENT_COST - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
+    return days * RENT_COST - SHORT_TERM_DISCOUNT;
+  }
+
+  return days * RENT_COST;
```
derpedcatto, line 19: The only thing I would change is to separate base `days * RENT_COST` logic to it's own variable (like `totalCost`).

In the future, if you would need to change base `totalCost` calculation logic, you would have to each `if` statement (and `return`) and change it manually.

But with `totalCost` as a separate variable, you only need to write this logic once, and use `if` statements to only **modify** your `totalCost`.

derpedcatto (changes requested):
Everything looks solid, but I would advise on changes left in my comment.

src/calculateRentalCost.js
derpedcatto: Sorry about the second comment, I accidentally sent my comment without finishing my additional through and didn't know how to edit my comment with code changes included 🙏

derpedcatto (approved):

PR #5790, opened 2025-07-09
Solution

derpedcatto:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const DAILY_RENTAL_PRICE = 40;
+  const DISCOUNT_SHORT_TERM = 20;
+  const DISCOUNT_LONG_TERM = 50;
+  const LONG_TERM = 7;
+  const SHORT_TERM = 3;
+  let totalCost = 0;
```
isobol-js, line 12: "let totalCost = 0;"

unnecessary line of code. This line in your code is unnecessary because immediately after it you assign a new value.

isobol-js (approved):

PR #5791, opened 2025-07-09
completed task

isobol-js:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  const price = PRICE_PER_DAY * days;
+
+  if (days >= LONG_TERM) {
+    return price - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
```
AARON-FOX: 5. DON'T add else after if with return - the code after it won't be executed anyway.

Good job, man!👍
But you don't need to use `if else` construction if you use `return`. Just use `if` for every condition.

AARON-FOX (changes requested):

PR #5792, opened 2025-07-09
task done

PopelinovDanilo:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,30 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const costs = 40;
+  let rentalCost = 0;
```
maxrori228: 1. Make names descriptive, so they explain the purpose of variables and functions.
2. Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
- BAD EXAMPLE:
    ```js
    if (numberOfDays >= 7) {
      return basePrice - 50;
    }
    ```
    - GOOD EXAMPLE: 
    ```js
    const LONG_TERM = 7;

    }
    ```
    
   check the file checklist.md and follow the steps from there

maxrori228 (changes requested):

PR #5796, opened 2025-07-09
Complete the task 

YevheniiKa:
Add solution to the task


romkas228 (approved):
1. I would remove '&& days <= LONG_TERM' as it has no effect on the result. (line 19)
2. Also i would replace 'return (finalPrice -= LONG_TERM_DISCOUNT)'' 
by 'return finalPrice - LONG_TERM_DISCOUNT'

PR #5797, opened 2025-07-09
add task solution

vladyslav-antoniv:


src/calculateRentalCost.js
vladyslav-antoniv, line 1: 1. [checklist] Make names descriptive, so they explain the purpose of variables and functions.

src/calculateRentalCost.js
```diff
@@ -5,6 +5,15 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  let cost = days * 40;
+
+  if (days >= 7) {
```
vladyslav-antoniv: 2. [checklist] Prefer if with return over if else to simplify later conditions.

src/calculateRentalCost.js
```diff
@@ -5,6 +5,15 @@
...
+    cost -= 50;
+  } else if (days >= 3) {
```
vladyslav-antoniv: 3. [checklist] DON'T add else after if with return - the code after it won't be executed anyway

vladyslav-antoniv (commented):
good

vladsulyma (changes requested):
The code does not meet the following requirements:
1. Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
OPTIONAL
3. Prefer const over let where possible, to avoid unintentional changes.
4. Prefer if with return over if else to simplify later conditions.

Please, update your code:
1.  Please avoid hardcoded values and use named constants for better readability and maintainability.
OPTIONAL
2. You should use const on line 8, as the value is not reassigned afterwards if you follow all requirements.
3. Instead of reassining cost use "return cost - <your_value>".

src/calculateRentalCost.js
```diff
@@ -5,6 +5,23 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const pricePerDay = 40;
```
vladsulyma, line 8: const format should be in SCREAMING_SNAKE_CASE

vladsulyma (approved):
Great gob!
The code follows all requirements, except for the case of consts. Keep an eye on it next time!

PR #5798, opened 2025-07-09
Solution

vladsulyma:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
```
DmitriyALEX, line 7: remove this line " // write code here"

DmitriyALEX (changes requested):
Plz, fix code

src/calculateRentalCost.js
vladsulyma: This line has already been removed from the code. Please, look closely, there is "-" before the line. It means that this part of the code was removed

DmitriyALEX: ok, your solution is great

DmitriyALEX (approved):

PR #5801, opened 2025-07-09
Solution

Oleksandr-Mykyteichuk:


Ann-ovocheviy (approved):
The code is written correctly and concisely, meeting all of the requirements. Works as it should.

PR #5821, opened 2025-07-14
Solution

DidZhara:


Ann-ovocheviy:
i'm not sure if i should write something here by now, but great job

Ann-ovocheviy (approved):
Great job! There's some things you can consider in the future:
1. In JavaScript it's common practice to use  all uppercase letters, words separated by underscores LIKE_THAT for constants whose values do not change during program execution.
2. Here, 40 is your base price, and what you're using as basePrice is your price per day. It would be better to use 40 as a separate variable.

PR #5845, opened 2025-07-23
Solution

N3m0o:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,30 @@
...
+  if (!Number.isInteger(days)) {
+    throw new Error('Invalid value');
+  }
+
+  const totalPay = days * 40;
+  let discount;
+
+  if (days >= 3 && days < 7) {
+    discount = 20;
+  } else if (days >= 7) {
```
mkyloa, line 16: try using return statements in `if` instead of nested if/else. It will make your code more readable

src/calculateRentalCost.js
```diff
@@ -4,7 +4,30 @@
...
+    discount = 50;
+  }
+
+  let result;
```
mkyloa, line 20: you can return your result in if statement instead of reassigning it

mkyloa (approved):
awesome! but let's refactor and improve this code even further 🚀 
Moreover, let's not commit package-lock files and test templates

PR #5874, opened 2025-08-02
rental prices

serhash:


etojeDenys (changes requested):
You should use variables instead of hard-coded numbers, such as 50 and 20
For example:
SMALL_TERM_DISCOUNT = 20;
LONG_TERM_DISCOUNT = 50;

vitalii-bondarenko-ua (approved):
looks good

PR #5894, opened 2025-08-12
done

MaksimShe:


package-lock.json
mkyloa, line 1: let's omit committing package-locks in further PRs

src/calculateRentalCost.js
```diff
@@ -4,7 +4,13 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
+    return days * 40 - 20;
+  } else {
+    return days * 40;
+  }
```
mkyloa, line 13: awesome and simple solution 👍

src/calculateRentalCost.js
```diff
@@ -4,7 +4,13 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days >= 7) {
+    return days * 40 - 50;
+  } else if (days >= 3) {
```
mkyloa, line 9: you can omit using else statements if you return the result in if statements
```suggestion
  if (days >= 3) {
```

mkyloa, line 9: how bout that? 😄 
```suggestion
  if (days > 2) {
```

mkyloa (approved):

PR #5897, opened 2025-08-13
add solution

ProninaMariia:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
-  // write code here
+  if (days === 3) {
+    return 100;
+  }
+  if (days === 6) {
+    return 220;
+  }
+  if (days === 7) {
+    return 230;
+  }
```
Anton-Kuchmasov: You've added a hard-code values that (possibly) will satisfy the tests.
However, there is no solution here - the main goal of this task is to make & develop some calculations.

Here you go! Some tips for you to good start:

`  const DAILY_RATE = 40;
  const LONG_TERM = 7;
  const SHORT_TERM = 3;

  const price = days * DAILY_RATE;`
  
  
  After that, think how result of our function depends on days number.

Anton-Kuchmasov (changes requested):
Please re-write your solution with no hardcode values.

Comment below can help you deal with it

anastasiiavorobiova (changes requested):
Make sure to update your code accordingly to the checklist

vitalii-bondarenko-ua (changes requested):
You need to fix prev comments and pass all tests.
Screenshot 2025-10-10 at 16 40 00
Screenshot 2025-10-10 at 16 40 30

If you have some questions - feel free to ask in chat.

Anton-Kuchmasov (approved):
LGTM!

PR #5946, opened 2025-09-02
_first task_

S-Khraban:


S-Khraban:
Все виконано по чеклисту. 
10/10

src/calculateRentalCost.js
```diff
@@ -1,10 +1,13 @@
...
- * @return {number}
- */
 function calculateRentalCost(days) {
-  // write code here
+  let cost = days * 40;
+
+  if (days >= 7) {
+    cost -= 50;
+  } else if (days >= 3) {
+    cost -= 20;
```
Bohdan-Diatliuk, line 7: Great work! But you need to use variables for the two expressions. For example: days  >= 7 - bad, days >= vacationTime - good. If you don't understand me, please review the checklist.md in this task.

Bohdan-Diatliuk (changes requested):

PR #5960, opened 2025-09-09
Sokution

Pobedasg:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,21 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const totalMinDays = 3;
+  const totalMaxDays = 7;
+  const cost = 40;
+  const discountForThreeDays = 20;
+  const discountForSevenDays = 50;
+
+  if (days < totalMinDays) {
+    return days * cost;
```
MariaAlefirenko: You often use the expression ‘days * cost’.
So why not just create a variable for it?
You may think that it doesn't matter because 'return' is used when the condition is met.
However, in the third case, we will perform the same calculation three times.

MariaAlefirenko (changes requested):
Good work!
I really liked your solution algorithm for this task!
But there is one thing that could be improved.

MariaAlefirenko (approved):
Excellent!
Have a good day)))

PR #5961, opened 2025-09-09
Solution

Maevalar:


src/calculateRentalCost.js
```diff
@@ -5,14 +5,20 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const price = 40;
```
livadin: Use UPPER_CASE for constants to make your code more readable.

src/calculateRentalCost.js
```diff
@@ -5,14 +5,20 @@
...
+  const longTerm = 7;
+  const longTermDiscount = 50;
+
   let totalAmount = 0;
 
-  if (days >= 3 && days < 7) {
-    totalAmount = days * 40 - 20;
+  if (days >= shortTerm && days < longTerm) {
+    totalAmount = days * price - shortTermDiscount;
   } else if (days >= 7) {
```
livadin: 2. [checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
You hardcoded numbers here. Avoid using literal values — use constants instead

src/calculateRentalCost.js
```diff
@@ -5,14 +5,20 @@
...
-    totalAmount = days * 40 - 50;
+    totalAmount = days * price - longTermDiscount;
   } else {
```
livadin: Prefer early return to reduce nested if-else.

src/calculateRentalCost.js
```diff
@@ -5,14 +5,20 @@
  */
 function calculateRentalCost(days) {
   // write code here
+  const price = 40;
+  const shortTerm = 3;
+  const shortTermDiscount = 20;
+  const longTerm = 7;
```
livadin: Variable name is not descriptive enough. Use something like: LONG_TERM_DAYS

livadin (changes requested):
Hi 👋 Good job! Your function works and the logic is mostly correct. 👍 But your code needs a few minor fixes. You can do it! 💪

PR #5970, opened 2025-09-09
add solution

nastiaserkh:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,22 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const normalPrice = 40;
```
pshy4ok: Constants that represent fixed values should be written in UPPER_CASE.

This convention makes it immediately clear that the value is not supposed to change during runtime and is used as a constant rather than a regular variable.

This also applies to the following lines in the code

pshy4ok (approved):
In general, your solution is right, well done.

But you need to make minor changes about the comment I wrote.

pshy4ok (approved):

PR #5999, opened 2025-09-14
add solution

nazkov:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const RENT = 40;
+  const MIDDLE_TERM = 3;
+  const MIDDLE_TERM_DISCOUNT = 20;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
```
Moroz-Dmytro: Would be better to move these variables outside the function

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+
+  if (days >= MIDDLE_TERM && days < LONG_TERM) {
+    return RENT * days - MIDDLE_TERM_DISCOUNT;
+  }
+
+  if (days >= LONG_TERM) {
+    return RENT * days - LONG_TERM_DISCOUNT;
+  }
+
+  return RENT * days;
```
Moroz-Dmytro: You calculate `RENT * days` 3 times, create variable

Moroz-Dmytro (changes requested):
Good work 👍 
But fix small comments and it's done

src/calculateRentalCost.js
```diff
@@ -3,8 +3,24 @@
...
  */
+const RENT_PER_DAY = 40;
+const MIDDLE_TERM = 3;
+const MIDDLE_TERM_DISCOUNT = 20;
+const LONG_TERM = 7;
+const LONG_TERM_DISCOUNT = 50;
+
 function calculateRentalCost(days) {
-  // write code here
+  const RENT = RENT_PER_DAY * days;
```
Moroz-Dmytro, line 13:
```suggestion
  const rent = RENT_PER_DAY * days;
```

We use UPPER_CASE only for static values

Moroz-Dmytro (approved):

PR #6092, opened 2025-10-25
add solution

eduardo5010:


raulriato (approved):
You've done a good job!

If you're interested in another perspective of solution for the task:

You could solve it without having to repeat the `return` 3 times. 

```javascript
if (days <= LONG_TERM && days >= SHORT_TERM) {
  totalCost -= SHORT_TERM_DISCOUNT;
}

if (days >= LONG_TERM) {
  totalCost -= LONG_TERM_DISCOUNT;
}

return totalCost;
```

That way, your code would look cleaner, but your solution is fine, so I'll approve it.

PR #6126, opened 2025-11-05
Sln

de-vasta:
Implements a rental cost calculation in `calculateRentalCost.js` with daily fees and discounts for extended rentals (3+ and 7+ days).


DemonWayne (approved):
This is quite an interesting solution, although it is rather complicated given the requirements, but in my opinion it does not violate them.

PR #6140, opened 2025-11-10
Solution

DemonWayne:


src/calculateRentalCost.js
```diff
@@ -1,10 +1,28 @@
...
  */
-function calculateRentalCost(days) {
-  // write code here
-}
+const calculateRentalCost = (days) => {
+  const rentalCost = days * RENTAL_COST_PER_DAY;
+
+  if (days >= LONG_TERM) {
+    return rentalCost - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
```
dmytro-halieba: 2. [Checklist] DON'T add else after if with return - the code after it won't be executed anyway

dmytro-halieba (changes requested):
Amazing job on the task! Overall, the code looks clean and easy to follow - the constants are well-named, the logic is straightforward, and the function structure is clear and well-organized.

src/calculateRentalCost.js
DemonWayne: @dmytro-halieba I think that this point means not to use **exactly else** and **not else if**, and in this place else if has its own proposal, and we still have a moment when `days >= 0 and <= SHORT_TERM`, so code after this else-if condition still will be executed

PR #6166, opened 2025-11-17
Solution

dmytro-halieba:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,27 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40; // dolars$
+  const longTerm = 7; // days
+  const longTermDiscount = 50; // dolars$
+  const shortTerm = 3; // days
+  const shortTermDiscount = 20; // dolars$
+  let total = days * pricePerDay;
```
ArturErmolenko: why let? there should be const
```suggestion
  const total = days * pricePerDay;
```
total what? naming should be better like totalRentCost or just result ;)

src/calculateRentalCost.js
```diff
@@ -4,7 +4,27 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const pricePerDay = 40; // dolars$
+  const longTerm = 7; // days
+  const longTermDiscount = 50; // dolars$
+  const shortTerm = 3; // days
+  const shortTermDiscount = 20; // dolars$
```
ArturErmolenko: remove comments, code on GH should be clean

ArturErmolenko: also you can group variables like 

```
  const pricePerDay = 40;
 
  const longTerm = 7; 
  const shortTerm = 3;
  
  const longTermDiscount = 50; 
  const shortTermDiscount = 20;
```

src/calculateRentalCost.js
```diff
@@ -4,7 +4,27 @@
...
+  let total = days * pricePerDay;
+
+  switch (true) {
```
ArturErmolenko: no no no, we shouldn't move that way, it's magic string i can't understand why true what does it mean
let's follow checklist

ArturErmolenko (changes requested):
Good start , please double check your solution with [checklist](https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md)

src/calculateRentalCost.js
dmytro-halieba: Fixed - changes implemented in commit 3e47e5e

src/calculateRentalCost.js
dmytro-halieba: Fixed - changes implemented in commit 3e47e5e

src/calculateRentalCost.js
dmytro-halieba: Fixed - changes implemented in commit 3e47e5e

ArturErmolenko (approved):
Well done 🔥

PR #6186, opened 2025-11-29
transportation on vacation

maximtsyrulnyk:
done "transportation on vacation" task


maximtsyrulnyk:
Edited and updated "transportation on vacation" task code

PR #6222, opened 2025-12-08
Solution

mark-chep:


NastyBaster:
Your solution is correct and follows the checklist rules, although there is one minor extra point in the logic of the code. In lines 13–15, an redundant check is performed, and without it your code would still return 0. So you can reduce the number of lines by removing this part of the code.

mark-chep:
@NastyBaster Please check my last commit, approve or request changes.

NastyBaster (approved):

PR #6223, opened 2025-12-08
Solution

dianakerekesha:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const dailyRate = 40;
+  let total = dailyRate * days;
+
+  if (days >= 7) {
```
Starkillerr: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
same about the next lines

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
...
+    total -= 50;
+  } else if (days >= 3) {
```
Starkillerr: 3. [Checklist] Prefer `if` with `return` over `if else` to simplify later conditions.

Starkillerr (changes requested):

PR #6226, opened 2025-12-08
Add solution

nykyforov-dan:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,19 @@
...
-  // write code here
+  const DAY_COST = 40;
+  const LONG_TERM = 7;
+  const LONG_TERM_DISCOUNT = 50;
+  const SHORT_TERM = 3;
+  const SHORT_TERM_DISCOUNT = 20;
+
+  if (days >= LONG_TERM) {
+    return days * DAY_COST - LONG_TERM_DISCOUNT;
+  } else if (days >= SHORT_TERM) {
```
dianakerekesha: 5. [Checklist] DON'T add else after if with return - the code after it won't be executed anyway.

dianakerekesha (changes requested):

dianakerekesha (approved):
Nice job!

etojeDenys (approved):
good job

PR #6227, opened 2025-12-08
Solution

torianykhanna:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let price = days * 40;
+
+  if (days >= 7) {
+    price -= 50;
+  } else if (days >= 3) {
```
dianakerekesha: 4. [Checklist]  Prefer if with return over if else to simplify later conditions.

de-vasta (changes requested):
Great job here! The only thing I'd suggest is to avoid *magic numbers* and *hardcoded literals*; it might be harder to maintain and understand. For instance, it's better to extract `40` to a constant like `FEE` or so. That way, we will know what it's for, and in case the value changes, it will change everywhere it is used.

de-vasta (approved):
lgtm

PR #6228, opened 2025-12-09

SenyaKir:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const carCosts = 40;
```
Dimononon: Please use uppercase for constants (SCREAMING_SNAKE_CASE).
Same applies to all consts below

Dimononon: Consider changing const name to more specific(maybe PRICE_PER_DAY or CAR_PRICE_PER_DAY)

Dimononon (approved):
Except variable naming, everything is ok

Dimononon (approved):
Great job 😊

PR #6231, opened 2025-12-09
add task solution

SergiyKusiy:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentForOneDay = 40;
+  let total = days * rentForOneDay;
+
+  if (days >= 7) {
```
SenyaKir: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
and for the next lines

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentForOneDay = 40;
+  let total = days * rentForOneDay;
```
SenyaKir: 1. [Checklist] Make names descriptive, so they explain the purpose of variables and functions.
maybe the variable name TOTAL_COAST would be better

src/calculateRentalCost.js
```diff
@@ -4,7 +4,16 @@
...
+
+  if (days >= 7) {
+    total -= 50;
+  } else if (days >= 3) {
```
SenyaKir: 4. [Checklist] Prefer `if` with `return` over `if else` to simplify later conditions.

SenyaKir (changes requested):
Good job!
Please fix your code

PR #6232, opened 2025-12-09
solution

lilp3ka:


Etwaiz (approved):
agree with the checklist

TwistyD:
An elegant and short solution. The only point is that in the requirements it had to be in UPPERCASE, but you did it in camelCase.

TwistyD (approved):
An elegant and short solution. The only point is that in the requirements it had to be in UPPERCASE, but you did it in camelCase. And there’s also the let/const point.

DanilWeda (changes requested):
Please, view students and AI comments and fix it. 

Many thanks!

DanilWeda (approved):

PR #6326, opened 2026-01-26
Solution

gmarchese93:
- Implemented calculateRentalCost function according to task requirements
- All tests are passing
- Code follows linting rules


gmarchese93:
I've refactored the solution to replace magic numbers with named constants and updated the logic to use early returns, in line with the checklist requirements.

PR #6387, opened 2026-02-23
Solution

MaxShaggy:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
+
+  for (let i = 1; i <= days; i++) {
```
SergGun: The code has an extra loop.
for (let i = 1; i <= days; i++) {
sum += 40;
}
In fact, it is a multiplication let sum = days * 40;

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+    sum += 40;
+  }
+
+  if (days >= 3) {
```
SergGun: You can use else if in this situation

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
  * @return {number}
```
SergGun, line 4: No description of parameters @return {number}

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 0;
```
SergGun: Prefer `const` over `let` where possible, to avoid unintentional changes.

src/calculateRentalCost.js
```diff
@@ -4,7 +4,21 @@
...
+
+  for (let i = 1; i <= days; i++) {
+    sum += 40;
```
SergGun: Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

SergGun (changes requested):
Please fix issues

PR #6452, opened 2026-03-26
Solution

timurradkevic:


linska (approved):
Nice and clean solution 👍

Just a small style suggestion: UPPER_CASE is usually used for global constants, while local variables are typically written in camelCase.

But overall everything looks great 🔥

PR #6467, opened 2026-03-31
Solution

Misharkus:


Palfinger2002 (approved):
Okey. Great job! You use variables with correct values, it's right solution.

PR #6472, opened 2026-03-31
finished task Transportation on vocation

Evgeniy45:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,17 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const totalRentCount = days * 40;
+
+  if (days >= 7) {
```
Dima2702: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

and for the next lines

Dima2702 (changes requested):
Good job! Please fix your code

Dima2702 (approved):

PR #6548, opened 2026-04-28
Solution

Macsoti:


brespect (changes requested):
Good progress, but let's fix Luke comments about constants

brespect (changes requested):
Well done!

brespect (approved):

PR #6581, opened 2026-05-21
fix-calculate-rental-cost

carolsprak:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
...
+  
+  let price = days * 40;
+
+  if (days >= 7) {
+    return price - 50;
+  }
+
+  if (days >= 3) {
+    return price - 20;
+  }
```
carolsprak, line 16: checklist.md #2 Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

const LONG_TERM = 7;

}

src/calculateRentalCost.js
```diff
@@ -4,7 +4,18 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  
+  let price = days * 40;
```
carolsprak, line 8: checklist.md #3 Prefer const over let where possible, to avoid unintentional changes

PR #6689, opened 2026-08-01
Added solution

ggylyon:


src/calculateRentalCost.js
```diff
@@ -1,10 +1,29 @@
+'use strict';
+
 /**
  * @param {number} days
  *
  * @return {number}
  */
+
```
HiljSV, line 8: Minor: this blank line detaches the JSDoc block from the function, so it stops
documenting it. In the template the comment sits right above `function`.

src/calculateRentalCost.js
```diff
@@ -1,10 +1,29 @@
...
+  if (days >= LONG_TERM) {
+    totalPrice -= LONG_TERM_DISCOUNT;
+  }
+
+  if (days >= SHORT_TERM && days < LONG_TERM) {
+    totalPrice -= SHORT_TERM_DISCOUNT;
+  }
+
+  return totalPrice;
```
HiljSV: The logic is correct - I checked 1, 2, 3, 4, 6, 7, 8 and 10 days and every result
matches the task examples. Two checklist points though: `const` over `let` (p. 3)
and `if` with `return` (p. 4).

With early returns you don't need to mutate the total, and the `&& days < LONG_TERM`
guard disappears - the first `return` already handles that case. I also renamed the
daily rate: `RENT_PRICE` reads like the total price of the rental, but it is the
price per day.

```suggestion
  const PRICE_PER_DAY = 40;
  const LONG_TERM = 7;
  const LONG_TERM_DISCOUNT = 50;
  const SHORT_TERM = 3;
  const SHORT_TERM_DISCOUNT = 20;

  const totalPrice = days * PRICE_PER_DAY;

  if (days >= LONG_TERM) {
    return totalPrice - LONG_TERM_DISCOUNT;
  }

  if (days >= SHORT_TERM) {
    return totalPrice - SHORT_TERM_DISCOUNT;
  }

  return totalPrice;
```

HiljSV (changes requested):
Logic is correct - I ran 1, 2, 3, 4, 6, 7, 8 and 10 days and all of them match the
task examples, CI is green as well. Nice job pulling every magic number out into a
named constant.

Two checklist points to fix: `const` over `let` (p. 3) and `if` with `return` (p. 4).
Switching to early returns covers both at once and also removes the extra
`days < LONG_TERM` condition. Left a suggestion inline - it applies in one click.

HiljSV (approved):
Both checklist points are fixed now - `const` all the way through and early returns, and the extra `days < LONG_TERM` condition is gone with them. `RENT_PRICE_PER_DAY` is a better name than the one I suggested, it says exactly what the number is.

Re-ran 1, 2, 3, 4, 6, 7, 8 and 10 days: 40, 80, 100, 140, 220, 230, 270, 350 - all match the task examples, CI green.

One thing left from my earlier comment, non-blocking: the blank line between the JSDoc block and `function` is still there, which detaches the comment from the function it documents. Not a checklist item, so it does not hold anything up - just worth knowing for the next tasks.

Approving. Nice work.

PR #6693, opened 2026-08-02
add solution

recurse1337:
CHECKLIST:
- [x] Make names descriptive, so they explain the purpose of variables and functions.
- [x] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values
- [x] Prefer `const` over `let` where possible, to avoid unintentional changes.
- [x] Prefer `if` with `return` over `if else` to simplify later conditions. 
- [x] DON'T add `else` after `if` with `return` - the code after it won't be executed anyway.


haaos2009 (approved):
Good job!

PR #6696, opened 2026-08-03
develop

Tarasov-Serhii:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentForDay = 40;
```
pwcupcake228-png: 2. [Checklist] Don't use literals in expressions (HARDCODE). Instead use constants explaining this values

and for next lines

pwcupcake228-png (changes requested):
Good job!
Please fix your code

PR #6700, opened 2026-08-03
completed task

skuzmenk:


src/calculateRentalCost.js
petrisenatetana92-rgb, line 1: Good job! Your overall logic works well. Here are a few simple ways to make your code even better: 
1- Avoid using "magic numbers" (hardcoding): Values like 3, 7, 40, 20, and 50 are hard to understand at first glance. It is best to save them into const variables with clear names at the top of your code.
2- Remove else after a return:
When a code block has a return, the code stops right there. That means you don't need else if or else. Writing simple if statements one after another makes your code cleaner and easier to read!

src/calculateRentalCost.js
```diff
@@ -4,7 +4,13 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  if (days < 3) {
+    return days * 40;
```
arkatovnikita: Nice solution! One small suggestion: `days * 40` is repeated in every return statement. You could calculate the base price once and reuse it. This would make the code a little cleaner and easier to maintain.

arkatovnikita (changes requested):

PR #6701, opened 2026-08-03
calculateRentalCost solution

santana1719:


sofiatsebenko (commented):
good job

enot333lol333-coder (approved):
Good work! As a small suggestion for future tasks, you could avoid multiple return statements and update the total price before returning it. For example:
`let totalPrice = days * RENT_CAR_COST;

if (days >= MAX_DISCOUNT_DAYS) {
  totalPrice -= MAX_DISCOUNT;
} else if (days >= MIN_DISCOUNT_DAYS) {
  totalPrice -= MIN_DISCOUNT;
}

return totalPrice;`

PR #6705, opened 2026-08-03
Solution

Chempik120:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,17 @@
...
+  }
+
+  if (days < 7 && days >= 3) {
+    return days * 40 - 20;
+  }
+
+  if (days >= 7) {
+    return days * 40 - 50;
+  }
 }
```
olenabokhan1121, line 23: You can just leave this: return days * 40 - 50; - as a default.

src/calculateRentalCost.js
```diff
@@ -5,6 +5,17 @@
  */
 function calculateRentalCost(days) {
   // write code here
```
olenabokhan1121, line 7: Use constants explaining this values :                                                                                                                     
const LONG_TERM_DISCOUNT = 40; as checklist.md mentioned.

olenabokhan1121 (approved):
Looks good, left minor comments

PR #6709, opened 2026-08-03
Complete homework

enot333lol333-coder:


AlucardYagamiGame (commented):
( o.o ) 👍

    - BAD EXAMPLE:
    ```js
    if (numberOfDays >= 7) {
      return basePrice - 50;
    }
    ```
    - GOOD EXAMPLE: 
    ```js
    const LONG_TERM = 7;

    }

YatskuliakMa (approved):

PR #6711, opened 2026-08-03
solved the task

kirushamatvieiev008:


src/calculateRentalCost.js
petrisenatetana92-rgb, line 1: Good job! Your overall logic works well. Here are a few simple ways to make your code even better:
1- Avoid using "magic numbers" (hardcoding): Values like 3, 7, 40, 20, and 50 are hard to understand at first glance. It is best to save them into const variables with clear names at the top of your code.
2- Remove else after a return:
When a code block has a return, the code stops right there. That means you don't need else if or else. Writing simple if statements one after another makes your code cleaner and easier to read!

petrisenatetana92-rgb (changes requested):

petrisenatetana92-rgb (approved):
excellent!

PR #6715, opened 2026-08-04
add solution

ZorianaO:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,23 @@
  * @return {number}
  */
 function calculateRentalCost(days) {
-  // write code here
+  const rentCost = 40;
```
AlucardYagamiGame: (=^‥^=) 👍

camelCase for local variables is ok
while SCREAMING_SNAKE_CASE is used for general configurations across the whole project

Which to use in this case is up to you ^-^

AlucardYagamiGame (commented):
(=^‥^=) 👍

camelCase for local variables is ok
while SCREAMING_SNAKE_CASE is used for general configurations across the whole project

Which to use in this case is up to you ^-^

AlucardYagamiGame (changes requested):

src/calculateRentalCost.js
ZorianaO: Thank you! I've addressed the review comments )

PR #6717, opened 2026-08-04
Solution

VadymMararenko:


src/calculateRentalCost.js
```diff
@@ -5,6 +5,16 @@
...
+  let totalPrice = days * pricePerDay;
+
+  if (days >= 7) {
+    totalPrice -= 50;
+  } else if (days >= 3) {
+    totalPrice -= 20;
+  }
+
+  return totalPrice;
 }
```
Moroz-Dmytro, line 18: Create variables for ALL magic values
Name should be in UPPER_CASE

src/calculateRentalCost.js
```diff
@@ -5,6 +5,16 @@
...
 function calculateRentalCost(days) {
   // write code here
+  const pricePerDay = 40;
+  let totalPrice = days * pricePerDay;
+
+  if (days >= 7) {
+    totalPrice -= 50;
+  } else if (days >= 3) {
+    totalPrice -= 20;
+  }
```
Moroz-Dmytro, line 15: Use if with return instead of if/else if

Moroz-Dmytro (changes requested):
Check this - https://github.com/mate-academy/js_task-transportation-on-vacation/blob/master/checklist.md 
Don't skip checklist before sending to review

mkyloa (approved):

PR #6742, opened 2026-08-11
Solution

SophiiaBratanovska:


4ksandrey (changes requested):
your PR includes irrelevant files, you just need to provide solution in src/calculateRentalCost.js file
please revise it and fix

PR #6762, opened 2026-08-20
Deploy

TeoKisel:


src/calculateRentalCost.js
```diff
@@ -3,8 +3,27 @@
...
+
+  let rentCosts = days * oneDayCost;
+
+  if (days >= longRent) {
+    rentCosts -= longRentDiscount;
+  }
+
+  if (days >= shortRent && days < longRent) {
+    rentCosts -= shortRentDiscount;
+  }
```
GrizlixXx, line 24: You better use if (...) { return ...} instead if(...) {...} else if(...) {...}
Like in checklist.md p.5:
GOOD EXAMPLE:
if (condition) {
  return x;
}

return y;

Other than that, your code looks good. Keep it up! 🔥

GrizlixXx (changes requested):

GrizlixXx (approved):
Nice! Good job, Keep it up! 🔥

PR #6794, opened 2026-09-08
Solution

K1ttyHelp:


src/calculateRentalCost.js
```diff
@@ -4,7 +4,15 @@
...
  */
 function calculateRentalCost(days) {
-  // write code here
+  let sum = 40 * days;
+
+  if (days >= 7) {
+    sum -= 50;
+  } else if (days >= 3) {
+    sum -= 20;
+  }
```
brespect, line 21: define constants and remove `else`

brespect (changes requested):

